PR #5, opened 2015-01-30
Update README

### LINKED ISSUE #4: README refers to unexisting files
The README file mentions to create a Makefile  like this:
```
	include $(GOROOT)/src/Make.$(GOARCH)

	...
	include $(GOROOT)/src/Make.pkg
```
But those files no (longer?) exists in go 1.4
What's the content of those files?

OscarRyz:
Actually the makefile is not needed. Running ``go build`` compiles the files. closes #4

### DIFF
--- a/README
+++ b/README
@@ -112,18 +112,7 @@ Consider file test.proto, containing
 	  }
 	}
 
-To build a package from test.proto and some other Go files, write a
-Makefile like this:
-
-	include $(GOROOT)/src/Make.$(GOARCH)
-
-	TARG=path/to/example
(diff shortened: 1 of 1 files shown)


dsymonds:
It is unsupported to compile the generated `.pb.go` files with other `.go` files, so your replacement sentence isn't quite accurate. But I said on #4 that I'd take care of this myself.

dsymonds:
(To be clear: the original sentence was just as bad, but I want to fix it properly, not just update the bad suggestion.)

Issue #10, opened 2015-03-02
two proto files belongto same package 


coder-suncc:
For example:
a.proto
     package Test;
     message A{
	  required string name=1;
     }

b.proto
     package Test;
     import "a.proto";
     message B{
          required A a=1;	
     }

for this case ,protoc b.proto will see this : 
      import Test1 "a.pb"

thanks for reading my poor English

dsymonds:
If two files are in the same package, one shouldn't import the other, but rather both filenames should be passed to protoc together.

If they aren't actually meant to be in the same package, this is a dupe of #8.

tv42:
Actually, regardless of the bug where generated Go code imports a `foo.pb` path (#8, I see it, it's at least worked around by f7a7cb5a401616f8f2533b684d9a2fdc3d29ec14), there's some confusion in the air here.

You say two proto files in the same package should not import each other, and should just be passed to protoc together, but that doesn't seem to work:

```console
$ cat a.proto
syntax = "proto3";

package helloworld;

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}
$ cat b.proto
syntax = "proto3";

package helloworld;

message HelloRequest {
  string name = 1;
}

message HelloReply {
  string message = 1;
}
$ protoc --go_out=plugins=grpc:. a.proto b.proto || echo FAIL
a.proto:6:17: "HelloRequest" is not defined.
a.proto:6:40: "HelloReply" is not defined.
FAIL
$ protoc --go_out=plugins=grpc:. b.proto a.proto || echo FAIL
a.proto:6:17: "helloworld.HelloRequest" seems to be defined in "b.proto", which is not imported by "a.proto".  To use it here, please add the necessary import.
a.proto:6:40: "helloworld.HelloReply" seems to be defined in "b.proto", which is not imported by "a.proto".  To use it here, please add the necessary import.
FAIL
```

vrecan:
This fails for me as well.

vrecan:
I'm not sure why this is closed, I still can't get two protofiles in the same package to work with what @dsymonds  described. If I import it I get this error for every variable:
```
counter.proto:20:19: "clipboard.Counter.name" is already defined in file "counter.proto".
```
if I remove the import and use a *.proto I get this error
```
clipboard.proto:27:12: "Counter" is not defined.
```

tv42:
[...]

Issue #8, opened 2015-02-28
Incorrect imports for generated files


pdoherty:
I am unable to build a package when a proto file is included from a separate package.

src/foo/foo.proto
```
package foo;

message Foo
{
  optional string foo = 1;
}
```
protoc --go_out=src/ --proto_path=src/ src/foo/foo.proto
src/foo/foo.pb.go
```
// Code generated by protoc-gen-go.
// source: foo/foo.proto
// DO NOT EDIT!

/*
Package foo is a generated protocol buffer package.

It is generated from these files:
	foo/foo.proto

It has these top-level messages:
	Foo
*/
package foo

import proto "github.com/golang/protobuf/proto"
import math "math"

// Reference imports to suppress errors if they are not otherwise used.
var _ = proto.Marshal
var _ = math.Inf

type Foo struct {
	Foo              *string `protobuf:"bytes,1,opt,name=foo" json:"foo,omitempty"`
	XXX_unrecognized []byte  `json:"-"`
}

func (m *Foo) Reset()         { *m = Foo{} }
func (m *Foo) String() string { return proto.CompactTextString(m) }
func (*Foo) ProtoMessage()    {}

func (m *Foo) GetFoo() string {
	if m != nil && m.Foo != nil {
		return *m.Foo
	}
	return ""
}

func init() {
}
```

src/bar/bar.proto
```
package bar;

import "foo/foo.proto";

message Bar
{
  optional foo.Foo foo = 1;
  optional string bar = 2;
}
```

protoc --go_out=src/ --proto_path=src/ src/bar/bar.proto
src/bar/bar.pb.go
```
// Code generated by protoc-gen-go.
// source: bar/bar.proto
// DO NOT EDIT!

/*
Package bar is a generated protocol buffer package.

It is generated from these files:
	bar/bar.proto

It has these top-level messages:
	Bar
*/
package bar

import proto "github.com/golang/protobuf/proto"
import math "math"
import foo "foo/foo.pb"

// Reference imports to suppress errors if they are not otherwise used.
var _ = proto.Marshal
var _ = math.Inf

type Bar struct {
	Foo              *foo.Foo `protobuf:"bytes,1,opt,name=foo" json:"foo,omitempty"`
	Bar              *string  `protobuf:"bytes,2,opt,name=bar" json:"bar,omitempty"`
	XXX_unrecognized []byte   `json:"-"`
}

func (m *Bar) Reset()         { *m = Bar{} }
[...]

PR #12, opened 2015-03-02
Fix generated import statements

mattheath:
(via @mattoddie)

As described in #8 and #10 the import statements written to the generated `.pb.go` files don't seem to work as expected, which results in import statements trying to import files ending in `.pb` rather than the package - eg. `import foo "foo/foo.pb"` rather than `import foo "foo"`.

I'm uncertain if this is due to incorrect usage on my part, as our import statements are usually fully qualified import paths, such as:
`import "github.com/<org>/<repo>/proto/common/common.proto";`
which I would expect to generate an import statement of:
`import common "github.com/<org>/<repo>/proto/common"`
_not_:
`import common "github.com/<org>/<repo>/proto/common/common.pb"`

This patch determines the directory that the proto file exists in, and uses that for the import, rather than stripping off `.go` from the filename. It also stops duplicate imports being given unique names, by incrementing each, so we can just import the package.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -541,10 +541,6 @@ func RegisterUniquePackageName(pkg string, f *FileDescriptor) string {
 	// Convert dots to underscores before finding a unique alias.
 	pkg = strings.Map(badToUnderscore, pkg)
 
-	for i, orig := 1, pkg; pkgNamesInUse[pkg]; i++ {
-		// It's a duplicate; must rename.
-		pkg = orig + strconv.Itoa(i)
-	}
 	// Install it.
 	pkgNamesInUse[pkg] = true
 	if f != nil {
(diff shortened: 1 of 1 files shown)


attilaolah:
:+1: 

Without this, I see no way to use the new `proto3` messages defined in [google/protobuf][1].

[1]: /google/protobuf/tree/master/src/google/protobuf

dsymonds:
I've got something similar coming down the pipeline that should fix this equivalently.

attilaolah:
Thanks @dsymonds, would you be so kind to ping this PR once that "something similar" hits `master`?

:bowtie:

dsymonds:
@attilaolah: It went in the same day (commit dded913).

attilaolah:
Oh, I didn't notice that. Thanks, this looks great!

mattheath:
Great, thanks @dsymonds

PR #13, opened 2015-03-02
Document the grpc plugin usage for protoc

wfarr:
This was a bit non-obvious until I ended up locating https://github.com/grpc/grpc-go/blob/master/codegen.sh

Calling this out directly in these docs I think will be helpful.


tamird:
:+1: cc @dsymonds

PR #26, opened 2015-04-25
add bson note after json comment, "xx, omitempty"

ishare:
bson is widely used in golang communicate with mongodb, for example protobuf escape _id to Xid, but in mongodb `_id` is the primary key, so it need to add bson comment for column definition. 
So I suggest adding `bson` note for struct define, it benefit users using mongo, without any bad effect on others.


dsymonds:
No thanks. BSON is not widely used enough to warrant this.

PR #28, opened 2015-04-29
Fix generation of import paths.

octavore:
This is a cleaned up version of an fork of the golang/protobuf plugin used internally at Square which fixes the generation of import paths. An important change is that import paths now match exactly the proto package, or `go_package`* if that is provided, instead of using the filepath of the `*.proto` file.

fixes: https://github.com/golang/protobuf/issues/8
R: @dsymonds

### DIFF
--- a/protoc-gen-go/descriptor/Makefile
+++ b/protoc-gen-go/descriptor/Makefile
@@ -35,7 +35,7 @@ regenerate:
 	echo WARNING! THIS RULE IS PROBABLY NOT RIGHT FOR YOUR INSTALLATION
 	cd $(HOME)/src/protobuf/src && \
 	protoc --go_out=. ./google/protobuf/descriptor.proto && \
-	sed -i 's,^package google_protobuf,package descriptor,' google/protobuf/descriptor.pb.go && \
+	sed -i 's,^package protobuf,package descriptor,' google/protobuf/descriptor.pb.go && \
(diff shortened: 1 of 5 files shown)


dsymonds:
This change is doing way more than what #8 requires, and I don't think it's doing the right think w.r.t. the `go_package` option, which should only affect the declared package name and not affect the import path.

I have a fix coming down the pipe for getting the imports right (and fixing #8); watch for that and see whether it solves things for you.

octavore:
Yes, this change was broader than required by https://github.com/golang/protobuf/issues/8, I wanted to push it up in case it might be helpful. Our use case was particularly challenging because when we started using Go, we already had a large extant corpus of protos primarily developed for Java services, and we encountered the following issues:

1. files in the same folder may have different proto packages (corollary: a proto with package `a.b.c` might be in folder `d/e/f`)
2. our proto packages sometimes exhibit circular dependencies which do not manifest in our generated Java code because imports are resolved at a file level in Java, but at a package level in Go.

Those were the motivations behind the changes in this PR. Once we knew we couldn't rely on the implicit folder path -> go package mapping, it became natural to use a proto package -> go package mapping instead, and to allow the `go_package` flag to managed that in a fine-grained way to break circular dependencies. Additionally, the generated `.pb.go` was placed in a folder matching the go package. It might be useful to account for this behind a flag.

dsymonds:
You'd probably be better off using the `I` and `M` parameters that get passed through the `--go_out` flag to `protoc`. `protoc-gen-go` parses them and that is the better mechanism for overridding the mapping of .proto to .pb.go (or .proto to import path).

PR #41, opened 2015-06-15
Fix go vet warning

LK4D4:
Warning was:
github.com/golang/protobuf/proto/all_test.go:1284: result of fmt.Sprintf call not used


dsymonds:
No thanks. The code isn't wrong.

PR #42, opened 2015-06-18
Compile protos from multiple packages

tamird:
This is an issue report in the form of a PR (to ease demonstration). The primary changes reside in the second commit.

It seems to be not possible to generate go code for multiple packages (in multiple invocations of protoc, even), when dependencies exist between those packages.

Using `import_prefix` to prepend the relative path from `$GOPATH/src` won't work because the [protobuf import itself](https://github.com/golang/protobuf/blob/master/protoc-gen-go/generator/generator.go#L1153) will no longer be valid. Using the fully-qualified go import path in the protos themselves will break compilation in other languages where the directory structure required by the `go` tool isn't necessarily present. Is there a standard way to address this?

Here's the actual failure:

```
$ make test
go test ./proto
ok      github.com/golang/protobuf/proto    0.042s
make -C protoc-gen-go/testdata test
make -B my_test/test.pb.go
protoc --go_out=. my_test/test.proto
[libprotobuf WARNING /Users/tamird/src/go/src/github.com/cockroachdb/c-protobuf/descriptor.cc:6117] Warning: Unused import: "my_test/test.proto" imports "multifile/multi1.proto" which is not used.
diff -w my_test/test.pb.go my_test/test.pb.go.golden
# Invoke protoc once to generate three independent .pb.go files in the same package.
protoc --go_out=. multifile/multi{1,2,3}.proto
# Invoke protoc multiple times to build packages that are interdependent.
for dir in multipkg/bar/ multipkg/foo/; do \
      protoc --go_out=. $dir/*.proto; \
    done
go test ./...
multipkg/bar/bar.pb.go:18:8: cannot find package "multipkg/foo" in any of:
    /usr/local/Cellar/go/1.4.2/libexec/src/multipkg/foo (from $GOROOT)
    /Users/tamird/src/go/src/multipkg/foo (from $GOPATH)
make[1]: *** [testbuild] Error 1
make: *** [test] Error 2
```

[Review on Reviewable](https://reviewable.io/reviews/golang/protobuf/42)


dsymonds:
Ignore that. Rogue bot.

dsymonds:
Ignore that. Rogue bot.

neild:
Closing in favor of discussion on issue #39, see also comment on PR #40.

Issue #63, opened 2015-08-14
protoc-gen-go: import_path flag doesn't do anything


mwitkow:
We'd like to add a `go get` like prefix to all of our generated protobuf packages.

The option in `generator.go` called `PackageImportPath` seems like it should do what we want:

```
PackageImportPath string            // Go import path of the package we're generating code for
```

However, that value is never used in the codebase.

The `ImportPrefix`, unfortunately, prefixes all imports, including the full `go get` paths internal to protobufs.

mwitkow-io:
PR with a proposed fix ready. We checked and it seems to be doing exactly as advertised :)

dsymonds:
What do you mean it isn't used? It's used in the `defaultGoPackage` method.

But if you want to control the import path on a per-input basis, that's what the `M` parameter is for. It populates `ImportMap` and has full flexibility.

mwitkow-io:
Apologies, I haven't made the connection between `defaultGoPackage` and the written go lines, partially because it's not clear what it's meant to do.

I thought that when  I specify `protoc ... --go_out=plugins=grpc,import_path=mygit.com/autogen:/tmp/dest` (using `import_path` as the way to specify `PackageImportPath`) , I'd get the following:

```go
import proto "github.com/golang/protobuf/proto"

// discarding unused import google_api1 "mygit.com/autogen/google/api"
import google_protobuf1 "mygit.com/autogen/google/protobuf"
import base1 "mygit.com/autogen/base"
```
(where in a file generated from `base/serverstatus.proto`)

Instead I get:

```go
import proto "github.com/golang/protobuf/proto"

// discarding unused import google_api1 "google/api"
import google_protobuf1 "google/protobuf"
import base1 "base"
```

If, I use `import_prefix=mygit.com/autogen`, I get:
```go
import proto "mygit.com/autogen/github.com/golang/protobuf/proto"

// discarding unused import google_api1 "mygit.com/autogen/google/api"
import google_protobuf1 "mygit.com/autogen/google/protobuf"
import base1 "mygit.com/autogen/base"
```

So basically, there's no way to be able to say: prefix all the Go import paths that are an output of protoc. This is really useful when you use the `go get` convention for imports.
We're working around that by using `import_prefix` and then `sed`ing the file to remove the prefix from `github.com/golang/protobuf/proto`. How can we achieve this without such ugly hacks?

dsymonds:
[...]

PR #64, opened 2015-08-14
fix import_path parameter

mwitkow:
Fixes:
https://github.com/golang/protobuf/issues/63

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1173,7 +1173,7 @@ func (g *Generator) generateImports() {
 		if substitution, ok := g.ImportMap[s]; ok {
 			importPath = substitution
 		}
-		importPath = g.ImportPrefix + importPath
+		importPath = g.ImportPrefix + g.PackageImportPath + importPath
 		// Skip weak imports.
 		if g.weak(int32(i)) {
 			g.P("// skipping weak import ", fd.PackageName(), " ", strconv.Quote(importPath))


dsymonds:
Ignore that. Rogue bot.

dsymonds:
Ignore that. Rogue bot.

colinmarc:
This would also fix #167 for me.

theobat:
Why is this not merged ? Fixes #63 for me, and it's not a huge change

ashishnegi:
+1 . 
Any explanation for why this can not be merged will help community understand where the problem is.

zombiezen:
[David's comment](https://github.com/golang/protobuf/pull/167#issuecomment-209226123) on #167 basically sums it up: the interplay between the existing import mechanisms is complicated and subtle.  We would like to make the whole mechanism simpler, which requires careful design.

awalterschulze:
Is anyone busy doing this careful design?

roadrunner:
Hello, any progress so far?

zombiezen:
@neild is looking at it now, and we may have something soon.

jacovieASAPP:
Where's this issue left off?

neild:
(Sorry, I never noticed myself being tagged on this way back when.)

I don't understand this change. The `import_path` parameter is documented as "used as the package if no input files declare `go_package`". This PR modifies the import path of packages imported by the generated code, which is something completely different.

I admittedly don't see any reason to ever use the `import_path` parameter, since specifying the package name and import path as a `go_package` option is a strictly better alternative, but this seems to extend it in a surprising fashion inconsistent with its current behavior.

PR #68, opened 2015-08-27
Fix Make targets

tamird:
This PR fixes the make targets by:
- Updating `protoc-gen-go/testdata regenerate`
- Making the `sed` invocation BSD `sed` friendly

It also regenerates the descriptor and plugin protos from google/protobuf at v3.0.0-alpha-3.1.

### DIFF
--- a/protoc-gen-go/descriptor/Makefile
+++ b/protoc-gen-go/descriptor/Makefile
@@ -33,10 +33,10 @@
 # at src/google/protobuf/descriptor.proto
 regenerate:
 	echo WARNING! THIS RULE IS PROBABLY NOT RIGHT FOR YOUR INSTALLATION
-	cd $(HOME)/src/protobuf/src && \
-	protoc --go_out=. ./google/protobuf/descriptor.proto && \
-	sed -i 's,^package google_protobuf,package descriptor,' google/protobuf/descriptor.pb.go && \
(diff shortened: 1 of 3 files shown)


dsymonds:
Thanks, but I can't see the need for any of this. I regenerate some things here when needed, but nothing new in descriptor.proto is needed. This is just churn.

tamird:
I'm happy to remove the last commit if you think it's unneeded. The first two commits are the minimum to get `make` working on systems that have non-gnu sed.

dsymonds:
Okay, let's do just that. Squash the commits please.

tamird:
Done

Makefile
```diff
@@ -51,6 +51,6 @@ nuke:
 regenerate:
 	make -C protoc-gen-go/descriptor regenerate
 	make -C protoc-gen-go/plugin regenerate
-	make -C protoc-gen-go/testdata regenerate
+	make -C protoc-gen-go/testdata buildprotos
```
dsymonds: actually, I don't  know why that target is named differently. Can you revert this change and fix `protoc-gen-go/testdata/Makefile` instead?

tamird: sure, done.

dsymonds:
LGTM

dsymonds:
Merged in 09bcc32. Thanks.

PR #79, opened 2015-09-28
Rename README to README.md

francoishill:


dsymonds:
No thanks. It's fine how it is.

francoishill:
The reason I did this is because I can see the text is actually Markdown. Rename it on your side to `README.md` and you will see much better formatting, especially for code blocks. It is extremely hard to read like it is now as plain text.

adg:
It might resemble markdown but it's not actually a markdown document.

PR #89, opened 2015-10-28
Update README to be Markdown formatted README.md.

tswast:
I missed the installation of the C++ protocol buffer compiler step when
I first started. This is because with the plain-text formatting it was
hard to read, so I skipped to the line "to install this code." By adding
headings, it should make this mistake less common.

I also added quotes to enable
[syntax
highlighting](https://help.github.com/articles/github-flavored-markdown/#syntax-highlighting),
but tried to keep edits to a minimum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Go support for Protocol Buffers - Google's data interchange format
+# Go support for Protocol Buffers - Google's data interchange format
 Copyright 2010 The Go Authors.
 https://github.com/golang/protobuf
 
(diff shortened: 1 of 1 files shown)


dsymonds:
Thanks, but no thanks. Its current form is readable enough, and the markdown makes it less readable in non-GitHub environments.

tswast:
I disagree that the current form is readable enough, especially on GitHub where I found it particularly hard to read. Also, the changes are quite minimal to markdown-ify it, just a small change to the heading format. I could even remove the `code_format` quotes if you prefer.

I ended up wasting at least an hour trying to figure out where that `protoc` command was coming from (the C++ proto compiler), since it wasn't installed when I used the commands under "To install this software." When I search Google for "golang protobuf" it takes me to this repo, so it should be as nice as possible for developers getting things set up.

tswast:
I mailed https://github.com/golang/protobuf/pull/90 as a compromise, only adding the minimum markup needed to get it rendering nicely.

PR #93, opened 2015-11-01
Change JSON generation to lowerCamelCase

justinsb:
According to the proto3 specs, JSON fields are generated using
lowerCamelCase, not UpperCamelCase.

Change the generator to do this.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1665,7 +1665,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		ns := allocNames(base, "Get"+base)
 		fieldName, fieldGetterName := ns[0], ns[1]
 		typename, wiretype := g.GoType(message, field)
-		jsonName := *field.Name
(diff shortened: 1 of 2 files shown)


justinsb:
This doesn't change jsonpb, but does mean we can use Go's json support to go to/from JSON.

justinsb:
(Spec is here: https://developers.google.com/protocol-buffers/docs/proto3#json, but really I noticed this when trying to interop with Java.)

dsymonds:
Thanks, but I'll be taking care of this via jsonpb soon.

justinsb:
OK - great to hear.  Will there be a comparable fix to generation too, so that go's encoding/json will also work?

dsymonds:
There won't. First, we can't simply break people who might be relying on the current behaviour; second, there's certain field types that simply won't be renderable correctly by `encoding/json`.

justinsb:
Unfortunate, but reasonable.  Thanks for the explanation!

PR #98, opened 2015-11-19
Generate IDs in Go-style

gdm85:
Some tools (namely `golint`) return a warning for IDs spelled as `Id`.


gdm85:
I signed it!

dsymonds:
No thanks. We've considered this before, but it's not accurate, it's not comprehensive, and we just don't hold generated code to the same standard as human-written code.

PR #99, opened 2015-11-24
handle grpc services with no package declaration

aybabtme:
If a protobuf file for a GRPC service definition has no `package name` declaration, the service name currently generated is `.{{ServiceName}}`.

Clients in other languages seem to expect package-less services to be named ``{{ServiceName}}``.

For instance, in the NodeJS client: https://github.com/grpc/grpc/blob/8bbc4e0ac98901ca83b0f71791fba04ffbd84401/src/node/src/common.js#L92-L96


dsymonds:
Thanks. The spec (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) does indeed say that the package name is optional.

I'll take care of this.

PR #100, opened 2015-11-24
Document parameter interpretation in README

zellyn:
Adds a section to the README file on Parameter interpretation.

### DIFF
--- a/README
+++ b/README
@@ -151,6 +151,28 @@ To create and play with a Test object from the example package,
 	}
 
 
+Parameters
+==========
+
+To pass extra parameters to the plugin, use a comma-separated
+parameter list separated from the output directory by a colon:
+
+
+	protoc --go_out=plugins=grpc,import_path=mypackage:. *.proto
+
+
+- `import_prefix=xxx` - a prefix that is added onto the beginning of
+  all imports. Useful for things like generating protos in a
+  subdirectory, or regenerating vendored protobufs in-place.
+- `import_path=foo/bar` - used as the package if no input files
+  declare `go_package`. If it contains slashes, everything up to the
+  rightmost slash is ignored.
+- `plugins=plugin1+plugin2` - specifies the list of sub-plugins to
+  load. The only plugin in this repo is `grpc`.
+- `Mfoo/bar.proto=quux/shme` - declares that foo/bar.proto is
+  associated with Go package quux/shme.  This is subject to the
+  import_prefix parameter.
+
 gRPC Support
 ============
 If a proto file specifies RPC services, protoc-gen-go can be instructed to


zellyn:
I signed it! (I'm part of a CCLA signature and a member of the group): squareup.com

README
```diff
@@ -151,6 +151,32 @@ To create and play with a Test object from the example package,
 	}
 
 
+Parameter interpretation
```
dsymonds: Parameters

README
```diff
@@ -151,6 +151,32 @@ To create and play with a Test object from the example package,
...
+parameter list separated from the output directory by a colon:
+
+
+	protoc --go_out=plugins=grpc,import_path=mypackage:. *.proto
+
+
+- `import_prefix=xxx` - a prefix that is prepended onto the beginning
+  of all imports. Useful for things like generating protos in a
+  subdirectory, or regenerating vendored protobufs in-place.
+- `import_path=foo/bar` - struct comments say, “Go import path of the
```
dsymonds: rephrase so it doesn't sound like a voyage of discovery.

README
zellyn: Hmmm. My Sough African / American sense of English must be subtly different. How does  "Interpretation of parameters" look?

Also, the field in plugin.proto is called "parameter" which is even more confusing :-)  https://github.com/google/protobuf/blob/b640f216a56ae25aab9a1fc6f21aa1a56c304494/src/google/protobuf/compiler/plugin.proto#L64

README
```diff
@@ -151,6 +151,32 @@ To create and play with a Test object from the example package,
...
+  package we're generating code for” - this is ignored unless there is
+  no explicit `go_package` declared in any of the files. In that case,
+  this `import_path` is used as the package (ignoring any part of it
+  up to the rightmost slash).
+- `plugins=plugin1+plugin2` - specifies the list of sub-plugins to
+  load. The only plugin is `grpc` unless you've forked the `main`
+  package.
+- `Mfoo/bar.proto=quux/shme` - an explicit import mapping: this
```
dsymonds: ... - declares that `foo/bar.proto` is associated with Go package `quux/shme`.
  This is subject to the `import_prefix` parameter.

README
dsymonds: Your suggestion is still too narrow. The headline doesn't need to be that specific.

That `parameter` field in `plugin.proto` is exactly what is being used here.

zellyn: Ah, just realized I misread your comment as meaning s/Parameter/Parameters/.

Done.

README
zellyn: ptal

README
zellyn: Much better.

README
```diff
@@ -151,6 +151,29 @@ To create and play with a Test object from the example package,
...
+==========
+
+To pass extra parameters to the plugin, use a comma-separated
+parameter list separated from the output directory by a colon:
+
+
+	protoc --go_out=plugins=grpc,import_path=mypackage:. *.proto
+
+
+- `import_prefix=xxx` - a prefix that is prepended onto the beginning
```
dsymonds: s/prepended/added/

Prefixes always appear at the start. Also Rob gets grumpy when people use "prepend".  ;-)

README
```diff
@@ -151,6 +151,29 @@ To create and play with a Test object from the example package,
...
+  of all imports. Useful for things like generating protos in a
+  subdirectory, or regenerating vendored protobufs in-place.
+- `import_path=foo/bar` - used as the package if no input files
+  declare `go_package`. If it contains slashes, everything up to the
+  rightmost slash is ignored.
+- `plugins=plugin1+plugin2` - specifies the list of sub-plugins to
+  load. The only plugin is `grpc` unless you've forked the `main`
```
dsymonds: The only plugin in this repo is `grpc`.

dsymonds:
Thanks. This looks pretty close. Just a couple small nits, then please squash these commits and I'll merge.

zellyn:
Done

dsymonds:
Merged in 8e6e0a5. Thanks!

PR #101, opened 2015-11-24
Allow for interception of package and file mapping

zellyn:
Introduces a PackageMapper interface, and moves functionality relating
to package names and paths into a default implementation. This allows
clients to instantiate a generator and easily customize how generated
files are arranged. It also allows the remapping of standard helper
packages.

I don't necessarily expect this PR to survive intact, but wanted to take a
stab at a minimally-invasive attempt to allow alternate interpretations
of go_package. Comments and suggestions welcome.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -258,27 +258,6 @@ type FileDescriptor struct {
 // PackageName is the package name we'll use in the generated code to refer to this file.
 func (d *FileDescriptor) PackageName() string { return uniquePackageOf(d.FileDescriptorProto) }
 
-// goPackageName returns the Go package name to use in the
-// generated Go file.  The result explicit reports whether the name
-// came from an option go_package statement.  If explicit is false,
-// the name was derived from the protocol buffer's package statement
-// or the input file name.
-func (d *FileDescriptor) goPackageName() (name string, explicit bool) {
-	// Does the file have a "go_package" option?
-	if opts := d.Options; opts != nil {
-		if pkg := opts.GetGoPackage(); pkg != "" {
-			return pkg, true
-		}
-	}
-
-	// Does the file have a package clause?
-	if pkg := d.GetPackage(); pkg != "" {
-		return pkg, false
-	}
-	// Use the file base name.
-	return baseName(d.GetName()), false
-}
-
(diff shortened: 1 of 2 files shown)


zellyn:
To comment on my own initial stab: on reflection, it seems clear that _all_ uses of `import_prefix` should move inside the PackageMapper, and probably also the M-parameter mapping functionality. I'm less sure about `import_path`, although I could see that moving inside too. I'll update with at least `import_prefix` moved inside.

zellyn:
I'm going to close this and work on it further - it's a lot more subtle than it appears at first blush.

dsymonds:
Yep, there's a reason why I don't immediately accept all the changes that people want to make to the package semantics here: it's subtle.  ;-)

zellyn:
Tell me if this sounds sane:

1. Create a (pluggable) package mapper that can tell you, for a given filedescriptor, what full filename it wants, what full package path it wants to be in, and what package name it wants to declare. Also handles import_prefix, M mappings, and import_path (for figuring out default package when none is specified)
2. Remove the restriction that all generated files have to be in the same package
3. Don't use package names for things that are in the same package (which we can figure out by asking the mapper if two things have the same full package path and package declaration)
4. (Optional, but would make the generated code a little cleaner) keep track of unique import names per-file, and use the same uniquified import name for things in the same package

Open questions: how to know whether we should output things that should be once per package? eg. package doc comment: is there anything else?

dsymonds:
I'm not sure about your 2nd point. We have a few assumptions baked in that a set of proto files passed to proto maps to a single Go package. I suppose you're suggesting undoing that, and I agree it'd be more flexible, but it seems like it'd end up more confusing than today's simple mapping.

I'm nervous about making this too flexible. Sure, it's useful, but it's a lot of complexity too. We already have the parameters, which are admittedly clumsy to use, but technically cover a reasonable spectrum of things that can be done where the default isn't suitable.

zellyn:
Sorry for missing this earlier.

Yes, I'm definitely trying to undo the all-packages-at-once assumption. Also, I'm trying to substitute one clear pluggable mapper for a bunch of flags and reliance on package, etc. - I wholly agree that even the existing degree of flexibility is what makes things painful :-)

PR #108, opened 2015-12-07
protoc-gen-go: add `url` flag to protobuf fields with json name

parkr:
Adds support for https://github.com/google/go-querystring

Need to look into tests.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1690,7 +1690,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		fieldName, fieldGetterName := ns[0], ns[1]
 		typename, wiretype := g.GoType(message, field)
 		jsonName := *field.Name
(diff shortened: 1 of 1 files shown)


dsymonds:
Thanks, but no. We aren't going to bloat out generated code for everyone just to support go-querystring.

parkr:
> We aren't going to bloat out generated code for everyone just to support go-querystring.

Ok! Curious how you would go about interoperability. At this point, I'm using code generation: inserting a `//go:generate` comment at the top of my `.pb.go` files and generating a `func (f myStruct) UrlValues() url.Values` based on the code it reads. It's very heavy-handed. Is there a way I can tell `protoc-gen-go` to add more tags? We have also wanted tags customization for much simpler database (mysql, mongo, etc) deserialization.

dsymonds:
Struct tags simply aren't a good solution when you don't control the types. The same would apply if the types were defined in a package that you don't own. Ideally the relevant package would provide some other form of control, or you need to implement a shadow set of wrapper types. It's going to depend on the situation.

PR #110, opened 2015-12-15
Add package_comments parameter to enable or disable generated package comments

peter-edge:
This adds a parameter to enable or disable package comments. A lot of the time, a go package will have generated protobuf code along side it, and the generated package comments will override any other package comments in godoc. Whether or not this is a good design decision is a good debate heh, but that's a separate issue - a lot of users do this (including myself half the time).

By default, package comments will be turned on. To disable:

```
# assuming foo.proto in your current directory
protoc --go_out=package_comments=false:. foo.proto
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ parameter list separated from the output directory by a colon:
   rightmost slash is ignored.
 - `plugins=plugin1+plugin2` - specifies the list of sub-plugins to
   load. The only plugin in this repo is `grpc`.
+- `package_comments=false` - will disable package comments in the generated go files
 - `Mfoo/bar.proto=quux/shme` - declares that foo/bar.proto is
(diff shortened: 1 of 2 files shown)


dsymonds:
Sorry, we don't support intermingling generated protobuf code in the same package as hand-written code. The package doc for that is the least of your problems. I don't want this added parameter and complexity to support that path.

peter-edge:
I understand the reasoning, but for the record, I think more than half of OSS code does just this, and in my humble opinion it's not terrible to support it :)

dsymonds:
Naming conflicts are the immediate problem (protoc-gen-go knows to avoid names that conflict with other names it produces, but not with anything else), but we're also pretty liberal in changing unexported details with no notice, and those details are the only particularly good reason to put human written code in the same package as protobuf generated code.

PR #113, opened 2016-01-13
Use JSON name if available

bashtian:


bashtian:
I signed it!

bashtian:
I signed it!

dsymonds:
Thanks, but I'm waiting for the upstream change to set that field to what it ultimately will be (they're doing camelCase), and I have a change ready to go when that happens already.

PR #120, opened 2016-01-24
Adding repeated field data population example

zmullett:
Data population for the 'repeated' field rule was not previously described in the code example. This addition also aids in highlighting the purpose of the proto field helper functions; to return pointers values. At first glance I interpreted them as type wrappers and couldn't understand why I couldn't set Reps to "[]proto.Int64{...}".


zmullett:
I signed it!

zmullett:
I signed it! (again)

PR #130, opened 2016-02-13
added VisitMessage function

davyxu:
VisitMessage is used for global iterate message


dsymonds:
Thanks, but no thanks. We're not ready to expand this API yet.

PR #131, opened 2016-02-15
Add descriptor package

myitcv:
Per 0dfe8f37844c14cb32c7247925270e0f7ba90973 et al


types/regen.sh
```diff
@@ -58,6 +59,7 @@ for up in "${!filename_map[@]}"; do
   cat $tmpdir/$UPSTREAM_SUBDIR/$up |
     # Adjust proto package.
     # TODO(dsymonds): Upstream the go_package option instead.
+    grep -v '^\s*option\s\+go_package' |
```
myitcv: This hack can similarly be removed once we consistently have `go_package` set in the upstream `.protof` files

dsymonds:
Descriptor isn't really a "well-known type" like the others are.

myitcv:
It's required when defining custom options:

https://developers.google.com/protocol-buffers/docs/proto#customoptions

When defining a custom option the generated code references either
FileOptions or FieldOptions (both defined in descriptor.proto)

dsymonds:
That doesn't really address what I said.

myitcv:
@dsymonds sorry, I didn't mean to suggest that the `descriptor` package contains "well-known" types like `Duration` et al (especially given the documentation describes custom options as an _advanced feature_)

Rather that we could reuse the mechanism you have created for generating the well-known types for the `descriptor` package.

Unless you had a different approach in mind for the `descriptor` package, or more generally custom options?

peter-edge:
Most people use https://github.com/golang/protobuf/tree/master/protoc-gen-go/descriptor, although it would be nice to add the descriptor.proto file there as well.

Globegitter:
I tested this PR out on a fork (https://github.com/Globegitter/protobuf) and to get it to work I had to add a `descriptor.go` file into the ptypes folder: https://github.com/Globegitter/protobuf/blob/master/ptypes/descriptor.go

Otherwise I would get `proto_name.pb.go:34: can't find import: "github.com/globegitter/protobuf/ptypes/descriptor"`

yohcop:
should the other descriptor.pb.go in protoc-gen-go/descriptor be removed? (this pull request is very old now, so at that time maybe the one in protoc-gen-go/descriptor didn't exist... but still, something to remember before merging)

PR #137, opened 2016-02-25
Add a `package` override parameter.

zellyn:
The commandline flags allow overriding output location, package import
locations, and the default package name if none is declared. The
`package` flag allows forcing an explicit package name.

Context: we need to move Go code around to break import cycles. I'm
writing a go-protoc-wrapper that interprets `package` and `go_package`
as specifying full package path (like Java does), but that mangles the
declared package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ parameter list separated from the output directory by a colon:
 - `Mfoo/bar.proto=quux/shme` - declares that foo/bar.proto is
   associated with Go package quux/shme.  This is subject to the
   import_prefix parameter.
+- `package=mypackage` - an explicit package name that will override
+  `go_package` in the generated files.
 
 ## gRPC Support ##
 
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -513,6 +513,7 @@ type Generator struct {
 	PackageImportPath string            // Go import path of the package we're generating code for
 	ImportPrefix      string            // String to prefix to imported package file names.
 	ImportMap         map[string]string // Mapping from import name to generated name
+	ForcedPackageName string            // Forced declared package for generated files.
 
 	Pkg map[string]string // The names under which we import support packages
 
@@ -574,6 +575,8 @@ func (g *Generator) CommandLineParameters(parameter string) {
 			g.PackageImportPath = v
 		case "plugins":
 			pluginList = v
+		case "package":
+			g.ForcedPackageName = v
 		default:
 			if len(k) > 0 && k[0] == 'M' {
 				g.ImportMap[k[1:]] = v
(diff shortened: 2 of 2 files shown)


dsymonds:
This is a lot of code for a mechanism for which we have several existing hooks. `go_package` is the thing to use. We recommend renaming proto imports (e.g. writing `import foopb "path/to/the_proto"), which also makes the generated package name moot.

If you're already re-interpreting `go_package`, why can't you use that?

zellyn:
Hey @dsymonds, let me explain the background; perhaps you can see a cleaner solution I'm missing. Apologies for the long comment…

We have a large body of existing protos, mostly coming from Java and Ruby, not all perfectly arranged (eg. `package` doesn't always match location on disk). Go is relatively new here, so needs to adapt, rather than imposing changes. Go's package-level circular-import restriction is more strict that protobufs (which only disallows direct circular includes) or Java's (which allows circular imports within single compiles), so we have no choice but to move things around (package-wise) in the Go generated code.

Organizing protos into specific packages is easy at Google: everything already requires BUILD files. But here, currently, our Go developers are enjoying the lack of Make/Pants/Bazel build files, and I'd hate to impose something like that just to move protos around. (It also seems more Go-ish to include everything needed to build in the source code.) The obvious choice (to me) seems to be to interpret `go_package` as `java_package` is interpreted: as defining the full path, not just package _name_.

One of my predecessors here actually forked the Go proto plugin, making it interpret `go_package` that way. However, the fork was done… um, expediently… and is increasingly painful to maintain, because the package and import code is spread all over. It's also brittle - the changes are incompatible with both `public import` and the import name conflict resolution is broken - we just happen never to import two things with the same name.

Having packages move around, though necessary, is also painful: it makes the "package-at-a-time" idiosyncrasy of the go plugin difficult to adhere to: you never know just quite which files are going to land in which packages.

I've considered several options: (a) trying to add code upstream to make the parts of the code that deal with packages and paths pluggable. (b) just publicly forking the plugin, maintaining it, and taking community contributions, and my current plan (c) write a pre-processor that maps out the entire structure of packages (by examining `go_package` and `package`), and then calls `protoc` with one go-package-worth of files at a time, using a whole slew of options to import and name things properly. However, there's no way to directly specify the output package name, and if I use `go_package`, I have to find some other mechanism for specifying the desired location: I guess I could add a custom option, but that seems less general.

As a final note: as I hope you can tell, I'm eager to unfork our codebase as much as possible, especially in the package-dealing parts, because they're so entwined all over the place. I'm also eager to work with you folks to find minimal changes, surgical interception points, or pluggable functionality that would get us to where we need to be. We (Square) are working closely as partners with the GRPC team, and would love to work with the protobuf folks too; in fact, I think we're going to try to loop some of them in on our next biweekly video conference.

dsymonds:
I totally understand your motivations here, and I appreciate your concerns. They're definitely reasonable things to want to do, and I'd like to help get your fork minimal.

My role here is to make sure the changes we make will be suitable for lots of people, and that it's something we can commit to supporting over the long term. That's not easy, and unfortunately I don't have anywhere near enough time to do that as expediently as it needs to be done. I'm sorry for this situation.

You say your fork already re-interprets the `go_package` option. Is it a reasonable interim solution to overload that a bit more and inject the package name in that way? For instance, if that option ends in `;foo`, then consider `foo` to be the package name?

zellyn:
Well, I can create whatever semantics I want in the fork - the problem is that it's brittle, and I don't want to keep maintaining it, especially as proto3 seems to be dramatically increasing the rate of commits.

The problem this PR addresses is that whatever overloading I do in `go_package`, the generator will just run `bad_to_underscore` on them. So I could make the `go_package` "creditcard;foo/bar", let my wrapper interpret that as "output directory: foo/bar" and "package creditcard". But the generator would then generate a package statement of `package creditcard_foo_bar`. As you mentioned, if I change all proto references in our codebase to use explicit package aliases, that won't be a problem, but that's a bit ugly. I'd rather post-process with `sed`, as ugly as that is too :-)

dsymonds:
Does #139 now make this moot?

zellyn:
Yes!

PR #142, opened 2016-03-08
Add ptypes.EmptyInstance, ptypes.TimestampBefore

peter-edge:
Signed-off-by: Peter Edge peter.edge@gmail.com

This is bringing over a couple things from https://github.com/peter-edge/pb/blob/master/go/google/protobuf/protobuf.gen.go. I made some different design decisions there, but these are the couple things that might be interesting.


peter-edge:
ping/should we close this if it won't be merged?

cybrcodr:
Similar argument as I'd have with PR #369 .  I'm going to close this.  Thanks for the proposal though.

PR #144, opened 2016-03-08
Update ptypes/regen.sh to use new go_package specification and make some mac-fixes

peter-edge:
Signed-off-by: Peter Edge peter.edge@gmail.com

Makes a few fixes here and there in the script as well:
- Uses `set -Ee` instead of `#!/bin/bash -e` so that invocations like `bash -x ptypes/regen.sh` get the `-e` as well (also adds `-E`)
- Switches on `sed` depending on if this is being run from mac or linux since `sed -i` is weird on mac (not GNU sed)
- Does not do `xargs dirname` when going over directories as `dirname` can only take one argument on mac


ptypes/regen.sh
```diff
@@ -54,18 +76,23 @@ done
 # Pass 2: copy files, making necessary adjustments.
 for up in "${!filename_map[@]}"; do
   f=${filename_map[$up]}
-  shortname=$(basename $f | sed 's,\.proto$,,')
+  shortname=$(basename $f | $(which_sed) 's,\.proto$,,')
   cat $tmpdir/$UPSTREAM_SUBDIR/$up |
     # Adjust proto package.
     # TODO(dsymonds): Upstream the go_package option instead.
```
dsymonds: This TODO is in progress; I'd prefer to just wait for it to land, and then rip this out entirely.

ptypes/regen.sh
```diff
@@ -1,10 +1,12 @@
-#!/bin/bash -e
+#!/bin/bash
 #
 # This script fetches and rebuilds the "well-known types" protocol buffers.
 # To run this you will need protoc and goprotobuf installed;
 # see https://github.com/golang/protobuf for instructions.
 # You also need Go and Git installed.
 
+set -Ee
```
dsymonds: There's O(few) people that will ever run this script. They can just run it directly instead of subverting the shebang line.

ptypes/regen.sh
```diff
@@ -22,6 +24,26 @@ function die() {
   exit 1
 }
 
+function which_sed() {
+  local uname_s="$(uname -s)"
+  case "${uname_s}" in
+    Darwin)
+      if ! which gsed > /dev/null; then
+        die "must install gsed on a mac, try brew install gnu-sed"
```
dsymonds: ewww.

What GNU-isms are we actually using below? Can we fix those instead, and leave it as invoking plain "sed"?

ptypes/regen.sh
dsymonds: Actually the new WKTs that you were pushing for make this more pressing.  :-/

mwmahlberg:
I started the work in parallel and was able to further simplify the script and make it more portable by reducing potentially critical dependencies. See #200 for details.

ptypes/regen.sh
```diff
@@ -54,18 +76,23 @@ done
...
+    $(which_sed) '/^package /a option go_package = "'github.com\/golang\/protobuf\/ptypes\/${shortname}'";' |
     # Unfortunately "package struct" doesn't work.
-    sed '/option go_package/s,"struct","structpb",' |
+    $(which_sed) '/option go_package/s,struct",struct;structpb",' |
     cat > $PKG/$f
 done
 
 # Run protoc once per package.
-for dir in $(find $PKG -name '*.proto' | xargs dirname | sort | uniq); do
+#
```
bmkessler: `xargs -n1 dirname` will run dirname for each file in the output of find (-n is part of the posix xargs standard so should be cross platform).  The implementation below will run multiple times per directory if there are multiple .proto files in that directory

bmkessler:
If you are fixing regen.sh to work on OS X, the `declare -A` section L44+ will need to be fixed as well.  Associative arrays are a bash 4 feature and OS X runs bash 3 with no intention to upgrade.

Issue #146, opened 2016-03-10
Generator code requires Go 1.6.


tswast:
The commit https://github.com/golang/protobuf/commit/c75fbf01dc6cb73649c4cd4326182c3e44aa9dbb uses `strings.LastIndexByte`, which I believe was introduced in Go 1.6. We should update the README to say Go 1.6 is required, if this requirement was intentional.

tswast:
This broke the Go sample build in the main protobuf repo. See: https://github.com/google/protobuf/issues/1305

dsymonds:
My mistake. I thought that function had been there for much longer. Easy enough to fix.

Weird that it took a week for this to manifest...

dsymonds:
`api/go1.5.txt` in the Go tree implies that `strings.LastIndexByte` was actually added in Go 1.5.

tswast:
Okay, either way the README says Go 1.4, so we should update it. In the meantime, I'll get the protobuf repo using later versions of Go.

dsymonds:
Yeah, automated CI systems should probably just stick with the latest stable Go release.

I'll update the README. I think it's reasonable to expect at least Go 1.5 these days.

dsymonds:
Actually, it's so easy to leave it supporting Go 1.4 I'll do that instead.

PR #162, opened 2016-04-01
Add Travis configuration to ensure this builds.

tswast:
This should help catch changes where we break support for older versions, such as https://github.com/golang/protobuf/issues/146.

Based loosely on https://github.com/golang/oauth2/blob/master/.travis.yml.

In a conversation with the folks from the google/protobuf over https://github.com/google/protobuf/pull/1316, they would prefer if I move the Go protobuf sample to this repo, so having a Travis config in place can help make sure those are tested, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: go
+
+go:
+  - 1.4
+  - 1.5
+  - 1.6
+
+install:
+  - export GOPATH="$HOME/gopath"
+  - mkdir -p "$GOPATH/src/github.com/golang"
+  - mv "$TRAVIS_BUILD_DIR" "$GOPATH/src/github.com/golang/protobuf"
+  - go get -v -t -d github.com/golang/protobuf/...
+
+script:
+  - go test -v github.com/golang/protobuf/...


tswast:
Since we don't have Travis set up on this repo, yet, I tested it here: https://travis-ci.org/tswast/goprotobuf/builds/120171306

dsymonds:
I'd prefer not. It only rarely breaks, and I've found Travis more trouble than it's worth for a repo like this.

tswast:
Closing this PR. I will try to piggy-back on the google/protobuf Travis tests.

awalterschulze:
I just think I should mention.
I also kicked against and thought travis was more trouble than its worth, but now I am super happy with it.
Its really nice for running tests against the cross product of all the versions of go I support and all the versions of protoc I support.  Its really cool, but its your choice.

zellyn:
This is why we need and want Jenkins: https://github.com/golang/protobuf/commit/f6b4231c7fd7ac9df487f17759fc107fd61d98a0#commitcomment-18062074

zellyn:
@matloob fyi…

PR #165, opened 2016-04-12
add lenient option for parsing json

spf2:
This adds an option to make JSON parsing lenient.  For context: https://github.com/google/protobuf/issues/1389

I noticed there was no unmarshaler object for adding options, so I added one of those as well.  I left the old functions around for backwards compatibility.  I'm new to go, so not sure if reusing function names as method names is a nice practice or not?

If this looks like it's going in the right direction, I'll also add the missing doc and comments.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -76,6 +76,11 @@ type Marshaler struct {
 	OrigName bool
 }
 
+type Unmarshaler struct {
+	// Whether to ignore unknown fields when parsing JSON.
+	Lenient bool
+}
+
 // Marshal marshals a protocol buffer into JSON.
 func (m *Marshaler) Marshal(out io.Writer, pb proto.Message) error {
 	writer := &errWriter{writer: out}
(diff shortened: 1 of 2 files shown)


spf2:
I signed the CLA

dsymonds:
Sorry, the spec doesn't discuss lenient parsing, so I'm not inclined to agree with this, as useful as it may seem.

spf2:
Oh... is there more detailed version of the spec than the one I find here: https://developers.google.com/protocol-buffers/docs/proto3#services?

Is there a process for providing input into the spec? I don't think this is a minor issue.  As I'm sure you know, one of the key ideas of protobufs is that you can add tag numbers w/o breaking existing code. (Imagine if adding a tag number to a proto broke every service that used it until recompiled!) As I suggested in the bug, (and you surely are aware) both protos and json are intentionally designed to allow adding new fields without breaking existing code, and so it's an odd choice that the json mapping for protos would have this new strict behavior.  Additionally, it's pretty hard to work around this limitation, and so it's probably better to skip the json->proto parsing wherever possible, which would be unfortunate...

Strictness seems useful perhaps for testing, but not for production environments where services are incrementally updated and deployed.

My $0.02...

dsymonds:
Unfortunately there's nothing published any more specific than that. You can file a bug in the google/protobuf repo to get a more detailed spec, or to give them feedback.

The proto wire format is designed to cope with unknown (new) fields, but I'm not sure the JSON format is really intended for that, particularly since it often gets hand-written. There's a bunch of cases where the wire format and JSON format diverge around this.

But I'm not the one to make the call on what the JSON format should be. That should be defined upstream, and then we'll make sure jsonpb is compliant with whatever the spec is.

spf2:
Ok I'll file the bug... thanks for you response and conforming to a spec makes a lot of sense of course.

PR #167, opened 2016-04-13
Never use `go_package` to munge the output filename

colinmarc:
The `go_package` option is useful in defining how the generated files
can be imported. But the existing implementation also changes the
output filename based on it, causing some weird results if you are
running `protoc` from the project root, which is a common operation.

Let me illustrate with an example.

Imagine you're sitting in `$GOPATH/src/github.com/colinmarc/myproject.`
From there, you have the following directory structure:

    foo/
      something.go
      foo.proto
    bar/
      baz/
        baz.proto

foo.proto declares the following:

    option go_package = "github.com/colinmarc/myproject/foo";

and baz.proto wants to import it:

    import "foo/foo.proto";

Now you run `protoc --go_out=. -I. foo/foo.proto`, and the same for
the other file. It works, but the generated files end up in a totally
weird place! Your directory structure now looks like this:

    foo/
      something.go
      foo.proto
    bar/baz/
      baz.proto
    github.com/
      colinmarc/
        myproject/
          foo/
            foo.pb.go
          bar/
            baz/
              baz.pb.go

Which is not useful at all. So you try a few different things. First,
you change the `go_package` declaration in foo.proto, to:

    option go_package = "foo"

Again, everything compiles fine, and this time the generated files
in the right place. But the imports in `baz.pb.go` are broken:

    import "foo"

Which again, is not a thing that makes any sense in go.

Said another way, the `go_package` option is useful if it munges either
the imports *or* the filepath, but not both.

I can't really think of a useful situation for having the generated
file end up in a different place from the .proto file; so this just
removes that functionality altogether. If you generate a file in
the wrong package, the go compiler will fail and tell you that you
have two packages in the same directory - which is exactly what would
happen if you declared the wrong package in a go source file instead
of a protobuf source file.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -316,15 +316,6 @@ func (d *FileDescriptor) goFileName() string {
 	}
 	name += ".pb.go"
 
-	// Does the file have a "go_package" option?
-	// If it does, it may override the filename.
-	if impPath, _, ok := d.goPackageOption(); ok && impPath != "" {
-		// Replace the existing dirname with the declared import path.
-		_, name = path.Split(name)
-		name = path.Join(impPath, name)
-		return name
-	}
-
 	return name
 }
 
@@ -1327,11 +1318,15 @@ func (g *Generator) generateImports() {
 			continue
 		}
 		filename := fd.goFileName()
-		// By default, import path is the dirname of the Go filename.
+		// By default, import path is the dirname of the Go filename. If an explicit
+		// mapping or a go_package declaration is available, use that instead.
 		importPath := path.Dir(filename)
 		if substitution, ok := g.ImportMap[s]; ok {
 			importPath = substitution
+		} else if impPath, _, ok := fd.goPackageOption(); ok && impPath != "" {
+			importPath = impPath
 		}
+
 		importPath = g.ImportPrefix + importPath
 		// Skip weak imports.
 		if g.weak(int32(i)) {


colinmarc:
(signed the CLA!)

dsymonds:
No, this was intentional. Go package import paths correlate 1:1 with their position on the file system. If you say

```
option go_package = "github.com/colinmarc/myproject/foo";
```

then you're saying you want `"github.com/colinmarc/myproject/foo"` to be the import path for the generated package, which means the .pb.go files need to be in `github.com/colinmarc/myproject/foo`, relative to the GOPATH. The intended operation is to use `--go_out=$GOPATH/src` if you don't have some other way of moving files around.

colinmarc:
`--go_out=$GOPATH/src` seems kinda wonky, though, doesn't it? If I have a rule in my Makefile:

```
%.pb.go: %.proto
	protoc -I$(GOPATH)/src --go_out=$(GOPATH)/src $<
```

And some user checks out my library/whatever outside the GOPATH and runs `Make`, it'll overwrite the files in a totally different place.

What's the suggested "some other way of moving files around"? I think part of the problem here is that there's no obvious default way to set things up.

dsymonds:
Yeah, it's not perfect, but the `go_package` override is there for precisely the situations where you *want* to override the Go package (as its name suggests). If you've got the .proto files already in the right place, don't use a full import path in that option.

Have you looked at using the `M` parameter (see the README)? It allows you to override what package import path to use for any given .proto file, and that might solve your issue when you've got a simple `option go_package = "foo";`.

colinmarc:
Specifying the explicit mapping for each file is a chore. 

What about if we did it the same way go resolves a package name, if the package is in the `go_package = "foo"` form? We can shell out to `go list <dir of the .proto file>` to tell us the import path of the current directory, and root imports at that. There might be an actual importable library for that, I don't know.

By the way, I just tried `protoc -I"$GOPATH/src" --go_out="$GOPATH=src"`, and got the following error:

```
File does not reside within any path specified using --proto_path (or -I).  You must specify a --proto_path which encompasses this file.  Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think).
```

I think that might be because I'm one of those heathens that symlinks directories from `$GOPATH/src` into a `~/dev`, but it's another wrinkle in the plan.

dsymonds:
Have you tried running protoc _from_ $GOPATH/src, and just passing `.` to those flags?

colinmarc:
That appears to work, but that's *definitely* wonky.

One more suggestion. Cap'n Proto does this with an explicit `$Go.import("github.com/colinmarc/myproject/foo")`, which specifies how the file should be imported once it's generated; essentially, it moves the `M` mapping thing into the file itself, which is a bit cleaner.

dsymonds:
Argh! That's what `go_package` is for, if it can't be guessed based on filenames and the `M` arg.

colinmarc:
Sorry, not trying to be a pain. =) Feel free to close this, of course, but I think there's a major usability problem out of the box right now. When I googled around, it seemed like a lot of people were using `import_prefix` + `sed` to munge the `golang/protobuf` import. Perhaps there's a better way to correctly generate imports for `go_package "foo"` automagically?

dsymonds:
Yes, we're aware of the issues here. The tools we currently have handle about 80% of cases, but it's already quite complex (as you've seen), and we can't really close the gaps by layering on more complexity. It needs some careful thought.

PR #186, opened 2016-05-19
Work around gccgo bug (see golang.org/issue/15738)

mwhudson:
github.com/golang/protobuf/proto having both internal and external tests
triggers a surprising bug in gccgo. As the bug only affects tests, it seems
reasonable to workaround it by not compiling the internal tests with gccgo.


mwhudson:
Ping? My github account got flagged for a while so it's possible notifications for this PR didn't go out.

LMMilewski:
golang.org/issue/15738 is closed. Is the workaround still needed?

mwhudson:
I'm not sure if that fix is in any released version of gccgo yet, but you have a point.

paranoiacblack:
This fix is indeed released on the latest gccgo; please re-open if that doesn't satisfy the need for this feature.

PR #188, opened 2016-05-24
jsonpb: Adds support for other []byte types to be base64 encoded.

d4l3k:
The proto3 spec encodes the bytes type to be a base64 encoded string
when marshalling to jsonpb. This adds support to jsonpb to correctly
encode go types in the format `type Bytes []byte` into base64 instead of
being represented as an uint8 array.

While these types won't be generated by the standard protobuf compiler,
they can be generated by others such as gogo protobuf.

### DIFF
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -282,6 +282,21 @@ func init() {
 	}
 }
 
+// Bytes is used to test that []byte type aliases are serialized to base64.
+type Byte byte
+
+// Bytes is used to test that []byte type aliases are serialized to base64.
+type Bytes []Byte
+
+// BytesMessage is used to test that []byte type aliases are serialized to base64.
+type BytesMessage struct {
+	Bytes Bytes `protobuf:"bytes,1,opt,name=bytes,json=bytes" json:"bytes,omitempty"`
+}
+
+func (*BytesMessage) Reset()         {}
+func (*BytesMessage) String() string { return "" }
+func (*BytesMessage) ProtoMessage()  {}
+
 var marshalingTests = []struct {
 	desc      string
 	marshaler Marshaler
@@ -373,6 +388,7 @@ var marshalingTests = []struct {
 	{"BoolValue", marshaler, &pb.KnownTypes{Bool: &wpb.BoolValue{Value: true}}, `{"bool":true}`},
 	{"StringValue", marshaler, &pb.KnownTypes{Str: &wpb.StringValue{Value: "plush"}}, `{"str":"plush"}`},
 	{"BytesValue", marshaler, &pb.KnownTypes{Bytes: &wpb.BytesValue{Value: []byte("wow")}}, `{"bytes":"d293"}`},
+	{"BytesMessageValue", marshaler, &BytesMessage{Bytes: []Byte("wow")}, `{"bytes":"d293"}`},
 }
 
 func TestMarshaling(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tamird:
cc @awalterschultze

awalterschulze:
Yes it doesn't seem like this code will break any previous assumptions it just allows a very cases.

jsonpb/jsonpb_test.go
```diff
@@ -282,6 +282,18 @@ func init() {
 	}
 }
 
+// Bytes is used to test that []byte type aliases are serialized to base64.
+type Bytes []byte
```
tamird: this test can be made slightly more complete with:

```
// Byte is used to test that []byte type aliases are base64-encoded.
type Byte byte

// Bytes is used to test that []byte type aliases are base64-encoded.
type Bytes []Byte
```

tamird:
@matloob @dsymonds could you take a look at this?

jsonpb/jsonpb_test.go
d4l3k: Done.

d4l3k:
Any update on this? 
This is blocking https://github.com/cockroachdb/cockroach/pull/6973.

matloob:
Hi!

I'm unfamiliar with gogoprotobuf and the bug this is blocking. Could you help fill us in on some of the details?

I wasn't able to figure out how cockroachdb/cockroach#6973 uses this change. Could you provide some more information? Is there a way to unblock that change without making this change?

It would also be really helpful to get some more context on gogoprotobuf. When do the other []byte types get generated? If they're only generated by gogoprotobuf, could this change be made the gogoprotobuf fork directly?

Thanks,
Michael

tamird:
> I wasn't able to figure out how cockroachdb/cockroach#6973 uses this change. Could you provide some more information?

That change wants to use jsonpb marshalling of message which include aliases of `[]byte`, so this correct reflection code is needed to do that correctly.

> Is there a way to unblock that change without making this change?

No. This change has to be made - either here or in gogoprotobuf.

> It would also be really helpful to get some more context on gogoprotobuf. When do the other []byte types get generated?

They are generated when a field is tagged with the `gogoproto.casttype` option.

> If they're only generated by gogoprotobuf, could this change be made the gogoprotobuf fork directly?

Yes, this PR has been submitted to gogoprotobuf as well, but @awalterschulze would like to minimize drift.

I think this change is fine here -- it's subtle, but the proposed code is more correct.

matloob:
Thanks!

I think we have a better understanding of the issue here. Please correct me if we got anything wrong.

If possible, it would be ideal for the change made in gogoprotobuf since the other []byte fields are generated by the gogoprotobuf code generator. We'd prefer to avoid adding a test case on a generated field type that the proto package doesn't produce.

Michael

tamird:
I agree regarding the tests, but I don't know that there's any downside in accepting the slightly more correct reflection code. The upside is that it reduces the burden of maintainership downstream.

matloob:
We understand the desire to reduce the burden of maintainership downstream, but doing so would increase the burden of maintainership upstream.

The test in this change enforces a new invariant in how the proto package handles generated structs. The upstream protobuf library would then implicitly support the 'gogoproto.casttype' option in the 'downstream' gogoprotobuf library.

tamird:
Right, I was suggesting that the tests be backed out.

matloob:
Ah, sorry -- I misunderstood that!

In that case I don't see any issues in making this change. Let me double check, though. I'l get back to this thread tomorrow.

d4l3k:
@awalterschulze Any ETA on when https://github.com/golang/protobuf/commit/3b06fc7a4cad73efce5fe6217ab6c33e7231ab4a will get merged into gogo/protobuf?

awalterschulze:
I don't think this is the place.  Lets discuss it here https://github.com/gogo/protobuf/pull/183

PR #198, opened 2016-06-29
Make ExtensionDescs not error on empty protos

zellyn:
When called on a Message with nil XXX_InternalExtensions.p,
ExtensionDescs should not try to lock the nil Mutex returned from
extensionsRead().


zellyn:
@matloob FYI…

zellyn:
@zombiezen fyi

zellyn:
ping…

zombiezen:
We had a fix, forgot to export.  Just a minute.

Issue #196, opened 2016-06-20
regen.sh on osx


awalterschulze:
Could the regen.sh script possibly be changed to also work on osx?

```
./regen.sh: line 44: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

mwmahlberg:
Hm, first of all, as per official docs, there is [no **`-e`** option for bash ](http://linux.die.net/man/1/bash), which makes me curious. And `declare` is a bash built-in command. Every halfway recent version of bash should support it.

The funny thing is that the declare function of `sh` has options mentioned above (see manpage). So my first guess is that for some weird reason, regex.sh is run with the sh interpreter rather than bash – however that may happen.

mwmahlberg:
As a matter of fact, it might be just the opposite around. Not sure, but changing the shebang to

    #!/bin/sh

might actually solve the problem.

awalterschulze:
I still get the same error

mwmahlberg:
The same one? Ok, I will look into it to get it running. Any way I can verify the results?

awalterschulze:
I tried this regen diff
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/sh -e
 #
 # This script fetches and rebuilds the "well-known types" protocol buffers.
 # To run this you will need protoc and goprotobuf installed;
@@ -39,7 +39,7 @@ echo 1>&2 "base: $base"
 cd $base

 echo 1>&2 "fetching latest protos... "
-git clone -q $UPSTREAM $tmpdir
+git clone $UPSTREAM $tmpdir
 # Pass 1: build mapping from upstream filename to our filename.
 declare -A filename_map
 for f in $(cd $PKG && find * -name '*.proto'); do
```

I have this declare
```
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

Don't you have OSX?

mwmahlberg:
Aye, I am already working on it. It is a bit strange. Not sure what happens there.

mwmahlberg:
[...]

PR #200, opened 2016-07-01
issue 196 ptypes/regen.sh

mwmahlberg:
I have changed the regen.sh script as per issue #196. Only later I found out about #144. I took the comments made there into account, where feasible.

Tests were run on Linux (CentOS 6.7, go1.6.2 linux/amd64) and OS X(10.11.5 "El Capitan", go1.6.2 darwin/amd64).
# Overview of changes
- The unused `PROTO_FILES`variable was removed
- `PKG`,`UPSTREAM` and `UPSTREAM_SUBDIR` can now be configured by setting the respective `PTYPES_REGEN_*` environment variable. If they are not set or empty, the current values are used as a default.
- `find` was added to the tool sanitation
- `git clone` is not silent any more to give proper feedback
- the `basedir` lookup was simplified to the static `$GOPATH/src/$PKG`
- Upstream now provides `option go_package`, so the mangling of package names became unnecessary , hence it was removed
- The output directory of `protoc` was set to `$GOPATH/src`, as full package paths are generated now.
- Enhanced portability
  -  Instead of using the `declare` statement, which caused portability problems on OS X, the search for `.proto` files is done again for the second pass, again using `find`. Since we are sure that the environment is sane as of pass 1 not dying, we can do this and it is more portable and reasonably cheap.
  - The dependency on sed was removed entirely, as it is a good candidate for portability problems. The unproblematic `expr` or shell substitution patterns are used instead, where necessary.
  - There still is some reliance on GNU coreutils and findutils, but those are known to be unproblematic

### DIFF
--- a/ptypes/regen.sh
+++ b/ptypes/regen.sh
@@ -6,7 +6,7 @@
 
 set -Ee
 
-PKG=github.com/gogo/protobuf/ptypes
+PKG=github.com/golang/protobuf/ptypes
 UPSTREAM=https://github.com/google/protobuf
 UPSTREAM_SUBDIR=src/google/protobuf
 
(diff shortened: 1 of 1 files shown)


awalterschulze:
Yes please :)

bmkessler:
@dsymonds Any updates on this or #144 to fix the scripts for OS X?

paranoiacblack:
Seems like #196 is fixed in #412. Please re-open if this is not true.

PR #202, opened 2016-07-13
Update package name to 'main'

Mistobaan:
I see no benefit in having another package name. I found this trying to build the package without go build and just with go tool compile


paranoiacblack:
Hmm, I'm not sure about the idiom here, but using `go build` instead of `go tool compile` isn't expected to work on this package or many others. Closing this unless there's some particular reason to permit it.

PR #207, opened 2016-07-28
Add TimestampNow and TimestampFormat funcs

technosophos:
This adds two timestamp functions to the pytpe library. TimestampNow()
returns a protobuf Timestamp with the current time. TimestampFormat()
takes a Go time format string and formats a protobuf Timestamp
accordingly. With TimestampFormat here, TimestampString() was refactored
to use that function.

Additionally, one documentation fix was made to the Timestamp()
function.

### DIFF
--- a/ptypes/timestamp.go
+++ b/ptypes/timestamp.go
@@ -85,8 +85,7 @@ func validateTimestamp(ts *tspb.Timestamp) error {
 // locale. This may or may not be a meaningful time; many invalid Timestamps
 // do map to valid time.Times.
 //
-// A nil Timestamp returns an error. The first return value in that case is
-// undefined.
+// A nil Timestamp returns the UNIX epoch time (not Go's empty time).
 func Timestamp(ts *tspb.Timestamp) (time.Time, error) {
 	// Don't return the zero value on error, because corresponds to a valid
 	// timestamp. Instead return whatever time.Unix gives us.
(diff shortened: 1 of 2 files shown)


technosophos:
Relates to https://github.com/kubernetes/helm/issues/914

steeve:
Is there a way we could have them in the `ptypes/timestamp` packages, even better attached to `*Timestamp` ?

technosophos:
I don't mind refactoring to that goal. But the convention in the package seems to be to put all such time functions here. I'm not too familiar with the leadership structure of the project. Who's qualified to make a definitive statement on where this feature belongs?

zombiezen:
It would belong in ptypes.  Sorry for the delay in looking at this PR; we have a long backlog of proto requests.  I'm still thinking about this one: can you help me understand what this change helps you do that you couldn't before?

cybrcodr:
Hi, thanks for sending this PR.

TimestampNow has been implemented.

While I can understand that TimestampFormat provides a convenience wrapper around time.Time.Format, it still relies on callers knowing the format layout of the method Time.Format.  And hence I hesitate extending the API surface for that.

I also understand that there is already the existing TimestampString and that TimestampFormat can be an improved version, but then again I think that one should probably be marked deprecated instead for similar reason as above and also that it returns error in a string instead of a real string. It also seems like it has really low usage at least from querying github.

Closing this.  Re-open if you feel strongly different.

Issue #166, opened 2016-04-13
jsonpb Unmarshalling support for Struct and Value wkt


rohitsaboo:
Do we have support in jsonpb for unmarshalling Struct and Value wkt? If not, is there any plan to add any?

dsymonds:
Yes, Struct and Value are supported by jsonpb. What are you missing?

rohitsaboo:
For some reason, I'm seeing error messages of the form, 'unknown field "xxx" in structpb.Struct/Value',
and the fields in the protobufs are empty.

Also, I cannot find a unit test that is unmarshalling to be sure I'm doing things correctly.

dsymonds:
The spec is at https://developers.google.com/protocol-buffers/docs/proto3#json, mentioned in the jsonpb package docs. You could also just marshal a populated structpb.Struct to see what the format is meant to look like.

Feel free to open a new issue if you find a bug.

rohitsaboo:
Hey David, I'm looking to unmarshal from json and not marshal. There are test cases for marshaling structs, but none for unmarshalling.

dsymonds:
You indicated you didn't know whether you were doing things right, which I took to mean you weren't sure what a JSON format Struct is meant to look like. I suggested marshaling from jsonpb to see what it looks like.

If you have specific input data that isn't parsed, show it. I can't help you with anything else. The code in jsonpb works to the best of my knowledge.

rohitsaboo:
Hey David,

Here's a somewhat minimal example:

The proto looks like
```
syntax = "proto3";

package test;

import "google/protobuf/struct.proto";

message Activity {
  Person actor = 3;
}

message Person {
  google.protobuf.Value location = 9;
}
```

The json looks like:
```
{
    "actor": {
        "location": {
            "objectType": "place",
            "displayName": "Denver, CO"
        }
    }
}
```

The unmarshalled proto comes out to
```
actor: <
  location: <
  >
>
```
with the error message `unknown field "objectType" in structpb.Value`

dsymonds:
Yes, and what happens in the other direction? i.e. start with the proto that you expect, and see what jsonpb produces.

rohitsaboo:
We get the correct json in the other direction (with the indentation and ordering as below):
```json
{
  "actor": {
    "location": {
          "displayName": "Denver, CO",
          "objectType": "place"
        }
  }
}
```

The code to populate the proto looks a bit ugly, but here it is:
```go
	a := test.Activity{
[...]

PR #208, opened 2016-07-30
Add unmarshal support for the Struct and Value WKTs.

AmandaCameron:
Since I saw no progress on #166 I decided to code up a CL for myself to do this.

Some things to note:
- Tests were addes, however this required disabling some of the map tests, which were causing a build failure, and I felt this was out-of-scope to fix for this CL.
- I'm not overly familiar with reflect code, so I might be doing something Bad™ in the code for this all.


AmandaCameron:
@zombiezen Any chance this could be looked at sometime soon? I'm currently stuck using my fork until this is merged as I need to do JSON deserialization into the Struct WKT for one of my projects.

zombiezen:
jsonpb needs work in general (it fails large swaths of the protobuf conformance test at the moment) and I think it will be less work overall to try to revamp the package wholecloth than trying to solve it piecemail.  That doesn't leave this PR in a good state, but we don't have a good answer for jsonpb in general at the moment.  Stay tuned.

AmandaCameron:
I've been hacking on the jsonpb stuff today, and if it'd be agreeable, I'd be happy to submit what I've done, and plan to do, to fix the conformance, at least regarding the conformance tests involving `JsonInput.*` tests.

The additional commit handles several of the number types, properly, according to the conformance tests, consiquently, it's brought down the number of failures from 91 to 56. I believe a large portion of the conformance issues stem from the go idiomatic ways differing from the way the spec expects. One low-hanging fruit I noticed was that `time.ParseDuration` doesn't expect decimal values from it's units, which the conformance spec does.

jsonpb/jsonpb.go
```diff
@@ -825,13 +838,99 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
...
+				target.Set(reflect.ValueOf(int64(tmp)))
+
+			case reflect.Float32:
+				target.Set(reflect.ValueOf(float32(tmp)))
+
+			case reflect.Float64:
+				target.Set(reflect.ValueOf(float64(tmp)))
+			}*/
+
+		var tmp float64
```
AmandaCameron: Of note here: The use of float64 is causing loss of precision for int64 values. The proper solution would be an exponent parser that outputs int64s instead of floats, as the `strconv.ParseInt` doesn't handle exponents.

atombender:
Can we please get this merged? It's not possible to use `Struct` or `Value` without it, in conjunction with grpc-gateway.

kellrott:
I've tested this with my project, it works, and it fixes the problem that I was having with the master branch.
Its been a few months since there has any movement on this. Is there something blocking it from being merged?

kyessenov:
Is there a timeline for when the underlying issue with protobuf `struct` going to be fixed for golang?
The issue has been open for quite a while and is blocking client libraries for any API that chooses to use `struct`.

kellrott:
@zombiezen are you still tracking this PR? 
Unfortunately grpc-gateway generated code links to golang/protobuf, and this bug breaking full JSON struct serialization. From this thread, it looks like this bug has blocked multiple projects. I'm curious what hack other people have used to get around this problem.

In your comment, the main reason for not accepting this PR was that jsonpb was going to be re-written. That was over 6 months ago, so it doesn't seem like that is going to happen. Can we just get this PR merged?

cybrcodr:
Done in commit 18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8.

PR #209, opened 2016-08-02
jsonpb: fix Unmarshal when single enum value is provided for a repeated enum field

maditya:
…d enum field

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -634,6 +634,9 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			target.Set(reflect.New(targetType.Elem()))
 			target = target.Elem()
 		}
+		if target.Kind() != reflect.Int32 {
+			return fmt.Errorf("mismatched type, want %v, got %v", reflect.Int32, target.Kind())
+		}
 		target.SetInt(int64(n))
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


maditya:
ping @bcmills

jsonpb/jsonpb.go
```diff
@@ -634,6 +634,9 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			target.Set(reflect.New(targetType.Elem()))
 			target = target.Elem()
 		}
+		if target.Kind() != reflect.Int32 {
```
hyangah: Instead this, in line 624, you can check whether the target can be an enum type.

maybeEnum := target.Kind() == reflect.Int32 ||
  target.Kind() == reflect.Ptr && targetType.Elem().Kind() == reflect.Int32  // proto2
if maybeEnum && inputValue[0] == '"' && prop != nil && prop.Enum != "" {
 ...
}

The advantage is the case this change addresses will be handled by the json unmarshal code so the error message will be a bit clearer.

"json: cannot unmarshal string into Go value of type []json.RawMessage"
vs. 
"mismatched type, want int32, got slice"

cybrcodr:
I'd like to follow up on this PR.  It would be nice to have this merged in if you can follow through the review comment.  Thanks.

dsnet:
Apparently this is a duplicate of #627, which was recently merged.

PR #213, opened 2016-08-09
jsonpb: fixing unmarshaling enum from int

goloveychuk:
Now it's unsafe, you can easily write any value and it will swallow.

### DIFF
--- a/proto/lib.go
+++ b/proto/lib.go
@@ -421,14 +421,14 @@ func EnumName(m map[int32]string, v int32) string {
 // value, it returns an int32 that can be cast to the enum type by the caller.
 //
 // The function can deal with both JSON representations, numeric and symbolic.
-func UnmarshalJSONEnum(m map[string]int32, data []byte, enumName string) (int32, error) {
+func UnmarshalJSONEnum(names map[int32]string, values map[string]int32, data []byte, enumName string) (int32, error) {
 	if data[0] == '"' {
 		// New style: enums are strings.
 		var repr string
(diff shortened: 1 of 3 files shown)


goloveychuk:
I signed it!

goloveychuk:
I signed it!

zombiezen:
Can you explain more what this does?  This changes the signature of an exported function for reasons that aren't clear to me.

goloveychuk:
if unmarshalling old style enum value (represented by ints) - it wont be validated. You can unmarshall 
```json
{
     type: 32
}

while you having 
enum Type {
  a = 1;
  b = 2;
}
, and it will not return a error, so it's not secure to use this function

dsymonds:
That's by design. The same thing happens when parsing from the binary format.

dsymonds:
That's by design. The same thing happens when parsing from the binary format.

dsymonds:
That's by design. The same thing happens when parsing from the binary format.

goloveychuk:
@dsymonds 
by binary you mean serialised by protobuf or something like jsonb?
if first, there is a big difference between trusted services communication and function-helper unmarshaller from json.

afking:
@goloveychuk I've made a similar change focusing on JSONPb. I think the function of marshaling enums should be `func UnmarshalJSONEnum(data []byte, enumName string) (int32, error)` as the maps are available by calling `proto.EnumValueMap(enumName)`.

dsnet:
The JSON<->ProtoBuf spec allows unmarshaling an enum value that does not map to an known name. So this is working as intended.

PR #224, opened 2016-08-29
Add bazel BUILD files for protoc-gen-go and the proto library.

pmbethe09:


awalterschulze:
This is a lot of code, to add a dependency on java, to build a go project.

pmbethe09:
Fair point, but:
* bazel is Google's official build system, and starting to get more
open-source adoption
* new_go_repository will soon be in https://github.com/bazelbuild/rules_go,
but it can't yet handle the selection of pointer_reflect.go vs
pointer_unsafe.go ("//build +appengine"); so explicit BUILD files are needed
* This is only for folks using bazel and pulling in a WORKSPACE dependency
on https://github.com/golang/protobuf and won't effect pure "go build"
people.

Also, I volunteer to maintain these files for future changes.

goloveychuk:
@dsymonds please look at this pr, or invite somebody, thanks

zombiezen:
@goloveychuk I've been discussing this with @pmbethe09 off-thread.  There's a couple of extra details that we're juggling, but this is moving forward.

goloveychuk:
@zombiezen great sorry, I missed thread, I was talking about #213

hochhaus:
FWIW, `rules_go` just [added](https://github.com/bazelbuild/rules_go#new_go_repository) `new_go_repository()` which makes using `golang/protobuf` as easy as adding something like the following to your `WORKSPACE`.

```
new_go_repository(
    name = "com_github_golang_protobuf",
    commit = "1f49d83d9aa00e6ce4fc8258c71cc7786aec968a",
    importpath = "github.com/golang/protobuf",
)
```

pmbethe09:
Yes/no.  new_go_repository is great, but only for pure go directories.

Note here there are .proto files which we want to expose via a filegroup,
e.g. ptypes/empty/empty.proto

I suppose we could try and work around that, by asking gazelle to follow a
convention of exposing .proto files if found (or something configurable)

hochhaus:
> Note here there are .proto files which we want to expose via a filegroup,
e.g. ptypes/empty/empty.proto

Good point. As you suggest I'm only using the pure-code code, not the `*.proto` files.

pmbethe09:
A small hack to rules_go's "gazelle", and it appears we can now use new_go_repository without needing this PR, so closing.

Issue #218, opened 2016-08-21
types/known: one home for the WKTs (Well Known Types)


awalterschulze:
How are the .pb.go files going to end up in one home.
Continuing this discussion in https://github.com/golang/protobuf/issues/143

Ok so the idea is that all the code generated in ptypes will move to https://github.com/google/go-genproto
So each user of proto3 wkts will now also have to import this new package.

Also unlike ptypes which has each proto and pb.go in its own package go-genproto has all the protos and .pb.go files in one package called protobuf.

What is the eventual plan for this?
Is moving all these pb.go files into one home really depending on this very contentions issue:
https://golang.org/issue/16339
Or is there a backup plan?

What would you have done if you could have done this from scratch?

zombiezen:
Adding type aliases is not a hard blocker for moving WKT to one home, but it certainly would help.  I mention it to raise awareness of the problems it solves.  If we had type aliases, instead of rolling out one breaking change after six months of announcement (as per our new [compatibility promise](https://github.com/golang/protobuf/blob/master/README.md#compatibility)), we could make the announcement and folks could migrate their import paths well ahead of the deadline and not face issues.  Both options have their own tradeoffs.

An alternative approach would be to just add the generated protos in go-genproto repo &mdash; this would mean they exist in both places.  This would be unfortunate because if part of a large codebase uses the type as it exists in go-genproto and another part of a large codebase uses the type as it exists in ptypes, the types would not be interoperable.  We've rejected this approach because it's confusing to end-users and the path to resolution is not obvious.

@okdave and I are still thinking about this.  If we were starting from scratch, we would have them all in the new go-genproto repository.  That is the direction we want to move toward.  The implementation plan is still fuzzy at the moment, but we agree it's a problem that should be tackled.

awalterschulze:
Interesting.  Thanks for you explanation.  I really appreciate it :)

I guess I'll still have to make my own fork of gen-proto and then it might as well live in gogoprotobuf/types.

Either way the user of protoc will always have to use a lot of import path remapping with the `M` parameter.  Do you see any way around this in future?

Do you know why the googleapis folder also has to live in go-genproto
It seems like to things that should probably be kept separate?

dsnet:
Fixed with https://github.com/protocolbuffers/protobuf-go/commit/81db48ad09ccc6ee1fcae434baa4bb1b7a7a7a72

PR #227, opened 2016-09-02
Generate field_mask.proto WKT.

### LINKED ISSUE #226: FieldMask is missing from ptypes directory.
I see https://github.com/google/protobuf/blob/master/src/google/protobuf/field_mask.proto is missing a go_package identifier and it does not appear in the ptypes directory with the other common proto files.

This file is needed for me to write the conformance tests in https://github.com/google/protobuf/tree/master/conformance and surely the FieldMask type would be needed in some cloud APIs.

tswast:
Upstream change is https://github.com/google/protobuf/pull/2058.

Fixes #226.


dsymonds:
That doesn't look like the right output filename.

tswast:
Oops. Yep, you're right. Fixed now.

zombiezen:
Note that `FieldMask` already exists in [google.golang.org/genproto/protobuf](https://godoc.org/google.golang.org/genproto/protobuf).  See also #218.

tswast:
@zombiezen Thanks for pointing that out. I'll read through that discussion. Possibly I'll need to change my upstream change to point to the genproto package name.

PR #232, opened 2016-09-14
Emit empty array in JsonPB if EmitDefaults is enabled

zackangelo:
Forces an empty-valued array field to be serialized if `EmitDefaults` is set to `true`. 

Mirrors Java's protobuf to JSON implementation when `includingDefaultValueFields` is enabled.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -180,7 +180,7 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
 
 		// IsNil will panic on most value kinds.
 		switch value.Kind() {
-		case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Ptr, reflect.Slice:
+		case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Ptr:
 			if value.IsNil() {
 				continue
 			}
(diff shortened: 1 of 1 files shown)


zackangelo:
@dsymonds @zombiezen *nudge* :)

zombiezen:
LGTM from a casual glance.  @bcmills, thoughts?

bcmills:
Unless there are conflicts with [the proto3 spec](https://developers.google.com/protocol-buffers/docs/proto3#json), the behavior of `jsonpb`  should arguably mimic the standard `json` package.  (`EmitDefaults: false` should correspond to `omitempty`.)

By my reading of the spec, the non-`EmitDefaults` behavior is currently out-of-spec:
> The default value for repeated fields is empty (generally an empty list in the appropriate language).
> [...]
> If a field has the default value in the protocol buffer, it will be omitted in the JSON-encoded data by default to save space.

We do not appear to omit empty slices when `EmitDefaults` is `false`, but we probably should.  (Perhaps that's a separate issue?)

That doesn't give us much guidance for `nil` slices when `EmitDefaults` is `true`, though.  I would argue that we should follow the `json` package and emit a `null` in this case rather than an empty list:
https://play.golang.org/p/r6LBqu4mLI

Per the proto3 spec, "null is accepted as the empty list []", and that would remove another inconsistency between `jsonpb` and `json`.

hyangah:
I agree with @bcmills on emitting null for an empty list. Any objection?

@zackangelo can you update your PR and also add a test?

bcmills:
@zackangelo Are you still working on this?  Is the PR ready for another round of reviews?

cybrcodr:
I just reviewed and merged in 7a211bcf3bce0e3f1d74f9894916e6f116ae83b4.  I didn't see this PR beforehand and wasn't aware that there's a preference for null over [] from this discussion.  Note that [] uses less characters than null.  Deserialization should handle either case anyways.  Let me know if there is any strong objection to that.

PR #233, opened 2016-09-16
proto: add proto.False() and proto.True() helper routines

5:
I noticed that there are 6,800 occurrences of "proto.Bool(true)" and 3,546 occurrences of "proto.Bool(false)" in our codebase.
Having these two helper routines would make things a little bit more readable.

### DIFF
--- a/proto/lib.go
+++ b/proto/lib.go
@@ -354,6 +354,18 @@ func Bool(v bool) *bool {
 	return &v
 }
 
+// False is a helper routine that allocates a new bool value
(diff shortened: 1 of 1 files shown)


zombiezen:
While this is shorter to type, I don't see the current form as particularly unreadable.

bcmills:
Note that when using [proto3](https://developers.google.com/protocol-buffers/docs/proto3), neither of those is necessary anyway: plain old `true` and `false` suffice.

PR #235, opened 2016-09-24
jsonpb: Fix Timestamp unmarshaling for pre-epoch times

lstoll:
Go's`%` operation [returns the remainder](https://github.com/golang/go/issues/448), which can be a negative number. This was causing the nanos component for a pre-epoch time to be set to a negative, rather than the nanos component. This change checks for a negative remainder, and if found flips it round to the correct sign.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -608,11 +608,8 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			if err != nil {
 				return fmt.Errorf("bad Timestamp: %v", err)
 			}
-			ns := t.UnixNano()
-			s := ns / 1e9
-			ns %= 1e9
-			target.Field(0).SetInt(s)
-			target.Field(1).SetInt(ns)
+			target.Field(0).SetInt(int64(t.Unix()))
+			target.Field(1).SetInt(int64(t.Nanosecond()))
 			return nil
 		}
 	}
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -468,6 +468,8 @@ var unmarshalingTests = []struct {
 
 	{"Duration", Unmarshaler{}, `{"dur":"3.000s"}`, &pb.KnownTypes{Dur: &durpb.Duration{Seconds: 3}}},
 	{"Timestamp", Unmarshaler{}, `{"ts":"2014-05-13T16:53:20.021Z"}`, &pb.KnownTypes{Ts: &tspb.Timestamp{Seconds: 14e8, Nanos: 21e6}}},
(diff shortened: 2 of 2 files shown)


lstoll:
I signed it!

lstoll:
I wasn't convinced if the seconds value in my new test case should be `-1` or `-2`, so I set up a quick java test case and it seems like it has the same bad handing of pre-epoch times.

```
Date testDate;
try {
      testDate = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSX").parse("1969-12-31T23:59:58.999Z");
} catch (ParseException e) {
    e.printStackTrace();
    return;
}

System.out.print("date: ");
System.out.println(testDate);

long millis = testDate.getTime();

Timestamp timestamp = Timestamp.newBuilder().setSeconds(millis / 1000)
        .setNanos((int) ((millis % 1000) * 1000000)).build();

System.out.println("timestamp:");
System.out.println(timestamp);
System.out.print("timestamp isValid: ");
System.out.println(Timestamps.isValid(timestamp));
```

```
date: Wed Dec 31 18:59:58 EST 1969
timestamp:
seconds: -1
nanos: -1000000

timestamp isValid: false
```

Is that broken too, or am I missing something? the [proto file](https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto#L106-L110) and the [helper](https://github.com/golang/protobuf/blob/master/ptypes/timestamp.go#L73-L75) seem to be pretty clear that is should be positive.

dsymonds:
Fixing this LGTM. Your change seems plausible.

Java might be broken, but C++ should be the canonical example. Can you try its version?

lstoll:
Yep, I'll put something together for that now

lstoll:
OK I set up a little C++ deal to verify the behavior

```
#include <iostream>
#include <google/protobuf/timestamp.pb.h>
#include <google/protobuf/util/time_util.h>
using namespace std;
using google::protobuf::Timestamp;
using google::protobuf::util::TimeUtil;

int main(int argc, char* argv[]) {
    Timestamp ts;
    TimeUtil::FromString("1969-12-31T23:59:58.999999995Z", &ts);
    cout << "ts: " << ts << "\n";
    cout << "seconds: " << ts.seconds() << "\n";
    cout << "nanos: " << ts.nanos() << "\n";
}
```

```
ts: 1969-12-31T23:59:58.999999995Z
seconds: -2
nanos: 999999995
```

I [found how it's handling this](https://github.com/google/protobuf/blob/master/src/google/protobuf/util/time_util.cc#L69-L72), and updated the commit to match it's behavior.

I think this should be good to go now. I'll open a PR for a fix for the Java implementation soon.

lstoll:
I ran in to another problem where unixnano was overflowing, turns out [`time.Nanosecond`](https://golang.org/pkg/time/#Time.Nanosecond) does what we need anyway to pushed up a new version that just uses that.

bcmills (approved):

Issue #231, opened 2016-09-07
jsonpb should serialize NaN as the string value "NaN"


tswast:
See: https://github.com/google/protobuf/issues/2086#issuecomment-245435675

This is described in the Proto3 language guide: https://developers.google.com/protocol-buffers/docs/proto3#json

> proto3 : float, double
> JSON: number
> JSON example: 1.1, -10.0, 0, "NaN", "Infinity"
> JSON value will be a number or one of the special string values "NaN", "Infinity", and "-Infinity". Either numbers or strings are accepted. Exponent notation is also accepted.

PR #242, opened 2016-10-16
Fix jsonpb to serialize NaN and Infinity as special strings

narqo:
As described in #231 float and double fields, `nan` and `(-)inf` must be serialized to strings `"NaN"`, `"Infinity"`, and `"-Infinity"`.


narqo:
I signed it!

jsonpb/jsonpb_test_proto/test_objects.proto
```diff
@@ -54,6 +54,13 @@ message Simple {
   optional bytes o_bytes = 11;
 }
 
+// Test message for holding special primitives.
+message Specials {
```
cybrcodr: Let's name this NonFinites or NonFiniteFloats to be more specific.

jsonpb/jsonpb_test_proto/test_objects.proto
```diff
@@ -54,6 +54,13 @@ message Simple {
...
+    optional float o_nan = 1;
+    optional float o_pinf = 2;
+    optional float o_ninf = 3;
```
cybrcodr: Just for completeness, can you also add double fields?

jsonpb/jsonpb.go
```diff
@@ -494,6 +494,24 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 		return out.err
 	}
 
+	// Handle special cases for float, double.
```
cybrcodr: Sorry, another nit.  Please update comment to be more specific, perhaps ...

// Handle non-finite floats like NaN, Infinity and -Infinity.

cybrcodr:
@narqo I'd like to follow up on this PR.  If I don't hear back next week, I'll submit similar fix in instead.

narqo:
Hey, @cybrcodr I've done a rebase onto master's head and fixed what you'd asked. Hope things look better now.

kesutton:
What about the unmarshaling side?  With this change, jsonpb can not successfully deserialize its own output, as it falls back to the standard json decoder due to error: `json: cannot unmarshal string into Go value of type float64`.

I think something very similar to how string encoded integers are handled here can be done: https://github.com/golang/protobuf/blob/master/jsonpb/jsonpb.go#L969

I wrote up something quick locally right after the linked section.

```
import math
...
// floats can be encoded as strings, like "NaN", "+Infinity", and "-Infinity".  In this case, we
// deserialize into the proper float64 type.  If we don't have a
// string value, we just let the standard unmarshaler parse the float
if targetType.Kind() == reflect.Float32 || targetType.Kind() == reflect.Float64 &&
	strings.HasPrefix(string(inputValue), `"`) {
	s := string(inputValue)

	if s == "\"NaN\"" {
	target.SetFloat(math.NaN())
		return nil
	}
	if s == "\"Infinity\"" {
		target.SetFloat(math.Inf(0))
		return nil
	}
	if s == "\"-Infinity\"" {
		target.SetFloat(math.Inf(-1))
		return nil
	}

	return fmt.Errorf("unknown string value %s for float", s)
}
```

I'm not sure if the check for Float32 is necessary, nor am I sure the decision to return that error message is sound.  Just posting something to get the ball rolling.

cybrcodr:
Sorry, I should have mentioned.  Fix for deserialization is coming soon.

cybrcodr:
Fyi.  Fix for deserialization in 5a0f697c9ed9d68fef0116532c6e05cfeae00e55.

PR #243, opened 2016-10-21
make files list and messages list comment deterministic.

pattyshack:


dsnet:
This comment is no longer generated, so this change is moot.

PR #244, opened 2016-10-22
Updated grpc generator to generate service desc such that metadata in…

MakMukhi:
…fo is proto filename instead of FD


MakMukhi:
Closing this. I'll send a CL instead.

PR #249, opened 2016-11-08
Allow "make test" to run with Bourne shell.

bryanpkc:
"make install; make test" did not work out of the box on Ubuntu 16.04. Removing the bashism from protoc-gen-go/testdata/Makefile allows "make test" to run with a PASS result.


bryanpkc:
I signed it!

PR #250, opened 2016-11-11
JSON unmarshaler returns UnknownFieldError error

atombender:
This makes the JSON unmarshaler return a new error struct, UnknownFieldError, for unknown fields, when `AllowUnknownFields` is `false`. This is necessary in order for callers to be able to hide Go internals in APIs. Previously, the error of `Unmarshal` would be an error string such as:

```go
unknown field "glorp" in proto.Mutation
```

This leaks the type `proto.Mutation` to clients, which should not need to see such internals when they are consuming a pure-JSON HTTP API.

There are other instances of `fmt.Errorf` being used, but this is the main one that leaks.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -510,6 +510,20 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 	return out.err
 }
 
+// UnknownFieldError is an error if an unknown field was encountered
+// during unmarshaling.
+type UnknownFieldError struct {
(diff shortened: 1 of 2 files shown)


atombender:
Hold on, I'll add a test for this struct.

dsnet:
Before adding API to distinguish between different json errors, there should first be an issue that discusses what the use-case for this even should be. After that, we can figure out what the right API is. A new error type is not necessarily the API we would want to go with.

Feel free to open an issue justifying the ability to distinguish an unknown field error.

PR #255, opened 2016-11-16
jsonpb: treat `null` JSON values for `Timestamp` and `Duration` as defaults

### LINKED ISSUE #253: Error Unmarshalling `null` Value JSON Timestamp Fields
I *think* I've found an error here.

I have a sample protobuf message:

```proto
message SampleProto {
  google.protobuf.Timestamp ts = 1;
}
```

And the following json:
```json
{
  "ts": null
}
```

When I run

```go
buf := bytes.NewBuffer([]byte(`{"ts": null}`))
var s SampleProto

err := jsonpb.Unmarshal(buf, &s)
```

I get an "invalid syntax" error, which I've tracked down to this call to [`strconf.Unquote`](https://github.com/golang/protobuf/blob/da116c3771bf4a398a43f44e069195ef1c9688ef/jsonpb/jsonpb.go#L603).

I'm not sure if this is expected behavior - according to the [official documentation](https://developers.google.com/protocol-buffers/docs/proto3#json), `null` values in `json` will be treated as zero values:

> If a value is missing in the JSON-encoded data or if its value is null, it will be interpreted as the appropriate default value when parsed into a protocol buffer. If a field has the default value in the protocol buffer, it will be omitted in the JSON-encoded data by default to save space. An implementation may provide options to emit fields with default values in the JSON-encoded output.

If the `inputString` in this scenario is `null`, should a zero-value `Timestamp` be returned instead? Also, the same goes for a `Duration` type.

mcos:
This fixes #253 by ensuring `null` `json` values for `Timestamp` and `Duration` fields are treated as `zero` values, conforming to the [official documentation](https://developers.google.com/protocol-buffers/docs/proto3#json),

> If a value is missing in the JSON-encoded data or if its value is null, it will be interpreted as the appropriate default value when parsed into a protocol buffer. If a field has the default value in the protocol buffer, it will be omitted in the JSON-encoded data by default to save space. An implementation may provide options to emit fields with default values in the JSON-encoded output.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -585,7 +585,14 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		case "Any":
 			return fmt.Errorf("unmarshaling Any not supported yet")
 		case "Duration":
-			unq, err := strconv.Unquote(string(inputValue))
+			ivStr := string(inputValue)
+			if ivStr == "null" {
+				target.Field(0).SetInt(0)
+				target.Field(1).SetInt(0)
+				return nil
+			}
+
+			unq, err := strconv.Unquote(ivStr)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


mcos:
Thanks for merging this @bcmills!

PR #257, opened 2016-11-19
Add yaml type hints

jmcfarlane:


jmcfarlane:
I signed it!

bcmills:
I don't think we want this change.  `proto` is designed for cross-language compatibility, and without a standard YAML mapping in the [proto spec](https://developers.google.com/protocol-buffers/docs/proto3) it just becomes a possibly-incompatible dialect.

On top of that, there appear not to be any tests for this change.

PR #269, opened 2016-12-18
New varint with unique values

ivan386:
In the old code, they have the same value (0)
In the new code, they have:
```
varint [0] = 0
varint [128 0] = 128
varint [128 128 0] = 16512
```

### DIFF
--- a/proto/decode.go
+++ b/proto/decode.go
@@ -67,7 +67,7 @@ func DecodeVarint(buf []byte) (x uint64, n int) {
 		}
 		b := uint64(buf[n])
 		n++
-		x |= (b & 0x7F) << shift
+		x += (b & 0xFF) << shift
 		if (b & 0x80) == 0 {
 			return x, n
 		}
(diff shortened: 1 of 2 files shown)


dsnet:
The new logic is handling varints incorrectly. I'm not sure what this is trying to achieve.

The `binary.PutUvarint` function uses the same varint encoding as protos:
```go
	var arr [10]byte
	b := arr[:binary.PutUvarint(arr[:], 16512)]
	fmt.Println(b) // [128 129 1]
```
Its output disagrees with what your change does.

ivan386:
Maybe I chose the wrong place for my proposal. This is new more compact varint. All it values is unique. MSB now used in value calc.

Maybe need to add it as new functions DecodeCompactVarint and EncodeCompactVarint.

dsnet:
This is the wrong place for this. Please submit wire format encoding proposals to the issue tracker at github.com/google/protobuf.

ivan386:
Done: https://github.com/google/protobuf/issues/4376

Issue #221, opened 2016-08-24
Support ListValue well known type


calbach:
https://developers.google.com/protocol-buffers/docs/proto3#json

atombender:
Is this issue referring to the fact that `ListValue` is serialized as `{"values": [value]}` instead of just `[value]`?

chriswhelix:
Yes, and agree.

cybrcodr:
Should be fixed by https://github.com/golang/protobuf/commit/18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8.

PR #273, opened 2016-12-29
jsonpb: properly support ListValue well known type

alanshiehclone:
golang/protobuf/jsonpb does not serialize/deserialize ListValue according to the proto3
JSON spec. ListValue is supposed to be translated into a JSON array. Without this change,
ser-des outputs ListValue as a generic protobuf type, e.g.

ListValue{Values: {Value: Value_NumberValue{NumberValue: 1}}}
=> {"Values": [1]}

Correct output should be
=> [1]


alanshiehclone:
This change addresses https://github.com/golang/protobuf/issues/221

alanshiehclone:
I signed it!

cybrcodr:
Fix already in 18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8.

PR #274, opened 2016-12-30
protoc-gen-go: fix issue with self-import check

sarietta:
Previously the go generator only checked to ensure that original package name is not self-imported. When the  option is used, this check does not cover the option-defined package. This change adds in that check to ensure that the go generator doesn't self import in addition to avoiding prepending package names to types in the generated go code for those same option-defined packages


sarietta:
I signed it!

neild:
Many apologies for letting this PR languish.

9d4962b4 changed the DefaultPackageName method to work based on the import path of the package containing a symbol. I believe that this obviates the need for this PR. Thank you, and again apologies.

PR #275, opened 2017-01-08
Update to standard library context package.

goller:
Go 1.7 added context to the standard library.  With Go 1.8 coming soon here is a PR to switch from `golang.org/x/net/context` to `context.

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -53,7 +53,7 @@ const generatedCodeVersion = 4
 // Paths for packages used by code generated in this file,
 // relative to the import_prefix of the generator.Generator.
 const (
-	contextPkgPath = "golang.org/x/net/context"
+	contextPkgPath = "context"
 	grpcPkgPath    = "google.golang.org/grpc"
 )
 


dsnet:
Thanks for the suggestions. This would break compatibility with Go 1.6 and we generally support the last two versions of Go (1.6 and 1.7). As you've mentioned, Go 1.8 release in soon (less than a month). Let's hold off on this until then, so we don't have to complicate things with build tags.

adamryman:
Go 1.8 is now out, could this be merged in?

goller:
Hey there @dsnet, checking up on this.  Do you think this is ready to merge?

zaquestion:
@dsnet or @neild (you've been active lately) Can this be merged? It is blocking TuneLab/go-truss#148 which prevents us from updating our services to the latest gokit.

bcmills:
Isn't this going to break, like, everybody on AppEngine?  (They're still on 1.6, I think.)

A proper fix for this issue is for x/net to use aliases on 1.8.typealias and 1.9.

mcastilho:
Well I guess this change got reverted.

andremarianiello:
Any explanation for why this was reverted?

dfawley:
@andremarianiello, because it doesn't work with grpc, which still imports `x/net/context`.

Changing this and grpc means breaking backward compatibility, which is important to many of our users.

Meanwhile, if you're on Go1.9, `x/net/context` is a type alias of `context`, so they work interchangeably.

If you're using an older version, Go 1.8's fix tool has a recipe for this.  `go tool fix -force context <src path>` should update everything quickly.

andremarianiello:
Thanks for the info and the workarounds.

dsnet:
There are people who are still using Go1.8. Given that we usually support the last two releases, I don't think we can reasonably roll-forward this change until all major users are on Go1.9 and above. I also don't want to have the generator produce two different build-tagged files.

PR #278, opened 2017-01-16
proto: add iota on properties const

ritoon:
simplify the way the increment is done with iota

### DIFF
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -50,12 +50,12 @@ const debug bool = false
 
 // Constants that identify the encoding of a value on the wire.
 const (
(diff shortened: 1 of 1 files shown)


dsnet:
Thanks for the Pull request, but I think the code is clearer as is. Using iota is really useful for enumerations where the specific value does not matter. However, if the constant's value is determined by something outside of Go (e.g., the protobuf spec), explicitly using the value is clearer.

See https://go-review.googlesource.com/#/c/19862/ for an example where iota was intentionally not used in the standard library.

ritoon:
ok @dsnet thanks for your quick feedback

PR #284, opened 2017-01-30
Rename the local variable name of `len` to `l`

keijiyoshida:
The local variable name of `len` shadows a reference to the built-in `len` function and that makes the built-in `len` function unusable hereafter.

To fix this issue, the local variable name of `len` is renamed to `l`.


zombiezen:
Only in that scope. There's not really an issue here.

LMMilewski:
Thank you for the pull request. As you point out, the variable name shadows the built-in name in that scope. I don't see this as a problem though. I think that the code is fine as is.

I'm going to close this pull requests, but feel free to reopen if you feel strongly about this issue.

PR #286, opened 2017-02-02
jsonpb: marshal int64 and uint64 values without quotes

reinventer:
Hello.
Sometimes it is inconvenient to parse decimal values with quotes.
I don't see cases it is need.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -499,14 +499,7 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


bcmills:
Per https://developers.google.com/protocol-buffers/docs/proto3#json, in the `int64, fixed64, uint64` row:

> 	JSON value will be a decimal string. Either numbers or strings are accepted.

The current `jsonpb` encoding for int64 appears to be correct, and the proposed change would make it incorrect according to this spec.

ECMAScript numbers are floating-point with a 53-bit mantissa.  They cannot express 64-bit integers without rounding or truncation, which would hinder compatibility between proto marshaling and ECMAScript evaluation.

Issue #293, opened 2017-02-17
proposal: proto: add API to retrieve Descriptors


jhump:
C++ and Java both provide a set of rich `Descriptor` types, that wrap the various `*DescriptorProto` messages defined in `descriptor.proto`. They provide a much easier-to-use API and do the heavy-lifting of working with descriptors: type/dependency resolution.

This enables, for example, a program to get a `ServiceDescriptor`, examine methods and corresponding request and response types (including the full request/response schema, including any nested messages from other imported files) and do something useful with them. One obvious example for using the descriptors would be an implementation of `DynamicMessage` (also provided in Java and C++ runtimes). (See #199)

Another clear use case is for dynamically interacting with a remote service that exports GRPC services that a given program does not know about (e.g. no need to generate from `.proto` files and re-compile -- instead interact with the service via reflection).

I've got a working example here:
https://github.com/jhump/protoreflect/tree/master/desc

As well as a use case, a client for GRPC service reflection:
https://github.com/jhump/protoreflect/tree/master/grpcreflect

alandonovan:
Paraphrasing my comment on #295:

This is clearly useful functionality that a great many projects want, but my experience of two Google-internal packages providing this functionality (dynamicproto and descpool) suggests a piecemeal approach can easily result in a large, cumbersome, and concurrency-hostile API.  We need a coherent design for the addition of descriptors and dynamic messages to the Go protobuf API, and there are several use cases that ought to be considered together:
- loading FileDescriptorProto messages from a variety of sources such as generated types linked into the application, .proto files in the file system, and external databases.
- converting the message representation of descriptors ("DescriptorProtos") into a cross-linked and indexed data structure ("Descriptors").
- instantiating messages of arbitrary types, not necessarily known at compile time, from a map[string]interface{} and a Descriptor, and the inverse operation of converting a message to a map.
- serializing/deserializing messages of arbitrary types to binary or text.

jgiles:
It looks like https://github.com/golang/protobuf/pull/302 by @RJPercival is on pause pending this discussion? I was about to send an equivalent PR - I'm in a spot where I need the functionality in question (descriptor.ForEnum).

[...]

PR #295, opened 2017-02-22
Add Descriptor() to Message interface

RJPercival:


LMMilewski:
This change would break everyone who implements the proto.Message interface. It not backwards compatible. I'm afraid that we can't accept this PR.

RJPercival:
I expected as much, but thought I'd go for the simplest option first. Could we add another interface that aggregates this interface plus adds that method?

LMMilewski:
Would that be different from descriptor.Message?
https://github.com/golang/protobuf/blob/master/descriptor/descriptor.go#L74

RJPercival:
No, that's exactly what I was after, thanks! I had assumed it didn't already exist since I encountered people working around the lack of that interface, but I see that it's a relatively recent addition.

RJPercival:
Ah there is one thing missing: an `EnumMessage` interface, e.g.
```
type EnumMessage interface {
  EnumDescriptor() ([]byte, []int)
}
```

RJPercival:
What do you think of PR #302?

alandonovan:
Has anyone published a coherent design for the addition of descriptors and
dynamic messages to the Go protobuf API?  It's clearly useful functionality
that a great many projects want, but my experience of two Google-internal
packages that add this functionality (dynamicproto and descpool) suggests
that a piecemeal approach can easily result in a large, cumbersome, and
concurrency-hostile API.

There are several use cases that ought to be considered together:
- loading FileDescriptorProto messages from a variety of sources such as
generated types linked into the application, .proto files in the file
system, and external databases.
- converting the message representation of descriptors ("DescriptorProtos")
into a cross-linked and indexed data structure ("Descriptors").
- instantiating messages of arbitrary types, not necessarily known at
compile time, from a map[string]interface{} and a Descriptor, and the
inverse operation of converting a message to a map.
- serializing/deserializing messages of arbitrary types to binary or text.

RJPercival:
I don't have any involvement with either of those packages. This is actually the first change I've made to protobuf APIs, and I did so only because I encountered a number of Google-internal packages doing their own thing and saw that filling in this small gap in the descriptor API would be helpful.

However, from what I've seen so far, I certainly agree that a well thought-out design for where descriptors and dynamic messages should go from here is called for. I think issue #293 is probably the best place to have that discussion.

PR #299, opened 2017-02-28
Fix Properties.String() bug

jlburkhead:
obvious typo in `Properties.String` method is leaving out the `Wire` field.


jlburkhead:
I signed it!

tamird:
Also fixed in #316.

PR #302, opened 2017-03-07
descriptor: add Enum interface and ForEnum function

RJPercival:
Similar to descriptor.Message and descriptor.ForMessage, except for enums.


dsnet:
I apologize for the immensely delayed response; we're working through the backlog of protobuf issues. Thanks for the PR. At this present moment, there is a re-design of much of the proto ecosystem happening as a result of #364. The new API will continue comprehensive ways to obtain the descriptors, so I'm wary of adding additional ways to provide that same functionality.

PR #306, opened 2017-03-08
Correctly support JSON encoding/decoding of the Struct well known type

ateirney:
The pull request aims to enhance support for the well known "Struct" type.

Marshalling of the ListValue type resulted in an object being output, this has been corrected to be marshalled as a JSON array.

Unmarshalling of the Struct/Value/ListValue type(s) has also been implemented.

My experience in go is somewhat limited, and even more so with the internal of golang/protobuf so there are bound to be changes required.


ateirney:
Closing this particular pull request as there appear to already be a number of pull requests for similar functionality.

PR #308, opened 2017-03-11
Append comma to wire type name in String()

asehra:
Although String only seems to be used only in debugging, it is probably helpful to see the `p.Wire` property in the String representation. If not, maybe get rid of the assignment.


asehra:
I signed it!

asehra:
Just noticed that this is a duplicate of #299. Closing.

PR #316, opened 2017-03-17
Correct some mistakes

tamird:
Most notable is the bug in (*proto.Properties).Parse, for which I've
added a test.

Found with honnef.co/go/tools/cmd/staticcheck.

	proto/properties.go:218:2: this value of s is never used (SA4006)
	proto/properties.go:329:5: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
	protoc-gen-go/generator/generator.go:1686:5: this value of obj is never used (SA4006)


tamird:
@cybrcodr could you have a look please?

tamird:
Updated to merge to dev rather than master and rebased to fix all detected issues.

jsonpb/jsonpb.go
```diff
@@ -974,11 +974,12 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		if mp != nil {
 			target.Set(reflect.MakeMap(targetType))
 			var keyprop, valprop *proto.Properties
+			//lint:ignore SA9003 These could still be nil if the
```
dsnet: Revert. Please don't add lint annotations to the source code.

dsnet: If we really want to silence the lint check, just comment out the if statement.

tamird: Can you explain the objection? I was imagining that adding megacheck to CI would be the desired outcome here, and this comment is necessary for staticcheck-clean (which is a part of megacheck).

tamird: Commenting out the check is slightly worse, since it'll allow the code to rot.

Anyway, tell me what you prefer and I'll update this when I get home. Any other comments?

tamird:
Ping - could you please look at this? There's a bugfix here, too.

jsonpb/jsonpb.go
dsnet: As a language, Go tends to favor less compiler and tooling hints and more of the tools "just doing the right thing". Of course, tools will never be able to determine programmer intent correctly all the time, so false-positives exists.

Some programmers are happy with tooling hints in their code and that's fine with them; it is really up to the owner of a given code base. In the protobuf repo, we don't want any tooling hints in our source code and that's a reasonable stance for us to hold as well.

Please make the change. Either comment out the if-statement or delete it entirely.

tamird: Done.

jsonpb/jsonpb.go
```diff
@@ -973,13 +973,6 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		}
 		if mp != nil {
 			target.Set(reflect.MakeMap(targetType))
-			var keyprop, valprop *proto.Properties
-			if prop != nil {
-				// These could still be nil if the protobuf metadata is broken somehow.
-				// TODO: This won't work because the fields are unexported.
```
dsnet: Please keep remove the TODO comment. It's still a valid work to do.

tamird: Done (reworded so it makes sense now that the anchoring code is gone).

tamird: Note that the new comment follows the example of several other TODOs in this file.

PR #325, opened 2017-03-31
support custom JSON (de)serialization by delegating to message if it implements json.Marshaler/json.Unmarshaler

jhump:
This is achieved by delegating to the message if it implements `json.Marshaler` or `json.Unmarshaler`.

FWIW, this could simplify how the well-known types are handled (e.g. no need to special-case them in this package).

Both the binary and text encodings support the same concept, where a message can implement a particular interface (`proto.Marshaler`/`proto.Unmarshaler`, `encoding.TextMarshaler`/`encoding.TextUnmarshaler`) to customize serialization/de-serialization. This just extends that same concept to JSON encoding/decoding.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -102,6 +102,15 @@ type wkt interface {
 
 // marshalObject writes a struct to the Writer.
 func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeURL string) error {
+	if jsm, ok := v.(json.Marshaler); ok {
+		b, err := jsm.MarshalJSON()
+		if err != nil {
+			return err
+		}
+		out.write(string(b))
+		return out.err
+	}
+
 	s := reflect.ValueOf(v).Elem()
 
 	// Handle well-known types.
@@ -527,6 +536,9 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
 	inputValue := json.RawMessage{}
 	if err := dec.Decode(&inputValue); err != nil {
 		return err
+	}
+	if jsu, ok := pb.(json.Unmarshaler); ok {
+		return jsu.UnmarshalJSON([]byte(inputValue))
 	}
 	return u.unmarshalValue(reflect.ValueOf(pb).Elem(), inputValue, nil)
 }
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -74,6 +74,22 @@ type Marshaler struct {
 	OrigName bool
 }
 
+// JSONPBMarshaler is implemented by protobuf messages that customize the
+// way they are marshaled to JSON. Messages that implement this should
+// also implement JSONPBUnmarshaler so that the custom format can be
+// parsed.
+type JSONPBMarshaler interface {
+	MarshalJSONPB(*Marshaler) ([]byte, error)
+}
+
+// JSONPBUnmarshaler is implemented by protobuf messages that customize
+// the way they are unmarshaled from JSON. Messages that implement this
+// should also implement JSONPBMarshaler so that the custom format can be
+// produced.
+type JSONPBUnmarshaler interface {
+	UnmarshalJSONPB(*Unmarshaler, []byte) error
+}
+
 // Marshal marshals a protocol buffer into JSON.
 func (m *Marshaler) Marshal(out io.Writer, pb proto.Message) error {
 	writer := &errWriter{writer: out}
@@ -102,8 +118,8 @@ type wkt interface {
 
 // marshalObject writes a struct to the Writer.
 func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeURL string) error {
-	if jsm, ok := v.(json.Marshaler); ok {
-		b, err := jsm.MarshalJSON()
+	if jsm, ok := v.(JSONPBMarshaler); ok {
+		b, err := jsm.MarshalJSONPB(m)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


cybrcodr:
I had some discussion with @zombiezen on the approach for this.  We have a concern with the use of json.Marshaler or json.Unmarshaler as interfaces for doing custom serialization or deserialization because the jsonpb.Marshaler (or Unmarshaler) options would not be passed on and hence ignored.  This means that if a dynamic message contains other non-dynamic proto messages or WKT, it would not be able to consistently use the same options in (de)serializing those fields/subfields.

How about defining the following interfaces?

```
type JSONPBMarshaler interface {
    MarshalJSONPB(Marshaler) ([]byte, error)
}

type JSONPBUnmarshaler interface {
    UnmarshalJSONPB(Unmarshaler, []byte) error
}
```

Let me know if you're amenable to our suggestion.  Thanks!

jhump:
> Let me know if you're amenable to our suggestion. Thanks!

Sure, that sounds good. I can probably make the changes this weekend. I'll ping you after I've pushed them.

jhump:
@cybrcodr, the suggested changes were quite minimal. So the patch is still nice and small.

FYI, when I tried to run the tests, one of them failed. But it's not my change as it also fails when I roll my commits back (e.g. it fails on master). Maybe some sort of CI tests should be configured to run against every pull request, to prevent such regressions? Here's the test output:

```
jsonpb_test.go:570: Any with message: can't unmarshal nested Any proto: json: cannot unmarshal string into Go value of type bool
jsonpb_test.go:570: Any with message and indent: can't unmarshal nested Any proto: json: cannot unmarshal string into Go value of type bool
FAIL	github.com/golang/protobuf/jsonpb	0.012s
```

awalterschulze:
I suspect you are using go 1.7 and not go 1.8?

jhump:
@awalterschulze correct. Should this project require 1.8? We are still using 1.7 in our prod environments (and we also have GAE code [standard runtime] code that uses protobufs and runs on Go 1.6).

awalterschulze:
Thats a question for the maintainers of this library.  I am also interested in the answer.

jsonpb/jsonpb.go
```diff
@@ -74,6 +74,22 @@ type Marshaler struct {
 	OrigName bool
 }
 
+// JSONPBMarshaler is implemented by protobuf messages that customize the
+// way they are marshaled to JSON. Messages that implement this should
+// also implement JSONPBUnmarshaler so that the custom format can be
+// parsed.
+type JSONPBMarshaler interface {
+	MarshalJSONPB(*Marshaler) ([]byte, error)
```
jhump: @cybrcodr, I wasn't sure if this struct could ever become stateful, so I have it passed by ref. Should this instead be pass-by-value? I know it currently just represents marshaling options; but it seemed like, from the name, it could at some point change to such that a pointer is the correct way to pass it.

jhump: (same goes for passing `Unmarshaler` by pointer in the other interface)

cybrcodr:
I've discussed with @dsnet regarding Go versions.  We'd like to support at least the latest 1.6 for now (till Go AppEngine default has been upgraded).  I've filed issue #356 to fix these tests.  I don't know why you're seeing different errors though, but we'll investigate further.

And yes, we agree that we need to set up CI for this project #357.

jsonpb/jsonpb.go
cybrcodr: Pass by ref sounds good.

jsonpb/jsonpb.go
```diff
@@ -528,6 +553,9 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
 	if err := dec.Decode(&inputValue); err != nil {
 		return err
 	}
+	if jsu, ok := pb.(JSONPBUnmarshaler); ok {
```
cybrcodr: How about moving this inside unmarshalValue instead?  And I think it may need to be after block that allocates memory for pointer fields, i.e. 597-600.

You can do type assertion using ...
if jsu, ok := target.Addr().Interface().(JSONPBUnmarshaler); ok { ... }

jhump: Good catch

jhump:
@cybrcodr: I addressed your comments. I did not change the existing unmarshal test: simplest way to test customized marshaling/unmarshaling is a hand-written message, which won't work with `proto.Equals` anyway (its implementation relies on struct tags in generated messages).

(Also, my current environment is Go 1.7, in which the `TestUnmarshaling`case. So it felt safest for me to leave it alone.)

PTAL

jsonpb/jsonpb_test.go
```diff
@@ -635,3 +647,41 @@ func TestUnmarshalingBadInput(t *testing.T) {
...
+		t.Errorf("an unexpected error occurred when parsing into JSONPBUnmarshaler: %v", err)
+	}
+	if msg.rawJson != rawJson {
+		t.Errorf("message contents not set correctly after unmarshalling JSON\nExpected: %s\nGot: %s", rawJson, msg.rawJson)
+	}
+}
+
+// Implements protobuf.Message but is not a normal generated message type. Provides
+// implementations of JSONPBMarshaler and JSONPBUnmarshaler for JSON support.
+type notGeneratedMessage struct {
```
cybrcodr: Optional.  How about just naming this as dynamicMessage?

jsonpb/jsonpb_test.go
```diff
@@ -439,6 +439,18 @@ func TestMarshaling(t *testing.T) {
...
 
+func TestMarshalingWithJSONPBMarshaler(t *testing.T) {
+	rawJson := `{ "foo": "bar", "baz": [0, 1, 2, 3] }`
+	msg := notGeneratedMessage{rawJson: rawJson}
+	str, err := new(Marshaler).MarshalToString(&msg)
+	if err != nil {
+		t.Errorf("an unexpected error occurred when marshalling JSONPBMarshaler: %v", err)
+	}
+	if str != rawJson {
+		t.Errorf("Marshalling JSON produced incorrect output\nExpected: %s\nGot: %s", rawJson, str)
```
cybrcodr: A couple of nits.
(1) Start with lowercase for error message for consistency, i.e. "marshalling ...".  
(2) display got first before want.
"marshalling JSON produced incorrect output: got %s want %s"

jhump: done

jsonpb/jsonpb_test.go
jhump: done

jhump:
@cybrcodr - hopefully it's ready to merge now

Thanks for reviewing!

cybrcodr:
Fyi.  I just discovered issue #360 related to this.  Let me know if you're interested in providing a fix, else, I'll try to give it a shot next week.  Thanks.

PR #326, opened 2017-03-31
Revert "Update to standard library context package."

dsnet:
Reverts golang/protobuf#275

This is causing breakages for users. Will investigate more in-depth tomorrow.

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -53,7 +53,7 @@ const generatedCodeVersion = 4
 // Paths for packages used by code generated in this file,
 // relative to the import_prefix of the generator.Generator.
 const (
-	contextPkgPath = "context"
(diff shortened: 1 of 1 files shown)


adamryman:
Could you mention what users this is causing breakages for? Is there a plan to switch to stdlib "context" as some point in the future?

bradfitz:
Once Go 1.9 is out with type aliases, then we can do this without drama.

jsha:
Bump! Go 1.9 is out now. :-)

dsnet:
Unfortunately, the generation of context will not change until there are two releases of Go with type aliases. Thus, when Go1.10 is released, we'll make the change.

domudall:
Is there a reason for waiting until two releases of Go with type aliases as opposed to versioning one with and one without? Just curious.

bradfitz:
It'll be another 6 months or year due to App Engine stuff.

PR #329, opened 2017-03-31
Follow the new code generation comment convention

spenczar:
https://golang.org/s/generatedcode introduces a new convention for generated code. To mark a file as generated, there should be a comment with no whitespace in front of it. The comment should start with ' Code generated' and and with 'DO NOT EDIT.', all on the same line.

I've not regenerated protobuf files because I'm not confident that I would do it right. Is `make` from the repository root sufficient?


spenczar:
Hi @dsnet @zombiezen, can you take a look at this?

zombiezen:
Acknowledged, this seems reasonable.

zombiezen:
Merged and regenerated protos.

Issue #323, opened 2017-03-28
[jsonpb] Do not marshal empty slices, maps and objects


fische:
Hi,

The `Marshaler` from `jsonpb` does not marshal empty slices, maps and objects even with the `EmitDefaults` option.

For example, with following protobuf message:
```protobuf
message Message {
	message SubMessage {
		string str = 1;
	}
	repeated string arr = 1;
	map<string, string> m = 2;
	SubMessage sub = 3;
}
```

... and the following go code:
```go
func main() {
	var (
		msg       pb.Message
		marshaler = jsonpb.Marshaler{
			OrigName:     true,
			EmitDefaults: true,
		}
	)
	if s, err := marshaler.MarshalToString(&msg); err != nil {
		log.Fatalf("Could not marshal proto message : %v", err)
	} else {
		fmt.Println(s)
	}
}
```

I get:
```json
{}
```

Is that done by design? Or can/will it be fixed?

SamiHiltunen:
I'm also running into the same issue. Would it be possible to have a fix for this merged? There are multiple pull requests (#338, #232, #287, #327) for this and also some issues filed along with this one. Only difference I see with most of the related PRs whether they output both nil and empty slices as ```null```,  ```[]``` or empty slices as ```[]``` and nil slices as ```null``` The proto3 JSON spec seems to accept both ways of handling as a valid solution although I would go for representing omitted/empty maps and slices always as ```[]``` and ```{}```. Maybe it would make sense to have a configuration option for always forcing the empty/omitted values to be outputted as empty values rather than nulls?

AkihiroSuda:
PING @dsnet @nelid @dsymonds

AkihiroSuda:
ping ping...

dsnet:
\cc @cybrcodr

PR #287, opened 2017-02-02
jsonpb: Honor EmitDefaults for nilable types

timstudd:

PR #338, opened 2017-04-13
Update jsonpb to handle Map, Slice, Ptr nil values for EmitDefaults

ewang:
My attempt at making the Map, Slice, nil Ptr type marshaling match the proto3 JSON spec.

Fixes: https://github.com/golang/protobuf/issues/323

Related PRs:
https://github.com/golang/protobuf/pull/232
https://github.com/golang/protobuf/pull/287


ewang:
@cybrcodr Rebased

tamird:
@cybrcodr @ewang this commit seems to have the unfortunate effect of `(*string)(nil)` becoming `&""` when marshalled and unmarshalled. Specifically, we have this [proto](https://github.com/cockroachdb/cockroach/blob/aa22f21/pkg/build/info.proto#L41):
```protobuf
syntax = "proto2";
package cockroach.build;
option go_package = "build";

import "gogoproto/gogo.proto";

message Info {
  optional string go_version = 1 [(gogoproto.nullable) = false];
  optional string tag = 2 [(gogoproto.nullable) = false];
  optional string time = 3 [(gogoproto.nullable) = false];
  optional string revision = 4 [(gogoproto.nullable) = false];
  optional string cgo_compiler = 5 [(gogoproto.nullable) = false];
  optional string platform = 6 [(gogoproto.nullable) = false];
  optional string distribution = 7 [(gogoproto.nullable) = false];
  optional string type = 8 [(gogoproto.nullable) = false];

  optional string dependencies = 10000;
}
```

Here's my test case:
```golang
func TestProtobuf(t *testing.T) {
	var b bytes.Buffer

	var in build.Info
	if err := (&jsonpb.Marshaler{EmitDefaults: true}).Marshal(&b, &in); err != nil {
		t.Fatal(err)
	}
	var out build.Info
	if err := jsonpb.Unmarshal(&b, &out); err != nil {
		t.Fatal(err)
	}
	if out != in {
		t.Error(pretty.Diff(out, in))
	}
}
```
and it fails with
```
--- FAIL: TestProtobuf (0.00s)
	status_test.go:140: [Dependencies: &"" != nil]
```

cybrcodr:
Can you try out head version?  In #394, deserializing "null" for all message types except wkt ValueType should translate to nil value.

I also tried it out with a simplified example from yours.

```
syntax = "proto2";
option go_package = "optstring/pb";

message Info {
  optional string dependencies = 10000;
}
```

```
package main

import (
        "bytes"
        "fmt"
        "log"

        "optstring/pb"

        "github.com/golang/protobuf/jsonpb"
)

func main() {
        b := &bytes.Buffer{}
        p1 := &pb.Info{}
        fmt.Printf("before serializing %#+v\n", p1)
        m := jsonpb.Marshaler{EmitDefaults: true}
        if err := m.Marshal(b, p1); err != nil {
                log.Fatal(err)
        }
        fmt.Printf("serialized json: %s\n", b.String())

        p2 := &pb.Info{}
        if err := jsonpb.Unmarshal(b, p2); err != nil {
                log.Fatal(err)
        }
        fmt.Printf("deserialized into %#+v\n", p2)
        if p2.Dependencies == nil {
                fmt.Println("is nil!!!")
        }
}
```
Output...
```
before serializing &pb.Info{Dependencies:(*string)(nil), XXX_unrecognized:[]uint8(nil)}
serialized json: {"dependencies":null}
deserialized into &pb.Info{Dependencies:(*string)(nil), XXX_unrecognized:[]uint8(nil)}
is nil!!!
```

Let me know if head doesn't work out for you.

tamird:
Thanks! We're using gogo/protobuf, so I've filed https://github.com/gogo/protobuf/issues/319. I'll give it a shot once #394 is brought over.

Issue #170, opened 2016-04-15
Any type json marshaling.


AmandaCameron:
It'd be nice if the Any type could be marshaled properly by jsonpb, as currently it ends up marshaling as a couple of strings, one of which being the base64-encoded value of the type.

I recognise this might not be an easy-to-do thing (I had forked to try and get it going, but couldn't find an obvious place for it. ) however it is still wanted, so at the very least this can be a tracking bug or similar.

dsymonds:
It's just something we haven't gotten to yet. The text format supports Any messages; jsonpb just needs to be updated.

AmandaCameron:
If you can give me some hints as to where I should be hooking in, I'd be glad to try and tackle this problem, as it's something of great interest to me.

ATM, I'm mildly confused as to wether I should be poking at marshalValue or marshalObject, as neither seem particularly well-suited for types that are actually objects, instead of primitives, at least in the WKT corner.

Or am I to assume that this isn't exactly trivial in the current code layout, and that's why it's not been gotten to yet.

dsymonds:
I'll see if I can get it done today.

dsymonds:
I should have marshaling done later today. It was a bit trickier than the other well-known types.

Unmarshaling is going to be harder again; that'll be a TODO for now. Is that going to be something important for you?

AmandaCameron:
Marshaling only will be fine for now. However has there been a tracking bug for me to follow for possible updates in the future?

dsymonds:
There isn't, but feel free to open one.

PR #339, opened 2017-04-13
Implement JSON unmarshaling for Any

mmikulicic:
The Any type has already support for marshalling, but the unmarshaling of it was left unimplemented.

This PR simply implements the clearly missing feature. I tired to stay as close as possible in terms of approach and coding style as similar code in this file.

We get the test almost for free.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -558,6 +558,38 @@ func UnmarshalString(str string, pb proto.Message) error {
 	return new(Unmarshaler).Unmarshal(strings.NewReader(str), pb)
 }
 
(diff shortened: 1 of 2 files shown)


mmikulicic:
Hey @dsymonds,

I noticed you were the author of the Any marshaling part and you left unmarshaling as TODO.

This was originally tracked in #170. Should I open a new issue or can we just continue here?

cheers,
mkm

mmikulicic:
@calbach I noticed you add support for this in 18c9bb3261723cd5401db4d0c9fbc5c3b6c70fe8
which also handles unmarshaling well known types .

I believe this makes this PR no longer necessary, thus I'm abandoning it.

PR #343, opened 2017-04-17
Respect json_name option in go json tags

akrylysov:
```
message Foo {
    int64 Bar = 1 [json_name="b"];
}
```

Fixes a bug when the .proto file above doesn't respect "json_name" option and generates:

```
type Foo struct {
    Bar int64 `protobuf:"varint,1,opt,name=Bar,json=b,proto3" json:"Bar,omitempty"`
}
```

instead of:

```
type Foo struct {
    Bar int64 `protobuf:"varint,1,opt,name=Bar,json=b,proto3" json:"b,omitempty"`
}
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1783,6 +1783,9 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		fieldName, fieldGetterName := ns[0], ns[1]
 		typename, wiretype := g.GoType(message, field)
 		jsonName := *field.Name
+		if json := field.GetJsonName(); json != "" {
+			jsonName = json
+		}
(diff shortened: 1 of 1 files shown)


dsnet:
Thanks for the PR, but unfortunately we cannot accept this pull request even if the change is entirely reasonable.

The generation of the `json` tags preceded the creation of the `jsonpb` package with the intention of always using the standard library `encoding/json` package for serializing protobufs as JSON. However, it became clear over time that the stdlib version of `json` was not sufficient, which eventually led to the creation of `jsonpb`. At some point, the `json_name` option was added to control the JSON name, but it only affects the name used in `jsonpb`.

We cannot have `json_name` option also affect the `json` tag since it would break compatibility for users still using the stdlib `json` package and rely on the current behavior. Fiixing this is probably not worth the breakages it may cause.

PR #344, opened 2017-04-25
Convert JSON 0/1 bools to protobuf bools

jeffreypicard:
Many json messages use 0/1 values for booleans, specifically older
versions of openrtb messages. Other than this difference the json
and protobuf objects align perfectly. A setting to allow the
Unmarshaler to detect and convert these fields would be very useful
for integrating with systems using json openrtb messages while
using the current protobuf version internally.


dsnet:
The mapping of protos to JSON is defined by the [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) specification. In it, it says that bools are mapped to the `true` and `false` values in JSON.

Unfortunately, we cannot accept this change.

Issue #50, opened 2015-07-16
Compile google/protobuf/*.proto files into golang structs and add to a new package


peter-edge:
I did this https://github.com/peter-edge/go-google-protobuf which is fine, but I would love to not have to maintain this, and it makes a lot of sense to have it somewhere in here. I would be happy to do the work and commit it over, but before doing any work on this, wanted to see how open the maintainers were to it. Thoughts?

Also addresses https://github.com/golang/protobuf/issues/49

grepory:
@peter-edge Hi. Sorry for this, but I just started doing:

```
sed -i 's/import google_protobuf "google\/protobuf"/import google_protobuf "github.com\/peter-edge\/go-google-protobuf"/' ${proto_dir}/checker.pb.go
```

Because works. Let me know if I can do anything to help maintain. I'm not sure of how it might be integrated into this project, exactly. But yeah... I have this problem, too.

peter-edge:
@grepory note that you'll want to update this to go.pedge.io/google-protobuf instead (started using custom import urls for my active golang projects)

I wrote a tool that takes care of everything for you called protoc-all: https://github.com/peter-edge/go-tools/tree/master/protoc-all

It fixes all the import paths etc too. Sample usage: https://github.com/pachyderm/pachyderm/blob/master/Makefile#L85

You shouldn't need to set PROTOC_INCLUDE_PATH in general. Ping me at my email if you need help.

grepory:
Why are you so awesome? Thanks. :heart:

peter-edge:
Theoretically I'm paid to be amazing, but hey, I think the joke is on Pachyderm :)

ecnahc515:
This would be awesome. It would make using protobuf with Go significantly easier

peter-edge:
Hey, ping on this, I'm seeing a bunch of people starting to import go.pedge.io/google-protobuf and obviously that's sub-optimal :) I also just wrote go.pedge.io/googleapis for the same purpose, which references go.pedge.io/googleapis, I'd love if Google just took over both of them heh.

chai2010:
build `type.proto` failed:

```
protoc --go_out=. any.proto source_context.proto type.proto
google/protobuf/any.proto:94:10: "google.protobuf.Any.type_url" is already defined in file "any.proto".
google/protobuf/any.proto:97:9: "google.protobuf.Any.value" is already defined in file "any.proto".
[...]

PR #350, opened 2017-05-16
timestamp: add means to convert Timestamp to & from time.Time

azr:
I did this right after proposing it in #50 to gain some time, it's fine if you drop it :)

Cheers !

### DIFF
--- a/ptypes/timestamp/timestamp.go
+++ b/ptypes/timestamp/timestamp.go
@@ -1,0 +1,20 @@
+package timestamp
+
+import "time"
+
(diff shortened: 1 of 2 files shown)


cybrcodr:
These already exists.  See...
https://godoc.org/github.com/golang/protobuf/ptypes#Timestamp
https://godoc.org/github.com/golang/protobuf/ptypes#TimestampProto

azr:
Woops ! Sorry, I expected them to be in their respective pkgs !

Issue #360, opened 2017-05-26
jsonpb marshaling Any containing dynamic message missing @type


cybrcodr:
Sample proto file
```
syntax = "proto2";
package pb;

message Foo {
  optional google.protobuf.Any iany = 1;
}
```
Sample code that stores a dynamic message into Foo.iany field.
```
package test

import (
        "strings"
        "testing"

        "test/pb"

        "github.com/golang/protobuf/jsonpb"
        "github.com/golang/protobuf/proto"
        anypb "github.com/golang/protobuf/ptypes/any"
)

type dynamic struct {
        Json []byte `protobuf:"bytes,1,opt,name=json" json:"json,omitempty"`
}

func (d *dynamic) Reset() {
        d.Json = nil
}

func (d *dynamic) String() string {
        return string(d.Json)
}

func (d *dynamic) ProtoMessage() {}

func (d *dynamic) MarshalJSONPB(_ *jsonpb.Marshaler) ([]byte, error) {
        return d.Json, nil
}

func (d *dynamic) UnmarshalJSONPB(_ *jsonpb.Unmarshaler, jstr []byte) error {
        d.Json = jstr
        return nil
}

func init() {
        proto.RegisterType((*dynamic)(nil), "pb.Dynamic")
}

func TestMarshalAnyDynamic(t *testing.T) {
        d := &dynamic{Json: []byte(`{"hello": "world"}`)}
        b, err := proto.Marshal(d)
        if err != nil {
                t.Fatal(err)
        }

        foo := &pb.Foo{
                Iany: &anypb.Any{
                        TypeUrl: "ok.com/pb.Dynamic",
                        Value:   b,
                },
        }

        m := &jsonpb.Marshaler{}
        s, err := m.MarshalToString(foo)
        if err != nil {
                t.Error(err)
        } else {
                t.Log(s)
        }
}
```

Outputs:
```
{"iany":{"hello": "world"}}
```

Per JSON mapping spec (https://developers.google.com/protocol-buffers/docs/proto3#json)

If the Any contains a value that has a special JSON mapping, it will be converted as follows: {"@type": xxx, "value": yyy}. Otherwise, the value will be converted into a JSON object, and the "@type" field will be inserted to indicate the actual data type. 

Output is missing "@type" key-value.  I'm going to treat dynamic message as the latter case, i.e. not having any special JSON mapping.  So, above should have output 
```
{"iany":{"@type":"ok.com/pb.Dynamic", "hello": "dynamic"}}
```

[...]

PR #361, opened 2017-05-26
fix marshaling of any types to JSON when contained object has custom marshaling

jhump:
This should address #360 

I included support for the custom message type to marshal itself into JSON types other than object. So it's a little more flexible than the solution @cybrcodr suggested in #360.

Also fixed a potential nil-dereference panic if an Any type was unmarshaled from bad JSON that had a null type:
```json
{ "@type": null, "key": "value" }
```

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -77,12 +77,7 @@ type Marshaler struct {
 // JSONPBMarshaler is implemented by protobuf messages that customize the
 // way they are marshaled to JSON. Messages that implement this should
 // also implement JSONPBUnmarshaler so that the custom format can be
-// parsed. If this method produces JSON other than an object (for example,
-// produces an encoded JSON string), the type must also implement another
-// method:
-//    XXX_CustomJSON() bool
-// If this method is implemented and returns true when invoked, then the
-// message is allowed to marshal itself to JSON types other than objects.
+// parsed.
 type JSONPBMarshaler interface {
 	MarshalJSONPB(*Marshaler) ([]byte, error)
 }
@@ -90,13 +85,7 @@ type JSONPBMarshaler interface {
 // JSONPBUnmarshaler is implemented by protobuf messages that customize
 // the way they are unmarshaled from JSON. Messages that implement this
 // should also implement JSONPBMarshaler so that the custom format can be
-// produced. If this method expects to unmarshal itself from JSON other
-// than an object (for example, unmarshals from an encoded string), the
-// type must also implement another method:
-//    XXX_CustomJSON() bool
-// If this method is implemented and returns true when invoked, then the
-// message is allowed to unmarshal itself from JSON types other than
-// objects.
+// produced.
 type JSONPBUnmarshaler interface {
 	UnmarshalJSONPB(*Unmarshaler, []byte) error
 }
@@ -127,14 +116,6 @@ type wkt interface {
 	XXX_WellKnownType() string
 }
 
-// customJSON can be implemented by messages that have custom non-object JSON
-// representations. If the message implements this method and the method
-// returns true, the JSON representation will not be an object (for example,
-// the message marshals itself as a string).
-type customJSON interface {
-	XXX_CustomJSON() bool
-}
-
 // marshalObject writes a struct to the Writer.
 func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeURL string) error {
 	if jsm, ok := v.(JSONPBMarshaler); ok {
@@ -144,22 +125,16 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
 		}
 		if typeURL != "" {
 			// we are marshaling this object to an Any type
-			var js interface{}
+			var js map[string]*json.RawMessage
 			if err = json.Unmarshal(b, &js); err != nil {
-				return fmt.Errorf("Type %v produced invalid JSON: %v", reflect.TypeOf(v), err)
+				return fmt.Errorf("type %v produced invalid JSON: %v", reflect.TypeOf(v), err)
 			}
-			m, ok := js.(map[string]interface{});
-			if ok {
-				m["@type"] = typeURL
-			} else {
-				if cj, ok := v.(customJSON); !ok || !cj.XXX_CustomJSON() {
-					return fmt.Errorf("Type %v illegally produced JSON that is not an object", reflect.TypeOf(v))
-				}
-				m = map[string]interface{}{}
-				m["@type"] = typeURL
-				m["value"] = js
+			turl, err := json.Marshal(typeURL)
+			if err != nil {
+				return fmt.Errorf("failed to marshal type URL %q to JSON: %v", typeURL, err)
 			}
-			if b, err = json.Marshal(m); err != nil {
+			js["@type"] = (*json.RawMessage)(&turl)
+			if b, err = json.Marshal(js); err != nil {
 				return err
 			}
 		}
@@ -688,16 +663,7 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			}
 
 			m := reflect.New(mt.Elem()).Interface().(proto.Message)
-			hasCustomJson := false
 			if _, ok := m.(wkt); ok {
-				hasCustomJson = true
-			} else if _, ok := m.(JSONPBUnmarshaler); ok {
-				// custom JSON unmarshallers may expect non-object JSON type
-				if cj, ok := m.(customJSON); ok && cj.XXX_CustomJSON() {
-					hasCustomJson = true
-				}
-			}
-			if hasCustomJson {
 				val, ok := jsonFields["value"]
 				if !ok {
 					return errors.New("Any JSON doesn't have 'value'")
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -77,7 +77,12 @@ type Marshaler struct {
 // JSONPBMarshaler is implemented by protobuf messages that customize the
 // way they are marshaled to JSON. Messages that implement this should
 // also implement JSONPBUnmarshaler so that the custom format can be
-// parsed.
+// parsed. If this method produces JSON other than an object (for example,
+// produces an encoded JSON string), the type must also implement another
+// method:
+//    XXX_CustomJSON() bool
+// If this method is implemented and returns true when invoked, then the
+// message is allowed to marshal itself to JSON types other than objects.
 type JSONPBMarshaler interface {
 	MarshalJSONPB(*Marshaler) ([]byte, error)
 }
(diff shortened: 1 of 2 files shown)


jsonpb/jsonpb.go
```diff
@@ -116,13 +127,43 @@ type wkt interface {
 	XXX_WellKnownType() string
 }
 
+// customJSON can be implemented by messages that have custom non-object JSON
+// representations. If the message implements this method and the method
+// returns true, the JSON representation will not be an object (for example,
+// the message marshals itself as a string).
+type customJSON interface {
+	XXX_CustomJSON() bool
```
jhump: I suspect there will be opinions on what this method should be named, and possibly even whether this should even be supported (custom message type that can marshal itself into custom JSON).

The main case I can think of is a dynamic message that happens to have a well-known type as a nested field. In that case, depending on implementation, it may be possible that the well-known type is actually represented via a dynamic message vs. the generated message type. Perhaps there are others. Possibly worth pointing out that the doc for `Any` just talks about messages with custom JSON and doesn't specifically restrict this to only well-known types.

jsonpb/jsonpb.go
```diff
@@ -116,13 +127,43 @@ type wkt interface {
...
 		if err != nil {
 			return err
 		}
+		if typeURL != "" {
+			// we are marshaling this object to an Any type
+			var js interface{}
+			if err = json.Unmarshal(b, &js); err != nil {
+				return fmt.Errorf("Type %v produced invalid JSON: %v", reflect.TypeOf(v), err)
+			}
+			m, ok := js.(map[string]interface{});
```
cybrcodr: Use map[string]*json.RawMessage instead?

jsonpb/jsonpb.go
```diff
@@ -116,13 +127,43 @@ type wkt interface {
...
+			if ok {
+				m["@type"] = typeURL
+			} else {
+				if cj, ok := v.(customJSON); !ok || !cj.XXX_CustomJSON() {
```
cybrcodr: As mentioned above, I'm not sure if this is necessary.  I'd suggest error'ing here for now.  But if you think this is something needed, then I think you can simply take the RawMessage and set it to "value", right?

jsonpb/jsonpb.go
jhump: I did see your question a few moments ago.

Sure, I can cut this stuff out for now. It's not terribly complicated to add back later if/when a more concrete need arises.

jsonpb/jsonpb.go
jhump: If I cut out the code that supports non-object JSON types, then, yes, I can unmarshal into that type (which will certainly be more efficient to parse and then to re-marshal below).

The current formulation is because I didn't know apriori whether the JSON produced will be an object or something else. So I have to supply an `interface{}` to `json.Unmarshal`, in which case the type of the value will be `map[string]interface{}` if it was an object/hash.

jsonpb/jsonpb.go
jhump: > As mentioned above, I'm not sure if this is necessary. I'd suggest error'ing here for now. 

Okay, I'll tweak this branch to remove the `customJson` stuff and push changes.

> But if you think this is something needed, then I think you can simply take the RawMessage and set it to "value", right?

I think that's what the code is doing below. But perhaps I'm not following this remark. (Though it will be moot after removing the `customJson` stuff.)

jsonpb/jsonpb.go
```diff
@@ -633,13 +668,13 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			}
 
 			val, ok := jsonFields["@type"]
-			if !ok {
+			if !ok || val == nil {
```
jhump: Yeah, I had actually made the same change locally in my branch, in order for my tests to pass. After I rebased to latest master and reconciled with your fix, I figured I'd leave this change in :)

jsonpb/jsonpb.go
cybrcodr: Agree.  Remark is moot when going with the assumption that a dynamic message should produce a JSON object.  And we should go with that.

Sorry for lack of explanation on my end.  My previous comment was if it did not produce a JSON object, I'd assume that it should still produce a valid JSON value and hence you can simply set the bytes output from MarshalJSONPB to "value".

jsonpb/jsonpb.go
```diff
@@ -633,13 +668,13 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
...
 				return errors.New("Any JSON doesn't have '@type'")
 			}
 
 			var turl string
 			if err := json.Unmarshal([]byte(*val), &turl); err != nil {
-				return fmt.Errorf("can't unmarshal Any's '@type': %q", val)
+				return fmt.Errorf("can't unmarshal Any's '@type': %q", *val)
 			}
 			target.Field(0).SetString(turl)
```
jhump: I think it's probably okay without such a check. What will happen is that this method will end up returning an error like `unknown message type ""`, which seems reasonable that it's saying the empty string is not valid as the type URL.

jsonpb/jsonpb.go
jhump: Ah, I see what you meant. So, were I to have kept this code, basically change this:

```go
m["value"] = js
```

to this:

```go
m["value"] = (*json.RawMessage)(&b)
```

Anyhow, removed all of this code; but, now that I understand the suggestion, yes, that would have been a better way to do it! :)

jhump:
@cybrcodr I removed the support for "custom JSON" (e.g. non-JSON-object output) and addressed your other comments. PTAL

jsonpb/jsonpb_test.go
```diff
@@ -654,15 +674,42 @@ func TestUnmarshalingBadInput(t *testing.T) {
...
 	}
 }
 
+func TestUnmarshalFromAnyWithJSONPBUnmarshaler(t *testing.T) {
+	rawJson := `{ "@type": "blah.com/` + dynamicMessageName + `", "foo": "bar", "baz": [0, 1, 2, 3] }`
+	var a anypb.Any
+	if err := Unmarshal(strings.NewReader(rawJson), &a); err != nil {
+		t.Errorf("an unexpected error occurred when parsing into JSONPBUnmarshaler: %v", err)
+	}
+	var msg ptypes.DynamicAny
```
cybrcodr: How about constructing an expected Any and compared directly using proto.Equal?

d := &dynamicMessage{rawJson: `{"baz":[0,1,2,3],"foo":"bar"}`}
b, err := proto.Marshal(d)
if err != nil {
    ...
}
want := &anypb.Any{
    TypeURL: "blah.com/` + dynamicMessageName + `",
    Value:  b,
}

if proto.Equal(got, want) ...

jsonpb/jsonpb_test.go
```diff
@@ -684,7 +731,35 @@ func (m *dynamicMessage) MarshalJSONPB(jm *Marshaler) ([]byte, error) {
 	return []byte(m.rawJson), nil
 }
 
-func (m *dynamicMessage) UnmarshalJSONPB(jum *Unmarshaler, json []byte) error {
-	m.rawJson = string(json)
+func (m *dynamicMessage) UnmarshalJSONPB(jum *Unmarshaler, js []byte) error {
+	m.rawJson = string(js)
```
cybrcodr: Optional.  Perhaps change rawJson's type to []byte instead of string, then there'll be fewer type casting around.

jsonpb/jsonpb_test.go
```diff
@@ -684,7 +731,35 @@ func (m *dynamicMessage) MarshalJSONPB(jm *Marshaler) ([]byte, error) {
...
 	return nil
 }
+
+func (m *dynamicMessage) Marshal() ([]byte, error) {
```
cybrcodr: Is this necessary?  If field is exported, I think it'll marshal and unmarshal correctly.  I'm ok with this, just curious though.

jsonpb/jsonpb.go
```diff
@@ -660,7 +670,7 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 				}
 
 				if err := u.unmarshalValue(reflect.ValueOf(m).Elem(), *val, nil); err != nil {
-					return fmt.Errorf("can't unmarshal Any's WKT: %v", err)
+					return fmt.Errorf("can't unmarshal Any nested proto %v: %v", reflect.TypeOf(m), err)
```
jhump: done

jsonpb/jsonpb_test.go
```diff
@@ -442,7 +444,7 @@ func TestMarshaling(t *testing.T) {
 	}
 }
 
-func TestMarshalingWithJSONPBMarshaler(t *testing.T) {
+func TestMarshalWithJSONPBMarshaler(t *testing.T) {
```
jhump: done

jsonpb/jsonpb_test.go
```diff
@@ -454,6 +456,24 @@ func TestMarshalingWithJSONPBMarshaler(t *testing.T) {
 	}
 }
 
+func TestMarshalToAnyWithJSONPBMarshaler(t *testing.T) {
```
jhump: done

jsonpb/jsonpb_test.go
jhump: done

jsonpb/jsonpb_test.go
```diff
@@ -654,15 +674,42 @@ func TestUnmarshalingBadInput(t *testing.T) {
...
+	// already, so the keys are sorted, whitespace is compacted, and "@type" key has been removed
+	expected := `{"baz":[0,1,2,3],"foo":"bar"}`
+	if dm.rawJson != expected {
+		t.Errorf("message contents not set correctly after unmarshalling JSON: got %s, wanted %s", dm.rawJson, expected)
+	}
+}
+
+const (
+	dynamicMessageName = "google.protobuf.jsonpb.testing.dynamicMessage"
+	dynamicMessageCustomJsonName = "google.protobuf.jsonpb.testing.dynamicMessageCustomJson"
```
jhump: Oops. Didn't mean to leave that in there. Now gone.

jsonpb/jsonpb_test.go
jhump: I looked at doing this, but it would actually result in more casts, not fewer (6 casts instead of just 2).

jsonpb/jsonpb_test.go
jhump: You're right. If I add a properly formatted struct tag to the field, this is no longer necessary. Done.

jhump:
@cybrcodr, hopefully this is ready to merge now.

jsonpb/jsonpb_test.go
cybrcodr: Ah yes, it's the protobuf tag instead.

PR #369, opened 2017-06-09
Implement Compare for Timestamp

kunalgosar:
Implementation of a logical comparison for two Timestamps.

### DIFF
--- a/ptypes/timestamp.go
+++ b/ptypes/timestamp.go
@@ -123,3 +123,19 @@ func TimestampString(ts *tspb.Timestamp) string {
 	}
 	return t.Format(time.RFC3339Nano)
 }
+
+// Compare compares two timestamps logically.
+// Returns: -1 if a < b, 0 if a == b, and 1 if a > b
+func Compare(a, b *tspb.Timestamp) int {
+	if a.Seconds == b.Seconds {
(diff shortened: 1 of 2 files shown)


cybrcodr:
Given existing APIs, one can convert to time.Time and compare using that type.  By converting to time.Time, it also ensures that these will be valid time values else conversion will return error.

I can understand your request for direct comparison of the Timestamp proto w/o validation, but I'm not sure if that'll mislead users when there are invalid time values.  And if it'll be doing the time conversion then I'd rather keep the API surface smaller and be handled by callers instead.

Just my opinion.

kunalgosar:
That's very fair. It would not make sense to compare Timestamp protos without validation. However, when sorting a protobuf generated object by Timestamp field or extending an interface for data structure that requires comparison, the repeated conversions would add a significant overhead, and we may assume the Timestamps have been pre-validated. 

I understand if this is not a large enough use case to warrant complicating the API. Let me know what you think.

cybrcodr:
If using the sort package to sort, one will need to implement sort.Interface or provide a less function for sort.Slice call.  Agree that a compare function here can be used and wrapped by those sort APIs, but that also means that callers will still need to implement sort.Interface or the less function.

PR #375, opened 2017-06-15
Return registered nil Message

vaporz:
### What this PR do
1, Make a map:  
`registeredMessages = make(map[string]Message)`
2, Store registered Message:
`registeredMessages[name] = x`
3, Return this Message by name:
`func MessageRegistered(name string) Message { return registeredMessages[name] }`
### Why
To lookup field options of request struct in a "google.golang.org/grpc.UnaryInterceptor", I have to:
1, With "req interface{}" as input, find the TypeName using reflection.
2, With TypeName, call MessageRegistered() to get a proto.Message.
3, With a proto.Message, call 
`_, md = descriptor.ForMessage( protoMessage.(descriptor.Message)  )`
`opt, err := proto.GetExtension(md.GetField()[0].Options, E_SomeOption)`
to get option value.

This PR adds the func to be used in step 2.

### DIFF
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -826,8 +826,9 @@ func EnumValueMap(enumType string) map[string]int32 {
 // A registry of all linked message types.
 // The string is a fully-qualified proto name ("pkg.Message").
 var (
-	protoTypes    = make(map[string]reflect.Type)
-	revProtoTypes = make(map[reflect.Type]string)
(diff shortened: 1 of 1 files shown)


vaporz:
closed,  have found better solution:
_, md := descriptor.ForMessage(reflect.ValueOf(req).Interface().(descriptor.Message))

PR #381, opened 2017-06-22
Add Travis CI status

cybrcodr:
This PR should trigger a Travis build+test run as I've just enabled it.


cybrcodr:
Sigh, seems like it is not using the .travis.yml file from the PR.  So, I think the only way would be to submit this in and try again.  Anyways, I'll continue this tomorrow.

Issue #385, opened 2017-06-27
no way to access data for unrecognized extensions


jhump:
Access to extensions is provided by top-level functions in the `proto` package (in `extensions.go`). But they omit any way to query for the values of unrecognized extensions (as raw unparsed bytes).

My immediate use case is to be able to re-parse custom method options on a descriptor proto using a dynamic extension registry (not limited to just linked-in extensions). This allows a "server browser" tool to show more useful information about a server's exposed services instead of just "unknown fields/extensions". Using the [server reflection API](https://godoc.org/google.golang.org/grpc/reflection), the client can easily ask the server for all known extensions and then re-parse the descriptors (for example, using a [dynamic message](https://godoc.org/github.com/jhump/protoreflect/dynamic)).

For symmetry with how C++ and Java libraries support extension and unknown fields, it would be preferred for unrecognized extensions to end up in `XXX_unrecognized` instead of hidden inside of a map of unparsed extension values that is inaccessible. (It used to be accessible via the `ExtensionMap()` method, but that is no longer generated by protoc-gen-go.)

I also brought this up on the protobuf mailing list, where I think @xfxyjwf expressed a similar sentiment:
https://groups.google.com/d/topic/protobuf/HTNFwKLxtRg/discussion

I believe this could be addressed by merely changing the unmarshalling code to append the data to `XXX_unrecognized` instead of storing in the map when the looked up `*ExtensionDesc` is nil.

jhump:
@dsnet any updates on how/when this issue might be resolved?

jhump:
@dsnet, what do you think about #420 as a solution for this? Unlike #386, it has no compatibility issues. Instead, it increases API surface area by one method.

dsnet:
I'll take a look. With Go1.9 out the door, I'm swinging my attention back over to protos.

dsnet:
Closing as fixed with #583. Track #479 for the merging of `dev` into `master`.

Issue #364, opened 2017-05-30
APIv2: proto: make the Message interface behaviorally complete


dsnet:
Filing on behalf of @alandonovan.
> The proto.Message interface is unsatisfactory.  A behavioral interface is an abstraction over the underlying data types that exposes just the operations necessary for their correct use.  By contrast, proto.Message exposes essentially no useful functionality and serves only as a marker.  But a marker of what?  If one could assume all its implementations were protoc-generated struct types with field tags, then at least it would be possible to write reflection-based algorithms that do useful things.  However, there are many concrete types satisfying Message that are not of this form.  
> 
> It's not only the set of implementations that is unbounded; the set of useful operations is also large and growing. The two most important, Marshal and Unmarshal, are handled quite cleanly since there are separate behavioral interfaces for Marshaler and Unmarshaler that allow each concrete type to implement these operations.  But there are many functions in the proto API, for which no interface exists: proto.Merge, proto.Clone, the extensions API, and so on.
> 
> The cross-product of concrete implementations and operations is growing, but the fraction of these combinations that actually work is diminishing.
> 
> I think we should assess what it would take to change the proto.Message interface, and all its implementations, so that it is a true behavioral interface.  This would require at a minimum that the interface include a new method that provides complete control over the abstract state of a message: accessing and updating its fields, inspecting any extensions or unrecognized fields, and so on, without revealing the concrete representation.  It should be possible to implement all the major functions in the proto API, as well as most users' ad hoc functions, in terms of this interface so that they work with any concrete implementation.  If an optimized version of a crucial operation is available, the generic implementation should dispatch to it, as it does today for Marshal and Unmarshal.

We can't add methods to `proto.Message` without breaking backwards compatibility. One approach we can take is to define `proto.MessageV2` that is a much more semantically complete interface that provides a form of "protobuf reflection". In `Marshal`, `Unmarshal`, `Merge`, `Clone`, and so on, we can type assert if it implements `proto.MessageV2` and use that interface to implement generic versions of those functions. If `proto.Message` doesn't satisfy `proto.MessageV2`, then `Merge` can just fail (it already does on most third-party implementations of `proto.Message`).

jhump:
Interesting proposal. I faced a similar challenge when designing the public API for a [dynamic message implementation](https://godoc.org/github.com/jhump/protoreflect/dynamic#Message).

[...]

PR #386, opened 2017-06-27
unmarshal unknown extensions into XXX_unrecognized instead of into extenson map

jhump:
This change would address #385

### DIFF
--- a/proto/decode.go
+++ b/proto/decode.go
@@ -462,6 +462,8 @@ func (p *Buffer) Unmarshal(pb Message) error {
 func (o *Buffer) unmarshalType(st reflect.Type, prop *StructProperties, is_group bool, base structPointer) error {
 	var state errorState
 	required, reqFields := prop.reqCount, uint64(0)
+	var regExt map[int32]*ExtensionDesc
+	regExtInit := false
 
 	var err error
 	for err == nil && o.index < len(o.buf) {
@@ -492,11 +494,22 @@ func (o *Buffer) unmarshalType(st reflect.Type, prop *StructProperties, is_group
 			// Maybe it's an extension?
 			if prop.extendable {
 				if e, _ := extendable(structPointer_Interface(base, st)); isExtensionField(e, int32(tag)) {
-					if err = o.skip(st, tag, wire); err == nil {
-						extmap := e.extensionsWrite()
-						ext := extmap[int32(tag)] // may be missing
-						ext.enc = append(ext.enc, o.buf[oi:o.index]...)
-						extmap[int32(tag)] = ext
+					if !regExtInit {
+						msgType := reflect.Zero(reflect.PtrTo(st)).Interface().(Message)
+						regExt = RegisteredExtensions(msgType)
+					}
+					extdesc := regExt[int32(tag)]
+					if extdesc == nil {
+						// unknown extension
+						err = o.skipAndSave(st, tag, wire, base, prop.unrecField)
+					} else {
+						if err = o.skip(st, tag, wire); err == nil {
+							extmap := e.extensionsWrite()
+							ext := extmap[int32(tag)] // may be missing
+							ext.enc = append(ext.enc, o.buf[oi:o.index]...)
+							ext.desc = extdesc
+							extmap[int32(tag)] = ext
+						}
 					}
 					continue
 				}
--- a/proto/extensions.go
+++ b/proto/extensions.go
@@ -491,8 +491,10 @@ func GetExtensions(pb Message, es []*ExtensionDesc) (extensions []interface{}, e
 }
 
 // ExtensionDescs returns a new slice containing pb's extension descriptors, in undefined order.
-// For non-registered extensions, ExtensionDescs returns an incomplete descriptor containing
-// just the Field field, which defines the extension's field number.
+// If the message was de-serialized from a stream that referenced unknown extensions (e.g. fields
+// with a tag number in an extension range, but not registered), they will not be returned by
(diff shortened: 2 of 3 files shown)


jhump:
@matloob, you are the last person to make material changes in this stuff (I realize that was many months ago). Are you still working on protos? Could you take a look or perhaps tag whoever is the best person to look?

proto/decode.go
```diff
@@ -492,11 +494,22 @@ func (o *Buffer) unmarshalType(st reflect.Type, prop *StructProperties, is_group
...
 			if prop.extendable {
 				if e, _ := extendable(structPointer_Interface(base, st)); isExtensionField(e, int32(tag)) {
-					if err = o.skip(st, tag, wire); err == nil {
-						extmap := e.extensionsWrite()
-						ext := extmap[int32(tag)] // may be missing
-						ext.enc = append(ext.enc, o.buf[oi:o.index]...)
-						extmap[int32(tag)] = ext
+					if !regExtInit {
+						msgType := reflect.Zero(reflect.PtrTo(st)).Interface().(Message)
+						regExt = RegisteredExtensions(msgType)
```
jhump: Maybe it's more clear to just do `regExt := RegisteredExtensions(msgType)` here. But I'm weary of doing any work more than necessary in something like proto unmarshaling since that is likely to be a hot path in many servers. (Though, admittedly, extensions are likely used infrequently enough that this is not necessary. Suggestions?)

jhump:
Ping!

Do issues and pull requests go into a queue that someone will eventually look at? Or do they largely fall on deaf ears unless I tag the right person, and that person has time to take a look? (Based on previous attempts to contribute, it kind of feels like the latter.)

I know there might be low signal-to-noise ratio in the stuff that Github community sends Google's way. But if someone could at least triage and drop a brief comment -- like an ack or nack, not necessarily full review -- that would be super helpful. I'm hoping that this could bubble up onto someone's radar and be interpreted as signal, not noise.

dsnet:
My main concern with this change is backwards compatibility. There are users that clear out or use the XXX_unrecognized fields for various reasons (security, performance, etc). Some of them may have wanted the unknown extensions in there, while this would be breaking others. The API of message extensions in the Go proto implementation are not the most consistent, but plethora of existing codebases that depend on this makes this really difficult to change.

If I may ask, what is the major benefit to you that they are unmarshaled to XXX_unrecognized as opposed to some other mechnism? While XXX_unrecognized is exported, the XXX_ prefix was supposed to indicate that these are private fields.

Unless there is a good reason, I would rather keep the current behavior. Secondly, with the release of proto3, (hopefully) the use of proto2 and message extensions will diminish drastically.

dsnet:
> But if someone could at least triage and drop a brief comment -- like an ack or nack, not necessarily full review -- that would be super helpful.

I apologize for the slow rate of response. I will be focusing on protobufs more after the Go1.9 release.

jhump:
@dsnet: Thank you for the reply!

> There are users that clear out or use the XXX_unrecognized fields for various reasons (security, performance, etc). Some of them may have wanted the unknown extensions in there, while this would be breaking others

I don't really follow this logic. I would expect users that care about unknown fields would be *more* interested in ones defined in the message itself vs. those defined as extensions.

I do understand it's a change in backwards compatibility. Another thought I had was to provide an accessor, like `GetRawExtension()` (analogous to existing `GetExtension()` but returns raw bytes and could be used for unknown extensions).

> If I may ask, what is the major benefit to you that they are unmarshaled to XXX_unrecognized as opposed to some other mechnism?

If you see the bug report (#385), the main thing I am trying to accomplish is that they exposed somehow. This is kind of related to #364, if you consider handling of extensions and unrecognized fields as "behavior" of a protobuf message (like the APIs in Java and C++).

In particular, I have code to convert from a generated message to a [dynamic message](https://godoc.org/github.com/jhump/protoreflect/dynamic#Message). I would like to be able to re-parse unrecognized fields during this step since the dynamic message's descriptor and registered extensions may be more exhaustive than what was linked into the binary actually doing the conversion. I could serialize the message to bytes and then de-serialize into the dynamic message, but that feels too inefficient. So I instead was hoping to only de-serialize the bytes for unrecognized fields, but trying to simplify [awfulness in the current implementation](https://github.com/jhump/protoreflect/blob/f5aa422cb4a0e05e49dc41d88acea75d18cb9cf8/dynamic/dynamic_message.go#L1569) (to work around the fact that there are no ways to access unrecognized extensions).

dsnet:
> I don't really follow this logic. I would expect users that care about unknown fields would be more interested in ones defined in the message itself vs. those defined as extensions.
People use protos is all sorts of crazy ways that I'm no longer surprised by weird things people do with them. Subtle changes in behavior like this makes me very nervous.

That being said, I apologize for not seeing that you referenced #385, which seems a like legitimate need. Whether that does mean moving those unrecognized bytes to XXX_unrecognized or introducing new API surface will need to be thought through carefully. New API is certainly the least likely to break people, but it would be nice to not to have new API. Let me investigate more deeply the implications of unmarshalling into XXX_unrecognized.

dsnet:
I tested this change and too many targets broke since they were depending on extensions *not* being in XXX_unrecognized. I believe the proper fix for this is a behaviorally complete API for introspecting into all aspects of a proto message (#364). Access to unrecognized extensions would be part of that API and we can finally kill off the XXX_unrecognized field and friends.

I know it's been some time, but I've made good progress on an API for #364. I will probably publish a document in the near future.

jhump:
@dsnet, thanks for the update. I eagerly await your document, describing a fully-featured message API!

PR #387, opened 2017-06-28
generator: use consistent ModTime across go1.7 and go1.8

jdef:
without this change, protos generated by golang 1.7 and golang 1.8 differ because gzip on golang 1.7 uses "nonsense" values for ModTime when it's a zeroval (see golang 1.8 changelog w/ respect to gzip changes)

xref https://github.com/gogo/protobuf/pull/303

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -49,6 +49,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"time"
 	"unicode"
 	"unicode/utf8"
 
@@ -2607,6 +2608,7 @@ func (g *Generator) generateFileDescriptor(file *FileDescriptor) {
 
 	var buf bytes.Buffer
 	w, _ := gzip.NewWriterLevel(&buf, gzip.BestCompression)
+	w.Header.ModTime = time.Unix(0, 1) // consistent, non-zero val for time
 	w.Write(b)
 	w.Close()
 	b = buf.Bytes()


jdef:
@dsymonds PTAL (since it looks like you're the original author of the gzip related code)

dsnet:
I'd rather not add this. This reduces the chance that the generated files are different across versions of Go, but this does not fix the fundamental issue, which is that the output of gzip is not considered stable.

In the PR, we hack the ModTime to be consistent, but what happens when we change the implementation of the `compress/flate` algorithm again like we did in Go1.7 and Go1.8?

I understand the desire for a stable output for generated files, but this is not the right solution. The right approach might be to vendor the gzip and flate packages, but I need to think about whether we really want to vendor all that into this repo to fix this.

jdef:
Thanks for reviewing @dsnet, I'm trying to understand your concern. To paraphrase, it sounds like you're saying, "if we can't guarantee end-to-end, complete stability of generated code then it's not worth incrementally improving the stability of generated code to due known/possible upstream bugs/changes."

If that's a fair interpretation, then .. I'm not sure I agree with that assessment. At this point there's no guarantee that the std `gzip` implementation will ever be completely stable, but why is that an argument for not improving the relative stability of generated code, especially in the presence of known (and easily patched) issues? I'm under the impression that the stability of generated code is a high priority for this project.

dsymonds:
Honestly, I'm surprised that compress/gzip behaves weirdly with a zero ModTime. I guess it's good that Go 1.8 fixed that.

I think a small change like this is okay to work around some Go 1.7 weirdness. It's non-invasive. @dsnet is right that it isn't a complete solution, but I suspect a complete solution may take a while.

dsnet:
The reason I'm concerned about partial fixes is that it masks the existence of a more real problem: that the output of any package in the standard library is not stable. Even if we vendored a stable version of `compress/gzip` and `compress/flate`, any package that is used to produce output for the generated file is suspect for possible changing the output. For example, the exact output from strconv can change, or the even the String() method for any of the types used in the generator. There's an endless list of things that can cause the output to differ between Go versions. In some cases, the "fix" is easy to make, in others its nearly impossible.

Rather, I posit that users of the protoc-gen-go should be consistent about the version of Go that they use to generate the pb.go files that they check in. Perhaps, instead we should encode the version of Go toolchain used to generate the output so that it's obvious what version was used.

dsymonds:
I interpreted @jdef's report to be saying that Go 1.7 by itself was producing inconsistent gzipped output. I agree with you that it's not worth wrestling trying to make Go 1.7 and Go 1.8 behave identically.

dsnet:
I fixed the zero-timestamp bug in Go1.8, but it was there since Go1.0. In Go1.8 we also changed the `compress/flate` algorithm slightly, which is also why I find fixing the timestamp only strange, but not addressing that either.

dsnet:
I'm going to close this as we are coming up on Go1.10 soon. The generator should probably print versioning information in the "DO NOT EDIT" tag for reproducible generation. However, versioning of the generator is not a high priority problem at the moment.

Issue #373, opened 2017-06-13
grpc: plugin flag for generate server-only or client-only


adam-26:
Can some flags be added to the `grpc_go` plugin to allow only the required code to be generated? As done in the [C# plugin](https://github.com/grpc/grpc/issues/6077).

dsnet:
\cc @dfawley. Inside Google, I believe is a movement to provide finer granularity control over what stubs are general.

dfawley:
I have not heard of any such movement yet.  What is the significance of this request?  IIUC the only difference is that slightly fewer methods would be in the generated file, which would have a negligible effect on binary size.

dfawley:
Ping for @adam-26 since I forgot to at-mention earlier.

adam-26:
@dfawley, the issue was raised to allow for client code to be generated and distributed as part of an SDK without needing to include the server code. Nice to have, but not a blocking issue.

paranoiacblack:
I'm going to go ahead and say we're unlikely to add this to the upcoming release. As @dfawley notes changes here would likely have a negligible effect on binary size or performance. It seems that as a work around, the distributed SDK that @adam-26 mentioned can manually (or automatically) remove related server code from their generated proto package for the purpose of distributing a client-only implementation. It's not clear if this is overall useful for the grpc proto plugin.

Please re-open if you think there's more utility to this than meets the eye (or at least my eyes).

mattes:
From the [C# plugin](https://github.com/grpc/grpc/issues/6077):
> In a client library for other users, there's not much point in making the server API available, and it adds to pollution in the namespace.

Not only pollution, but also potentially confusing to user. 

> The distributed SDK can manually (or automatically) remove related server code.

That's what we are doing now. Yet, it would be nice if the server related code could be omitted with a flag in the first place.

PR #392, opened 2017-07-17
grpc: add an option to generate only client-side or server-side code

heltonmarx:
With this option `mode=client` or `mode=server` is it possible generates the client-side or server-side of a grpc protofile.
It's very useful when you only need the client-side signatures and generate the server-side with another generator (.e.g: a generic server-side generator to JSONRPC and gRPC).
Please let me know with this option make senses at all.


heltonmarx:
I signed it!

paranoiacblack:
I don't see really anything wrong with the implementation here, but closed the upstream issue in #373 because it's not clear what the utility of adding this is versus just pruning server or client code, respectively.

marcofranssen:
What is the reason this PR is not merged? This is a really helpfull feature where we don't need to manually remove the generated server code. We are generating clients in various different services. Now I still have a lot of manual work.

PR #404, opened 2017-08-07
jsonpb: don't allocate memory for null-value reference types

tamird:
This was missed in 748d386b5c1ea99658fd69fe9f03991ce86a90c1.

@cybrcodr another small issue following on from discussion in https://github.com/golang/protobuf/pull/338.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -919,11 +919,13 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		if err := json.Unmarshal(inputValue, &slc); err != nil {
 			return err
 		}
-		len := len(slc)
-		target.Set(reflect.MakeSlice(targetType, len, len))
-		for i := 0; i < len; i++ {
-			if err := u.unmarshalValue(target.Index(i), slc[i], prop); err != nil {
-				return err
+		if slc != nil {
+			l := len(slc)
+			target.Set(reflect.MakeSlice(targetType, l, l))
+			for i := 0; i < l; i++ {
+				if err := u.unmarshalValue(target.Index(i), slc[i], prop); err != nil {
+					return err
+				}
 			}
 		}
 		return nil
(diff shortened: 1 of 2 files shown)


jsonpb/jsonpb.go
```diff
@@ -637,15 +637,18 @@ func UnmarshalString(str string, pb proto.Message) error {
 func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMessage, prop *proto.Properties) error {
 	targetType := target.Type()
 
-	// Allocate memory for pointer fields.
-	if targetType.Kind() == reflect.Ptr {
+	switch targetType.Kind() {
+	case reflect.Map, reflect.Ptr, reflect.Slice:
```
cybrcodr: Let's not combine the logic for map and slice here.  It's better to just leave this for pointer type alone.

Suggest adding the following logic instead ...

```
if string(inputValue) == "null" {
    return nil
}
```

after original file lines 917 for handling slice
```
if targetType.Kind() == reflect.Slice...
```
and after line 933
```
if targetType.Kind() == reflect.Map {
```

tamird: Done, and added tests. Thanks for making me do that; this implementation turned out not to be correct.

jsonpb/jsonpb.go
```diff
@@ -919,11 +919,12 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		if err := json.Unmarshal(inputValue, &slc); err != nil {
 			return err
 		}
-		len := len(slc)
-		target.Set(reflect.MakeSlice(targetType, len, len))
-		for i := 0; i < len; i++ {
-			if err := u.unmarshalValue(target.Index(i), slc[i], prop); err != nil {
-				return err
+		if l := len(slc); l > 0 {
```
cybrcodr: I think if user explicitly specifies an empty JSON list/map then the deserialized Go struct field should also be an empty slice/map.  If JSON field is not specified or if explicitly set to "null", then leave Go field as nil.

jsonpb/jsonpb_test.go
```diff
@@ -645,6 +645,29 @@ func TestUnmarshaling(t *testing.T) {
...
 
+func TestUnmarshalNullReference(t *testing.T) {
+	var b bytes.Buffer
+
+	for _, ty := range []reflect.Type{
+		reflect.TypeOf(pb.Repeats{}),
+		reflect.TypeOf(pb.Maps{}),
+	} {
+		in := reflect.New(ty).Interface().(proto.Message)
+		if err := (&Marshaler{EmitDefaults: true}).Marshal(&b, in); err != nil {
```
cybrcodr: I prefer to test unmarshaler by itself and not rely on the Marshaler at all.  Basically, provide JSON strings to unmarshal and compare with expected result.

I also just realized that proto.Equal will not properly compare nil vs empty on these, so I reflect.DeepEqual is ok.

jsonpb/jsonpb.go
tamird: Done.

jsonpb/jsonpb_test.go
tamird: Done.

jsonpb/jsonpb.go
```diff
@@ -935,33 +937,35 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
...
-			// string, so we handle that specially. Other types were quoted post-serialization.
-			var k reflect.Value
-			if targetType.Key().Kind() == reflect.String {
-				k = reflect.ValueOf(ks)
-			} else {
-				k = reflect.New(targetType.Key()).Elem()
-				if err := u.unmarshalValue(k, json.RawMessage(ks), keyprop); err != nil {
-					return err
-				}
+		if mp != nil {
```
cybrcodr: Optional.  Do early return for mp == nil, that way there's no need to indent block below.
```
if mp == nil {
    return nil
}
target.Set(reflect.MakeMap(targetType))
...
```

Same above if you're going to make this change.

tamird: I'd rather leave it, avoiding the repetition of `return nil`. 6 of one, half-dozen of the other.

Issue #268, opened 2016-12-17
proto: remove or improve registration


awalterschulze:
We started a discussion here
https://github.com/golang/protobuf/commit/cf10ca0ea91371dd30946defd93a0ed43aa1efe7#commitcomment-20160444

awalterschulze:
I think Registering is breaking people's setups and I am wondering why this is needed in the first place

awalterschulze:
@bcmills says
> proto.MessageType and proto.MessageName require it. So does extension support, and ptypes.DynamicAny, and there may be other cases I'm not thinking of at the moment.

awalterschulze:
Also to be concrete here one place it is breaking people's setups
https://github.com/golang/protobuf/issues/178#issuecomment-215356264

awalterschulze:
Register fixes #179 only because ExtensionsMap was removed, which was another controversial move.

awalterschulze:
proto.MessageType and proto.MessageName are nice features, but not really necessary.
Extension support: ok yes maybe, but sacrificing some safety can also be used to get rid of Register.
How is register needed for DynamicAny?

bcmills:
> Register fixes #179 only because ExtensionsMap was removed, 

Register has ~nothing to do with #179.  #179 requires having a copy of the generated Descriptor proto package; the only impact that registration has on that is to require the generated Descriptor package to be the same one that other uses of Descriptor depend on.  (But that's true in general anyway: within a given binary, there should be one and only one copy of each package needed.  To do otherwise leads to bloated binaries.)

bcmills:
> proto.MessageType and proto.MessageName are nice features, but not really necessary.

This is a [Chesterton's Fence](https://www.chesterton.org/taking-a-fence-down/).  If you want to remove these methods, you should start by understanding what purpose they serve (or served).  "Not really necessary" merits further explanation.

> Extension support: ok yes maybe, but sacrificing some safety can also be used to get rid of Register.

Extensions are already error-prone; sacrificing safety is unlikely to help most users.  But can you be more specific about the tradeoff you're proposing?  What kinds of errors would it fail to detect?

> How is register needed for DynamicAny?

DynamicAny requires the ability to map from a fully-qualified message name to a Go type.  That is, it requires `proto.MessageType`.

awalterschulze:
How does having one copy of the generated Descriptor work with go vendoring?

[...]

PR #410, opened 2017-08-13
support custom URL resolver when marshaling/unmarshaling Any messages to/from JSON

jhump:
This is meant for use with dynamic message registries, where message descriptors could be downloaded from another source and still properly serialized/de-serializing from JSON, even when present in an `Any` message.

I've used an interface to represent the resolver instead of a function type because I have need for it to be backed by a non-function type (and just closing over variables isn't sufficient). The reason it needs to be an interface is for implementations of `JSONPBMarshaler` and `JSONPBUnmarshaler` that might want to decorate the resolver -- e.g. wrap it in a way that can supply additional message types. In such a case, we don't want to wrap the resolver more than once (which could happen since JSON serialization/de-serialization is recursive, and nested fields could also be instances of `JSONPBMarshaler`/`JSONPBUnmarshaler`). The only way to detect if it has already been wrapped is to do a type assertion, but function types are not comparable in Go.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -75,20 +75,19 @@ type Marshaler struct {
 	OrigName bool
 
 	// A custom URL resolver to use when marshaling Any messages to JSON.
-	// If unset, DefaultUrlResolver is used.
-	UrlResolver UrlResolver
+	// If unset, the default resolution strategy is to extract the
+	// fully-qualified type name from the type URL and pass that to
+	// proto.MessageType(string).
+	AnyResolver AnyResolver
 }
 
-// UrlResolver resolves a type URL into an instance of the associated message.
-// This is used for serializing and de-serializing Any messages to and from
-// JSON.
-type UrlResolver interface {
+// AnyResolver takes a type URL, present in an Any message, and resolves it into
+// an instance of the associated message.
+type AnyResolver interface {
 	Resolve(typeUrl string) (proto.Message, error)
 }
 
-type defaultResolver struct{}
-
-func (defaultResolver) Resolve(typeUrl string) (proto.Message, error) {
+func defaultResolveAny(typeUrl string) (proto.Message, error) {
 	// Only the part of typeUrl after the last slash is relevant.
 	mname := typeUrl
 	if slash := strings.LastIndex(mname, "/"); slash >= 0 {
@@ -100,12 +99,6 @@ func (defaultResolver) Resolve(typeUrl string) (proto.Message, error) {
 	}
 	return reflect.New(mt.Elem()).Interface().(proto.Message), nil
 }
-
-// DefaultUrlResolver can resolve any generated protobuf message that is
-// registered via proto.RegisterType. The fully-qualified message name is first
-// extracted from the type URL (e.g. the last path element). The resolver then
-// invokes proto.MessageType(string) to create a message instance for that name.
-var DefaultUrlResolver UrlResolver = defaultResolver{}
 
 // JSONPBMarshaler is implemented by protobuf messages that customize the
 // way they are marshaled to JSON. Messages that implement this should
@@ -376,11 +369,13 @@ func (m *Marshaler) marshalAny(out *errWriter, any proto.Message, indent string)
 	turl := v.Field(0).String()
 	val := v.Field(1).Bytes()
 
-	resolver := m.UrlResolver
-	if resolver == nil {
-		resolver = DefaultUrlResolver
+	var msg proto.Message
+	var err error
+	if m.AnyResolver != nil {
+		msg, err = m.AnyResolver.Resolve(turl)
+	} else {
+		msg, err = defaultResolveAny(turl)
 	}
-	msg, err := resolver.Resolve(turl)
 	if err != nil {
 		return err
 	}
@@ -623,8 +618,10 @@ type Unmarshaler struct {
 	AllowUnknownFields bool
 
 	// A custom URL resolver to use when unmarshaling Any messages from JSON.
-	// If unset, DefaultUrlResolver is used.
-	UrlResolver UrlResolver
+	// If unset, the default resolution strategy is to extract the
+	// fully-qualified type name from the type URL and pass that to
+	// proto.MessageType(string).
+	AnyResolver AnyResolver
 }
 
 // UnmarshalNext unmarshals the next protocol buffer from a JSON object stream.
@@ -714,11 +711,13 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			}
 			target.Field(0).SetString(turl)
 
-			resolver := u.UrlResolver
-			if resolver == nil {
-				resolver = DefaultUrlResolver
+			var m proto.Message
+			var err error
+			if u.AnyResolver != nil {
+				m, err = u.AnyResolver.Resolve(turl)
+			} else {
+				m, err = defaultResolveAny(turl)
 			}
-			m, err := resolver.Resolve(turl)
 			if err != nil {
 				return err
 			}
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -73,7 +73,39 @@ type Marshaler struct {
 
 	// Whether to use the original (.proto) name for fields.
 	OrigName bool
+
+	// A custom URL resolver to use when marshaling Any messages to JSON.
+	// If unset, DefaultUrlResolver is used.
+	UrlResolver UrlResolver
 }
+
+// UrlResolver resolves a type URL into an instance of the associated message.
+// This is used for serializing and de-serializing Any messages to and from
+// JSON.
+type UrlResolver interface {
+	Resolve(typeUrl string) (proto.Message, error)
+}
+
+type defaultResolver struct{}
+
+func (defaultResolver) Resolve(typeUrl string) (proto.Message, error) {
+	// Only the part of typeUrl after the last slash is relevant.
+	mname := typeUrl
+	if slash := strings.LastIndex(mname, "/"); slash >= 0 {
+		mname = mname[slash+1:]
+	}
+	mt := proto.MessageType(mname)
+	if mt == nil {
+		return nil, fmt.Errorf("unknown message type %q", mname)
+	}
+	return reflect.New(mt.Elem()).Interface().(proto.Message), nil
+}
+
+// DefaultUrlResolver can resolve any generated protobuf message that is
+// registered via proto.RegisterType. The fully-qualified message name is first
+// extracted from the type URL (e.g. the last path element). The resolver then
+// invokes proto.MessageType(string) to create a message instance for that name.
+var DefaultUrlResolver UrlResolver = defaultResolver{}
 
 // JSONPBMarshaler is implemented by protobuf messages that customize the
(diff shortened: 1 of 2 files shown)


jhump:
@cybrcodr, can you take a look at this? This is to provide better support for dynamic messages when serializing to and de-serializing from JSON.

jhump:
@cybrcodr: ping. Any opinion on this? Or is there someone else I should ask to take a look?

cybrcodr:
Sorry, I won't be available for a week.  @zombiezen, can you take a look?

@jhump, I'm guessing that in your use case(s), there are some Go types that represent these dynamic messages, implements proto.Message and are probably constructed via some descriptors.  Is that correct?  And hence w/o this feature in the library, one will need to register these Go types with corresponding proto name during runtime before being able to marshal/unmarshal.  The bad part I guess is that proto.RegisterType can only register one proto.Message with associated proto name.  I don't know what the effect would be for dynamic messages that are not globally registered, but perhaps that is already the case anyways.

jhump:
> I'm guessing that in your use case(s), there are some Go types that represent these dynamic messages, implements proto.Message and are probably constructed via some descriptors. Is that correct?

In fact there is just one Go type: [*dynamic.Message](http://godoc.org/github.com/jhump/protoreflect/dynamic).

> The bad part I guess is that proto.RegisterType can only register one proto.Message with associated proto name.

That is one issue. But worse, in this particular case, is that it assumes the mapping is bidirectional and that a zero value for the registered type is a sufficient instance of a given message name. With a dynamic message, the zero value is useless because it has no descriptor information.

Also, the fact that it is static/global is problematic. With dynamic messages, I could have two goroutines processing stuff that are potentially using different descriptor databases (they could even have an overlapping set of types defined in those databases, with a given type potentially varying from one database to the next).

So an ability to have local descriptor databases be used by the standard library and further to let those databases control message construction (e.g. setting necessary fields to preserve dynamic message invariants). That's basically what this pull request allows.

jhump:
Ping @zombiezen or @cybrcodr

zombiezen:
While we don't explicitly try to forbid alternate implementations of `proto.Message`, we don't really provide compatibility guarantees that they will work.  Adding this interface makes me a little nervous because it would allow implementations to return `proto.Message` values where not all the assumptions about the Go type to proto descriptor mapping hold.

If we were to accept a change like this, it would also have to alter the `ptypes` package to allow custom resolution of Anys, not just JSON.

I'm deferring to @cybrcodr who gets back from vacation this week.

jhump:
> If we were to accept a change like this, it would also have to alter the ptypes package to allow custom resolution of Anys, not just JSON.

I understand the desire to do this perhaps for consistency. But it is much less necessary. It is easy for an alternate package (that produces non-standard implementations of `proto.Message`) to provide its own `Any` [unmarshalling](https://github.com/jhump/protoreflect/blob/master/dynamic/message_registry.go#L351) since that is not a recursive operation and the entry point will be code that has a reference to an `Any` (vs. some other kind of message that may carry additional context for resolution of types therein).

The recursive nature of JSON (e.g. must handle a message with an `Any` that therein contains a message with an `Any`, etc) is what makes it essentially impossible to correctly handle purely in a 3rd-party package without that package basically forking the entirety of `jsonpb`.

jhump:
@cybrcodr, back from vacation yet?

cybrcodr:
Sorry, I've been swamped on other work.

Here are my thoughts on this.  I just read up again on the Any spec -- https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Any

It seems to me that type_url is a URL/resource name for describing the type of the serialized message.  Afaik, I don't think current practice/usages rely on having an external valid URL since that is not implemented.  But if it were to be implemented, would there still be a need for your proposed API?

With current implementation, would you not be able to achieve the same effect on your end instead by doing the resolution via URL through JSONPB{Marshaler,Unmarshaler}?

cybrcodr:
Sigh.  I think the reference to the Any spec in my previous posting is meant for binary serialization.  I'm not really sure about JSON serialization which is what your PR is about.

From https://developers.google.com/protocol-buffers/docs/proto3#json, it simply states ...

> If the Any contains a value that has a special JSON mapping, it will be converted as follows: {"@type": xxx, "value": yyy}. Otherwise, the value will be converted into a JSON object, and the "@type" field will be inserted to indicate the actual data type.

I'd like to think that the rules for binary serialization should still apply with the difference that the value is in JSON vs binary.

jhump:
> But if it were to be implemented, would there still be a need for your proposed API?

Yes, the need would still be there. Otherwise, the only solution for dynamic descriptors is to stand up infrastructure to serve these protos. What I am actually trying to do is to support `Any` types with a dynamic client that uses server reflection to download types. And there is certainly no guarantee that any of these types might be hosted at a real URL, especially due to how `Any` types are currently implemented.

Even if the URL resolution were supported, I would think many organizations would intentionally choose to *disable* it, for a variety of reasons -- largely around cost and availability of needing to stand up infrastructure to serve their own protos as well as to cache 3rd party protos (to improve availability for backends that might be behind lower-reliability links). But I can also see a security argument.

FWIW, I have actually written code in a library that [resolves messages via URL](https://godoc.org/github.com/jhump/protoreflect/dynamic#TypeFetcher). Doing so requires support for dynamic messages in fact (which is the whole point of my library). I wrote it based on the spec, but feel it will never get used since not even the google protos (google.protobuf...) are available at their ostensible URL (https://type.googleapis.com/google.protobuf.Timestamp for example).

jhump:
So, in a nutshell, I want the application doing serialization/de-serialization to be able to provide the description of the messages -- instead of relying on them to have been compiled and linked into the binary (current implementation) or relying on a server that serve the descriptions (per the spec). 

That's what this PR does -- allows the application to provide logic that resolves URLs into messages.

cybrcodr:
@dsnet and I discussed this as well as what it means for proto binary serialization.  In general, our current proto implementation relies on global registration.  Ideally, it'll be nice to have some way to delegate serialization w/o knowledge of global registration.

Now, I think proto binary serialization does not really have the same issue when it comes to marshaling/unmarshaling Any containing unregistered message because the unregistered message is simply a byte slice value that does not require any further processing.  In JSON marshaling/unmarshaling, the value is being converted from one form to another.

So, I think your request for a resolver does make sense.  I can't think of other types apart from Any that has such need.  Do correct me if I'm wrong.

I'll review this PR.

jsonpb/jsonpb.go
```diff
@@ -73,8 +73,40 @@ type Marshaler struct {
...
+
+	// A custom URL resolver to use when marshaling Any messages to JSON.
+	// If unset, DefaultUrlResolver is used.
+	UrlResolver UrlResolver
+}
+
+// UrlResolver resolves a type URL into an instance of the associated message.
+// This is used for serializing and de-serializing Any messages to and from
+// JSON.
+type UrlResolver interface {
```
cybrcodr: I'd suggest naming this TypeResolver or AnyResolver to be more specific.  And same for the field names in Marshaler and Unmarshaler.

jsonpb/jsonpb.go
```diff
@@ -73,8 +73,40 @@ type Marshaler struct {
...
+		return nil, fmt.Errorf("unknown message type %q", mname)
+	}
+	return reflect.New(mt.Elem()).Interface().(proto.Message), nil
 }
 
+// DefaultUrlResolver can resolve any generated protobuf message that is
+// registered via proto.RegisterType. The fully-qualified message name is first
+// extracted from the type URL (e.g. the last path element). The resolver then
+// invokes proto.MessageType(string) to create a message instance for that name.
+var DefaultUrlResolver UrlResolver = defaultResolver{}
```
cybrcodr: I don't think there is a need to export an instance of this.  If a custom resolver wants to use the global registry, then it can do so on its own.

jsonpb/jsonpb.go
```diff
@@ -679,16 +714,15 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			}
 			target.Field(0).SetString(turl)
 
-			mname := turl
-			if slash := strings.LastIndex(mname, "/"); slash >= 0 {
-				mname = mname[slash+1:]
+			resolver := u.UrlResolver
+			if resolver == nil {
+				resolver = DefaultUrlResolver
```
cybrcodr: How about just constructing the defaultResolver here and not need a package level var?

jsonpb/jsonpb_test.go
```diff
@@ -721,6 +721,61 @@ func TestUnmarshalingBadInput(t *testing.T) {
...
 
+type funcResolver func(turl string) (proto.Message, error)
+
+func (fn funcResolver) Resolve(turl string) (proto.Message, error) {
+	return fn(turl)
+}
+
+func TestAnyWithCustomResolver(t *testing.T) {
+	resolveCount := 0
+	resolver := funcResolver(func(turl string) (proto.Message, error) {
```
cybrcodr: Minor.  Capture turl and verify that it got the expected TypeUrl value.

jsonpb/jsonpb_test.go
```diff
@@ -721,6 +721,61 @@ func TestUnmarshalingBadInput(t *testing.T) {
...
+
+	m := Marshaler{UrlResolver: resolver}
+	js, err := m.MarshalToString(any)
+	if err != nil {
+		t.Errorf("an unexpected error occurred when marshaling any to JSON: %v", err)
+	}
+	if resolveCount != 1 {
+		t.Errorf("custom resolver was not invoked during marshaling")
+	}
+	wanted := `{"@type":"https://foobar.com/some.random.MessageKind","oBool":true,"oInt64":"1020304","oString":"foobar","oBytes":"AQIDBA=="}`
```
cybrcodr: Shouldn't the JSON keys be sorted?

jhump: If the object marshaled is a map, then the keys are sorted (since maps have no defined order, but JSON output wants to be deterministic). In this case, however, the object being marshaled is a struct, so it's fields are marshaled in declaration order.

jsonpb/jsonpb.go
jhump: Done. Went with `AnyResolver` since `Any` is the only thing this is used for.

jsonpb/jsonpb.go
jhump: Removed.

jsonpb/jsonpb.go
jhump: Refactored to just invoke a default resolve function.

jsonpb/jsonpb_test.go
jhump: Done

jhump:
Thanks for the review, @cybrcodr! I think I addressed all of your comments. Please take a look at your earlier convenience.

jsonpb/jsonpb_test.go
cybrcodr: Ah, yes.  I got confused looking at code section that marshals via JSONPBMarshaler -- https://github.com/golang/protobuf/blob/master/jsonpb/jsonpb.go#L144.  Thanks for pointing that out.

awalterschulze:
I think this ties nicely up with my issue on https://github.com/golang/protobuf/issues/268
If we now have a resolver for any, that means there is one less reason to Register in the first place.

Issue #411, opened 2017-08-14
proto files mirrored in ptypes get registered with wrong file name


jhump:
According to protobuf doc, the way to import well-known types into a proto file is via the "google/protobuf" path (which should typically resolve to an `include` directory included with protoc):

```
import "google/protobuf/any.proto";
import "google/protobuf/duration.proto";
import "google/protobuf/empty.proto";
import "google/protobuf/struct.proto";
import "google/protobuf/timestamp.proto";
import "google/protobuf/wrappers.proto";
```

However, some (but not all) of these proto files get registered at runtime with an incorrect path. So if using descriptors and the `proto.FileDescriptor(path)` function to access them, trying to crawl a file that imports via these standard paths will be unable to locate the actual descriptors for these files.

The result is that descriptors can only be loaded if the proto files actually do this instead:

```
import "github.com/golang/protobuf/ptypes/any/any.proto";
import "github.com/golang/protobuf/ptypes/duration/duration.proto";
import "github.com/golang/protobuf/ptypes/empty/empty.proto";
import "github.com/golang/protobuf/ptypes/struct/struct.proto";
import "github.com/golang/protobuf/ptypes/timestamp/timestamp.proto";
import "github.com/golang/protobuf/ptypes/wrappers/wrappers.proto";
```

This obviously doesn't work for protos that are meant to be cross-platform since only Go developers are likely to do this. Protos designed by other service authors will certainly use the "google/protobuf" import path instead.

Interestingly, only the `ptypes` folder does this. The generated code for `descriptor.proto` and `plugin.proto` (in `protoc-gen-go/descriptor` and `protoc-gen-go/plugin`, respectively) omits the source proto files. Similarly, the generated files at https://github.com/google/go-genproto/tree/master/protobuf omit the source proto files and have the correct registration path (at least they do as of https://github.com/google/go-genproto/pull/31).

I've developed a fairly well-featured library using descriptors and dynamic messages, and I've had to work around this in a way that seems a bit gross:
https://github.com/jhump/protoreflect/blob/de08c439017fa0995864581656b240956b863f1b/desc/descriptor.go#L1573

jhump:
This should be fixed as of #412.

PR #412, opened 2017-08-14
fix ptypes file paths

jhump:
This should address #411.

With this change, the proto sources are no longer mirrored in the "ptypes" directory (consistent with the treatment of other `google/protobuf` sources that are included with protoc). The script that generates the protos does so using paths that will ensure the generated code registers their file descriptors correctly (e.g. registered at `"google/protobuf/any.proto"` instead of `"github.com/golang/protobuf/ptypes/any/any.proto"`.

While testing, I re-ran code gen for other files in the repo which showed that `descriptor.pb.go` was slightly out of date. So that's included in here, too, as a separate commit.


jhump:
@zombiezen, is this something you could review?

jhump:
> Having a copy of them in this repository makes it easier to identify the deltas in case there are any.

Would it be valuable to also check in copies of `descriptor.proto` and `compiler/plugin.proto`?

jhump:
@cybrcodr, how does it look now?

jhump:
> The makefiles for descriptor.proto and plugin.proto need some improvement

@cybrcodr, yep, they are kind of awful. I had to add sym links in my home directory to make it happy. But I felt it was worth a tiny change for consistency and didn't want to shave the yak actually fixing them. Thanks for merging!

IMO, the `regen.sh` script would ideally be in the top directory of this repo and use more-or-less the same approach for re-generating all protos in the repo (including the test protos).

PR #415, opened 2017-08-17
Add a callback to `jsonpb.Unmarshaler` to report unknown fields.

jmillikin-stripe:
This is useful when parsing JSON that is expected to exactly match a
protobuf schema, but is _permitted_ to drift slightly. In this case
unknown fields can be reported to higher-level code to decide whether
the parse succeeded.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -590,6 +590,10 @@ type Unmarshaler struct {
 	// Whether to allow messages to contain unknown fields, as opposed to
 	// failing to unmarshal.
 	AllowUnknownFields bool
+
+	// If non-nil, AllowedUnknownField will be called for each unknown field
+	// encountered and allowed during the unmarshalling.
+	AllowedUnknownField func(fieldName string)
 }
 
 // UnmarshalNext unmarshals the next protocol buffer from a JSON object stream.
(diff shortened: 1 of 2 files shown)


dsnet:
\cc @cybrcodr 

Related to golang/go#15314

cybrcodr:
Here are my thoughts on this PR.  This requires that user sets AllowUnknownField=true for the handler to be called.  This means if user sets the handler and didn't set AllowUnknownField to true, nothing happens.  I talked to @dsnet regarding this and we have differing opinions on it.

Also, the handler you're proposing is simply given the field name.  I'd expect that it would be more useful if it is allowed to do more with the input value, and hence I'd also suggest naming it as ResolveUnknownField or HandleUnknownField, then again perhaps it can be an interface similar to  AnyResolver.

Anyways, the package does already allow for custom unmarshaling of message if the message implements jsonpb.UnmarshalJSONPB.  This means that the custom unmarshaler can possibly detect for unknown fields.  I understand that it may require a bit more work on your end, but it also allows us time to think of how to handle this situation more appropriately.

dsnet:
The JSON<->PB specification explicitly says that handling of unknown fields is failure:
> JSON conversion requires type information in both directions (from proto to JSON and from JSON to proto). If there are unknown fields, there is no way to preserve them and the conversion should fail.

Thus, unless the protobuf team changes their opinion about unknown JSON fields, we're not going to deviate in behavior from the other languages.

PR #418, opened 2017-08-26
proto: use sync.Pool for temporary Buffer types

irfansharif:
Lower GC pressure for whenever temporary Buffer types are allocated,
evidently it's an issue grpc/grpc-go ran into in the past. Additionally
remove the TODOs left regarding exactly this.

+cc https://github.com/grpc/grpc-go/pull/1478#issuecomment-325058960, https://github.com/gogo/protobuf/pull/328

### DIFF
--- a/proto/decode.go
+++ b/proto/decode.go
@@ -409,7 +409,17 @@ func UnmarshalMerge(buf []byte, pb Message) error {
 	if u, ok := pb.(Unmarshaler); ok {
 		return u.Unmarshal(buf)
 	}
-	return NewBuffer(buf).Unmarshal(pb)
+
+	b := bufPool.Get().(*Buffer)
+	b.Reset()
+	b.SetBuf(buf)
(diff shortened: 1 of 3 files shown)


irfansharif:
+cc @zombiezen (or @cybrcodr if back from vacation).

dsnet:
Thanks for the PR. I'm going to reject this (not because there's anything implicitly wrong with it), but because I'm in the process of exporting a lot of internal changes we've been making to proto. These changes completely rewrite the logic in decode.go and encode.go. So to avoid merge conflicts, let's avoid this optimization for now.

irfansharif:
Thanks for responding. Any ETA on when those changes will get pulled in here? Is it available publicly? (Still think this tiny ~50 line diff is fine, it's barely a merge conflict. The existing workaround I'm trying to replace in https://github.com/grpc/grpc-go/pull/1478#issuecomment-325075939 is non-ideal.)

PR #420, opened 2017-08-28
expose accessor for value of unregistered extensions [master]

jhump:


jhump:
@dsnet: Maybe this is a preferable solution to #385 than my previous PR (#386)?

jhump:
ping

jhump:
@dsnet, any chance you could take a quick look?

jhump:
@dsnet, ping

dsnet:
While I would rather see a complete API for #364 come to fruition to accomplish this. I'm okay with adding this API for the short-term. I've come to convince myself that we need to eventually create a v2 of the proto package that cleans out many of the functions.

jhump:
@dsnet, if you are okay with this as a short-term solution, I have just prepared the patch against the `dev` branch. See #483. If you have a chance, please take a look at the code and let me know if there's anything else I can do to get this merged (ideally, both PRs, so the patch is in both dev and master branches).

jhump:
I pushed a commit to #483, but have not updated this branch yet with the same change.

@dsnet, if we can get #483 merged to `dev`, I was hoping to get this branch into the same form (just re-based against `master`) and then merge this, too. Will that be okay?

dsnet:
> I was hoping to get this branch into the same form (just re-based against master) and then merge this, too. Will that be okay?

I discussed this with the team, and we decided to adopt a policy of not merging `dev` features into `master` unless it is a critical bug or a security issue. The reason is to set a clear segregation between what gets merged over, otherwise it becomes a slippery slope and miscellaneous commits get cherry picked into `master`. Later, merging of the two branches becomes a nightmare because of all the subtle differences. It was already a challenging task just to merge Google's internal changes up to the `dev` branch and I would like to avoid that situation again.

We will merge `dev` into `master` on June 1st. Until then, you will need to vendor the `dev` branch to make use of it. This situation is pretty awful, and the lack of good dependency management in doesn't help the situation.

Fortunately, some of the "breaking" changes in `dev` will help lay the groundwork to avoid a situation like this again. The reason we can't merge `dev` into `master` is because people depend on behavior of protos that were never guaranteed by the package. For example, we added an "XXX_NoUnkeyedLiterals" field to force users to use keyed struct literals.

jhump:
@dsnet, I'm a little confused by this approach. It seems like on June 1, you can trivially resolve any and all merge conflicts by just forcing `master` to look *exactly* like the `dev` branch post-merge. After all, there is nothing being merged to `master` that isn't also already merged to `dev`, right? Based on that, it seems like you could accept patches to `master` as long as an equivalent patch had already been merged to `dev`.

dsnet:
I want to avoid a force merge. I did that on a different repository I manage, and users vocalized both publicly and privately that it broke their workflow because they were pinning specific commits.

jhump:
@dsnet, what is a force merge? Did you mean force push -- like re-write master to *be* dev? If so, that's not at all what I was suggesting. What I was suggesting would have a plain and simple history, so people that pin particular commits would be safe. It would be just as if you had done a normal merge between the two branches, but the actual state of the tree on master after the merge is identical to the head (at merge time) of the dev branch. (Which, is in fact, a perfectly valid result of resolving merge conflicts: nothing sneaky going on under the hood in git with this approach.)

dsnet:
> what is a force merge? Did you mean force push -- like re-write master to be dev? 

Yes, that's what I thought you were suggesting. I see now that your suggesting just switching to master to be like dev using a single commit, but the downside is that the code history between master and dev is lost.

Even setting aside that downside, what would the policy be for cherrypicking changes? What if the change depends on logic that only exists on the dev branch?

The reason we leaned towards a policy of no cherry picks except critical and security bug fixes was to avoid this slipper slope.

PR #425, opened 2017-09-12
API Endpoint for micro API

asim:
This PR adds the ability to define HTTP endpoints for an RPC service. The API endpoint is metadata registered in service discovery. The micro API reads this metadata and generates routes based on it.


asim:
Wrong repo. Sorry.

PR #427, opened 2017-09-14
fix interpretation of default bytes values

### LINKED ISSUE #426: Default values for fields of type "bytes" are wrong
When protoc builds the descriptors that are passed to protoc-gen-go, it *encodes* default values for bytes fields: https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/parser.cc#L1202

However, the generated Go code is not correctly decoding the values and instead is using the values as-is.

----

Example of problem:

*t.proto*
```proto
syntax = "proto2";
message Test {
  optional bytes b = 1 [default = "\0\1\2\3\4\5\6\7fubar!"];
}
```

*t_test.go*
```go
package t

import (
    "reflect"
    "testing"
)

func TestDefaultBytes(t *testing.T) {
    var msg *Test
    actual := msg.GetB()
    expected := []byte{0,1,2,3,4,5,6,7,'f','u','b','a','r','!'}
    if !reflect.DeepEqual(actual, expected) {
        t.Errorf("Wrong default! want: %q, got %q", string(expected), string(actual))
    }
}
```

The resulting error is this:
```
Wrong default! want: "\x00\x01\x02\x03\x04\x05\x06\afubar!", got "\\000\\001\\002\\003\\004\\005\\006\\007fubar!"
```

jhump:
Fixes #426

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -44,6 +44,7 @@ import (
 	"go/parser"
 	"go/printer"
 	"go/token"
+	"io"
 	"log"
 	"os"
 	"path"
@@ -1984,7 +1985,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		case typename == "string":
 			def = strconv.Quote(def)
 		case typename == "[]byte":
-			def = "[]byte(" + strconv.Quote(def) + ")"
+			def = "[]byte(" + strconv.Quote(unescape(def)) + ")"
 			kind = "var "
 		case def == "inf", def == "-inf", def == "nan":
 			// These names are known to, and defined by, the protocol language.
@@ -2506,6 +2507,97 @@ func (g *Generator) generateMessage(message *Descriptor) {
 	}
 
 	g.addInitf("%s.RegisterType((*%s)(nil), %q)", g.Pkg["proto"], ccTypeName, fullName)
+}
+
+func unescape(s string) string {
+	// NB: Sadly, we can't use strconv.Unquote because protoc will escape both
+	// single and double quotes, but strconv.Unquote only allows one or the
+	// other (based on actual surrounding quotes of its input argument).
+	s, err := doUnescape(s)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	return s
+}
+
+func doUnescape(s string) (string, error) {
+	// reverses the "C" escaping that protoc does for default values of bytes fields
+	buf := bytes.NewBuffer(make([]byte, 0, len(s)))
+	in := strings.NewReader(s)
+	for {
+		b, err := in.ReadByte()
+		if err == io.EOF {
+			// Done!
+			return buf.String(), nil
+		} else if err != nil {
+			return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+		}
+		if b == '\\' {
+			b, err := in.ReadByte()
+			if err != nil {
+				return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+			}
+			switch b {
+			case 'a':
+				buf.WriteByte('\a')
+			case 'b':
+				buf.WriteByte('\b')
+			case 'f':
+				buf.WriteByte('\f')
+			case 'n':
+				buf.WriteByte('\n')
+			case 'r':
+				buf.WriteByte('\r')
+			case 't':
+				buf.WriteByte('\t')
+			case 'v':
+				buf.WriteByte('\v')
+			case '\\':
+				buf.WriteByte('\\')
+			case '"':
+				buf.WriteByte('"')
+			case '\'':
+				buf.WriteByte('\'')
+			case '?':
+				buf.WriteByte('?')
+			case 'x':
+				h1, err := in.ReadByte()
+				if err != nil {
+					return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+				}
+				h2, err := in.ReadByte()
+				if err != nil {
+					return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+				}
+				c, err := strconv.ParseUint(string([]byte{h1, h2}), 16, 8)
+				if err != nil {
+					return "", fmt.Errorf("error decoding hex escape in constant bytes %q: %s", s, err)
+				}
+				buf.WriteByte(byte(c))
+			default:
+				if b >= '0' && b <= '7' {
+					o1 := b
+					o2, err := in.ReadByte()
+					if err != nil {
+						return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+					}
+					o3, err := in.ReadByte()
+					if err != nil {
+						return "", fmt.Errorf("error decoding constant bytes %q: %s", s, err)
+					}
+					c, err := strconv.ParseUint(string([]byte{o1, o2, o3}), 8, 8)
+					if err != nil {
+						return "", fmt.Errorf("error decoding octal escape in constant bytes %q: %s", s, err)
+					}
+					buf.WriteByte(byte(c))
+				} else {
+					return "", fmt.Errorf("invalid escape sequence in constant bytes %q", s)
+				}
+			}
+		} else {
+			buf.WriteByte(b)
+		}
+	}
 }
 
 func (g *Generator) generateExtension(ext *ExtensionDescriptor) {
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -2513,6 +2513,8 @@ var escapeChars = [256]byte{
 }
 
 // unescape reverses the "C" escaping that protoc does for default values of bytes fields.
+// It is best effort in that it effectively ignores malformed input. Seemingly invalid escape
+// sequences are conveyed, unmodified, into the decoded result.
 func unescape(s string) string {
 	// NB: Sadly, we can't use strconv.Unquote because protoc will escape both
 	// single and double quotes, but strconv.Unquote only allows one or the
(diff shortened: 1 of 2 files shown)


jhump:
@dsnet or @cybrcodr: any chance this can get a review?

protoc-gen-go/generator/name_test.go
```diff
@@ -83,3 +84,39 @@ func TestGoPackageOption(t *testing.T) {
...
+		{"\\0", "", "EOF"},
+		{"\\01", "", "EOF"},
+	}
+	for _, tc := range tests {
+		s, err := doUnescape(tc.in)
+		if tc.err != "" {
+			if err == nil {
+				t.Errorf("expecting error unescaping %q but it succeeded", tc.in)
+			} else if !strings.Contains(err.Error(), tc.err) {
+				t.Errorf("wrong error unescaping %q: want %q; got %q", tc.in, tc.err, err.Error())
```
dsnet: t.Errorf("doEscape(%q) = %q, want %q")

protoc-gen-go/generator/name_test.go
```diff
@@ -83,3 +84,39 @@ func TestGoPackageOption(t *testing.T) {
...
+			}
+		} else if s != tc.out {
+			t.Errorf("wrong result unescaping %q: want %q; got %q", tc.in, tc.out, s)
```
dsnet: t.Errorf("doEscape(%q) = %q, want %q")

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2509,97 @@ func (g *Generator) generateMessage(message *Descriptor) {
 	g.addInitf("%s.RegisterType((*%s)(nil), %q)", g.Pkg["proto"], ccTypeName, fullName)
 }
 
+func unescape(s string) string {
+	// NB: Sadly, we can't use strconv.Unquote because protoc will escape both
+	// single and double quotes, but strconv.Unquote only allows one or the
+	// other (based on actual surrounding quotes of its input argument).
+	s, err := doUnescape(s)
+	if err != nil {
+		log.Fatal(err.Error())
```
dsnet: `log.Fatal(err)`

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2509,97 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+	return s
+}
+
+func doUnescape(s string) (string, error) {
+	// reverses the "C" escaping that protoc does for default values of bytes fields
+	buf := bytes.NewBuffer(make([]byte, 0, len(s)))
+	in := strings.NewReader(s)
+	for {
+		b, err := in.ReadByte()
+		if err == io.EOF {
```
dsnet: Fold `err == io.EOF` case under the `err != nil` case.

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2509,97 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+	// single and double quotes, but strconv.Unquote only allows one or the
+	// other (based on actual surrounding quotes of its input argument).
+	s, err := doUnescape(s)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	return s
+}
+
+func doUnescape(s string) (string, error) {
```
dsnet: This seems much more complicated than necessary.

See this version:
https://play.golang.org/p/DHt_EPXYyt

dsnet: I did not test mine, but you get the idea.

dsnet: Also, according to https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences,
the octal sequence may be be less than 3 octal digits.

Also, should we support `\e` or `\Uhhhhhhhh` or `\uhhhh`? I don't know if protoc even generates those.

protoc-gen-go/generator/name_test.go
```diff
@@ -83,3 +84,39 @@ func TestGoPackageOption(t *testing.T) {
 		}
 	}
 }
+
+func TestUnescape(t *testing.T) {
+	tests := []struct {
+		in  string
+		out string
+		err string
```
dsnet: Just check if it errored or not. String parsing on errors is generally frowned upon.

protoc-gen-go/generator/generator.go
dsnet: Also, what happens if the input string is non-ASCII. Does protoc ever generate that?

jhump: > I don't know if protoc even generates those.

The precise implementation of protoc currently is that it will never emit unicode escapes or '\e'`. In fact, it only emits octal ecapes, and they are not variable length (always 3 chars).
https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/strutil.cc#L590

I was worried about coupling this too tightly to the current implementation (don't want it to break if protoc impl changes), which is why I added code for hex escapes. But since the field's type is `bytes`, not `string`, it seems highly unlikely that it should be trying to support unicode escapes (and, by extension, UTF-8 encoding).

> Also, what happens if the input string is non-ASCII. Does protoc ever generate that?

Currently, protoc does not generate such a string. But, if it did, the bytes of non-ascii characters (as long as they don't include `\`) would be conveyed in the value, which seems correct given the field's type is `bytes`.

dsnet: Sounds good. We can revisit for non-ASCII and unicode runes if it's an issue in the future.

protoc-gen-go/generator/generator.go
```diff
@@ -2520,84 +2520,70 @@ func unescape(s string) string {
...
+			out = append(out, byte(v))
+			s = s[4:]
+			continue
+		}
+
+		// octal escape
+		if '0' <= s[1] && s[1] <= '7' {
+			// can vary from 1 to 3 octal digits. E.g., "\0" "\40" or "\164"
+			// so consume up to 2 more bytes or up to end-of-string
+			end := len(s) - 2
```
dsnet: Could be slightly simpler yet:
```go
n := len(s[1:]) - strings.TrimLeft(s[1:], "01234567")
if n > 3 {
	n = 3
}
v, err := strconv.ParseUint(s[1:1+n], 8, 8)
if err != nil {
	return "", err
}
out = append(out, byte(v))
s = s[1+n:]
continue
```

protoc-gen-go/generator/name_test.go
```diff
@@ -87,36 +86,36 @@ func TestGoPackageOption(t *testing.T) {
...
 	}{
 		// successful cases, including all kinds of escapes
-		{"", "", ""},
-		{"foo bar baz frob nitz", "foo bar baz frob nitz", ""},
-		{"\\000\\001\\002\\003\\004\\005\\006\\007", string([]byte{0, 1, 2, 3, 4, 5, 6, 7}), ""},
-		{"\\a\\b\\f\\n\\r\\t\\v\\\\\\?\\'\\\"", "\a\b\f\n\r\t\v\\?'\"", ""},
-		{"\\x10\\x20\\x30\\x40\\x50\\x60\\x70\\x80", "\x10\x20\x30\x40\x50\x60\x70\x80", ""},
+		{"", "", false},
+		{"foo bar baz frob nitz", "foo bar baz frob nitz", false},
+		{"\\000\\001\\002\\003\\004\\005\\006\\007", string([]byte{0, 1, 2, 3, 4, 5, 6, 7}), false},
```
dsnet: The double escaping is a little hard to read. Can you use raw-string literals?
```go
`\000\001\002\003\004\005\006\007`
```

protoc-gen-go/generator/generator.go
jhump: I just pushed a change so that it won't fail on bad input and simply propagate malformed escape sequences into the resulting bytes. WDYT?

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2508,69 @@ func (g *Generator) generateMessage(message *Descriptor) {
 	g.addInitf("%s.RegisterType((*%s)(nil), %q)", g.Pkg["proto"], ccTypeName, fullName)
 }
 
+var escapeChars = [256]byte{
+	'a': '\a', 'b': '\b', 'f': '\f', 'n': '\n', 'r': '\r', 't': '\t', 'v': '\v', '\\': '\\', '"': '"', '\'': '\'', '?': '?',
+}
+
+// unescape reverses the "C" escaping that protoc does for default values of bytes fields.
+func unescape(s string) string {
```
dsnet: Comment that this is a best effort at unescaping.

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2508,69 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+			s = s[1:]
+		} else if c := escapeChars[s[1]]; c != 0 {
+			// escape sequence
+			out = append(out, c)
+			s = s[2:]
+		} else if s[1] == 'x' || s[1] == 'X' {
+			// hex escape, e.g. "\x80
+			if len(s) < 4 {
+				// too short to be valid
+				out = append(out, s[0], s[1])
```
dsnet: `out = append(out, s[:2]...)`

protoc-gen-go/generator/generator.go
```diff
@@ -2508,6 +2508,69 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+				s = s[2:]
+				continue
+			}
+			v, err := strconv.ParseUint(s[2:4], 16, 8)
+			if err != nil {
+				out = append(out, s[0], s[1], s[2], s[3])
```
dsnet: `out = append(out, s[:4]...)`

protoc-gen-go/generator/generator.go
jhump: Interesting. I knew you could do that w/ slices but didn't realize you could do that with `string` and compiler would implicitly regard it as `[]byte`

protoc-gen-go/generator/generator.go
jhump: done

protoc-gen-go/generator/generator.go
jhump: done

dsnet:
Latest changes LGTM. Should I submit?

protoc-gen-go/generator/name_test.go
jhump: Done. When I did this, it made it harder to easily see the difference between inputs and expected outputs. In several cases, the only difference was the kind of quote mark, which seemed overly subtle when reading. So I changed expected outputs in a few cases to use a []byte instead of string, to make the expected decoding more obvious.

jhump:
@dsnet, if LGTY, merge away!
Thanks for the quick review!

PR #429, opened 2017-09-18
Unmarshal a value which implements JSONPBUnmarshaler

vaporz:
Hi, our test cases are failed by #394 .  
We defined a struct which implements interface `JSONPBUnmarshaler`, and use it as a field in another message, for example:
```protobuf
message MyString {
  bool null = 1;
  string value = 2;
}
message Request {
  MyString name = 1;
}
```
`MyString` implements `JSONPBUnmarshaler `.

Before #394 , everything works fine.  
After #394 , if input json is `{"name": null}`, func `UnmarshalJSONPB` on `MyString` is not executed. Because it will return earlier:
```go
if string(inputValue) == "null" && targetType != reflect.TypeOf(&stpb.Value{}) {
	return nil   // <---returned here
}
```

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -673,7 +673,8 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 	if targetType.Kind() == reflect.Ptr {
 		// If input value is "null" and target is a pointer type, then the field should be treated as not set
 		// UNLESS the target is structpb.Value, in which case it should be set to structpb.NullValue.
-		if string(inputValue) == "null" && targetType != reflect.TypeOf(&stpb.Value{}) {
+		_, isJSONPBUnmarshaler := target.Interface().(JSONPBUnmarshaler)
+		if string(inputValue) == "null" && targetType != reflect.TypeOf(&stpb.Value{}) && !isJSONPBUnmarshaler {
 			return nil
 		}
 		target.Set(reflect.New(targetType.Elem()))
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -791,6 +791,19 @@ func TestUnmarshalJSONPBUnmarshaler(t *testing.T) {
 	}
 }
 
+func TestUnmarshalNullWithJSONPBUnmarshaler(t *testing.T) {
+	rawJson := `{"stringField":null}`
+	var ptrFieldMsg ptrFieldMessage
+	if err := Unmarshal(strings.NewReader(rawJson), &ptrFieldMsg); err != nil {
+		t.Errorf("unmarshal error: %v", err)
+	}
+
+	want := ptrFieldMessage{StringField: &stringField{IsSet: true, StringValue: "null"}}
+	if !proto.Equal(&ptrFieldMsg, &want) {
+		t.Errorf("unmarshal result StringField: got %v, want %v", ptrFieldMsg, want)
+	}
+}
+
 func TestUnmarshalAnyJSONPBUnmarshaler(t *testing.T) {
 	rawJson := `{ "@type": "blah.com/` + dynamicMessageName + `", "foo": "bar", "baz": [0, 1, 2, 3] }`
 	var got anypb.Any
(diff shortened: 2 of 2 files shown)


cybrcodr:
Curious, how do you make MyString implement JSONPBUnmarshaler?  Do you hand modify the generated pb.go file or add another file under the same generated package?

vaporz:
Hi, yes, we add another file under the same generated package.

jsonpb/jsonpb.go
```diff
@@ -673,7 +673,8 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 	if targetType.Kind() == reflect.Ptr {
 		// If input value is "null" and target is a pointer type, then the field should be treated as not set
 		// UNLESS the target is structpb.Value, in which case it should be set to structpb.NullValue.
-		if string(inputValue) == "null" && targetType != reflect.TypeOf(&stpb.Value{}) {
+		_, ok := target.Interface().(JSONPBUnmarshaler)
```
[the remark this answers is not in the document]
vaporz: OK

vaporz:
Hi, @cybrcodr , renamed "ok" to "isJSONPBUnmarshaler", and added a test.

jsonpb/jsonpb_test.go
```diff
@@ -789,6 +789,18 @@ func TestUnmarshalJSONPBUnmarshaler(t *testing.T) {
 	if msg.rawJson != rawJson {
 		t.Errorf("message contents not set correctly after unmarshalling JSON: got %s, wanted %s", msg.rawJson, rawJson)
 	}
+
+	rawJson = `{"stringField":null}`
+	var ptrFieldMsg ptrFieldMessage
+	if err := Unmarshal(strings.NewReader(rawJson), &ptrFieldMsg); err != nil {
+		t.Errorf("an unexpected error occurred when parsing into JSONPBUnmarshaler: %v", err)
+	}
+	if ptrFieldMsg.StringField == nil {
```
cybrcodr: May be more precise to compare with expected value.  Use proto.Equal to compare.

Error message can be something like "unmarshal result StringField: got %v, want %v"

jsonpb/jsonpb_test.go
```diff
@@ -789,6 +789,18 @@ func TestUnmarshalJSONPBUnmarshaler(t *testing.T) {
...
+		t.Errorf("ptr field which implements JSONPBUnmarshaler is nil after unmarshalling with JSON {\"stringField\":null}:")
+	}
+	if ptrFieldMsg.StringField != nil && ptrFieldMsg.StringField.IsSet == false {
```
cybrcodr: Not sure if you need the IsSet field.  But if you really want to keep it, the validation here should not need to include check for StringField != nil, i.e. it should be independent.

jsonpb/jsonpb_test.go
```diff
@@ -821,6 +833,45 @@ func init() {
...
+}
+
+func (s *stringField) String() string {
+	return s.StringValue
+}
+
+func (s *stringField) ProtoMessage() {
+}
+
+func (s *stringField) MarshalJSONPB(jm *Marshaler) ([]byte, error) {
```
cybrcodr: Optional.  This is not necessary since the test is only on JSONPBUnmarshaler.

jsonpb/jsonpb_test.go
```diff
@@ -789,6 +789,18 @@ func TestUnmarshalJSONPBUnmarshaler(t *testing.T) {
 	if msg.rawJson != rawJson {
 		t.Errorf("message contents not set correctly after unmarshalling JSON: got %s, wanted %s", msg.rawJson, rawJson)
 	}
+
```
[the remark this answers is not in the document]
vaporz: Done

jsonpb/jsonpb_test.go
```diff
@@ -789,6 +789,18 @@ func TestUnmarshalJSONPBUnmarshaler(t *testing.T) {
...
+	rawJson = `{"stringField":null}`
+	var ptrFieldMsg ptrFieldMessage
+	if err := Unmarshal(strings.NewReader(rawJson), &ptrFieldMsg); err != nil {
+		t.Errorf("an unexpected error occurred when parsing into JSONPBUnmarshaler: %v", err)
```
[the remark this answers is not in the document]
vaporz: Done

jsonpb/jsonpb_test.go
vaporz: Done

jsonpb/jsonpb_test.go
vaporz: Removed this validation, "proto.Equal" is enough, but "IsSet" is not removed, I'd like to use it to make sure `func (s *stringField) UnmarshalJSONPB` is executed.

jsonpb/jsonpb_test.go
vaporz: OK, removed

Issue #192, opened 2016-06-13
proposal: protoc-gen-go: support embedding a nested type


phifty:
Given the code

```
synatx = "proto3";

package test;

message Password {
    string hash = 1;
    string salt = 2;
}
message User {
    Password password = 1;
}
```

The generated `User` struct contains than a pointer to a password struct. Is it possible to embed the struct? The `User` would contain the `Password` (without a pointer) and the `Password` would be initialized along with the `User`.

zombiezen:
This isn't possible with the current code generator.  Protobuf schemas are independent of language, so it would be difficult to specify the semantics you want in a language-neutral way.

phifty:
I would be fine with a language-specific option. Something like `Password password = 1 [go_embed=true];`.

Xopherus:
what sort of use case would this be used for @phifty?

sarbash:
At least, to provide automatic binding for chi.Render
```go
sendRequest := &mailer.SendRequest{
	Sender: &mailer.Sender{},
	Mail:   &mailer.Mail{},
}
if err := render.Bind(r, sendRequest); err != nil {}
```

travisjeffery:
Wrote the plugin: https://github.com/golang/protobuf/pull/430

dsnet:
Adding support for embedding in the mainline proto repo breaks the semantics of protobufs. According to the proto language spec, there is a distinction between the zero value for a message and an unset message (even for proto3). Embedding provides no such distinctions.

dsnet:
We're going to reject this for the following reasons:

**1) Embedded values violates the semantics of protos.**

It is unclear in this proposal whether the OP intended for `Password` to be embedded or `*Password`. If the former, then we have to reject this proposal since it violates the semantics of zero-vs-unset messages in protos. I spoke to the protobuf team and we will maintain a hard stance on that requirement.

**2) Embedded pointers lead to brittle code.**

Suppose that the intended behavior was the later (i.e., `*Password`), then this leads to brittle code as it is trivial for someone to do `User.Salt` without realizing that this can result in a nil-pointer derefence on `User.Password`, which isn't visible in the expression. It's annoying enough with `Foo.Bar` where you need to consciously think whether `Foo` is nil or not. It's even worse when you need to think about whether some intermediate de-reference that is not even visible may be nil.

**3) Embedding forwards unintended methods.**

[...]

PR #430, opened 2017-09-20
Add plugin to support struct embedding

travisjeffery:
Fixes https://github.com/golang/protobuf/issues/192. 

The use case I hit is using wanting my Kubernetes custom resource to be defined in protobuf, which reference k8s's TypeMeta and ObjectMeta types, and those need to be embedded to work with various interfaces.

Let me know if you're interested, and I can keep going with tests, etc.

The API uses a custom option go_embed:

``` go
message MyResource {
  k8s.TypeMeta type_meta = 1 [(go_embed)=true]; 
  k8s.ObjectMeta object_meta = 2 [(go_embed)=true];
  string status = 3; 
}
```

### DIFF
--- a/protoc-gen-go/embedded/embedded.go
+++ b/protoc-gen-go/embedded/embedded.go
@@ -1,0 +1,123 @@
+package embedded
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


travisjeffery:
I signed it!

travisjeffery:
ah just found https://github.com/gogo/protobuf/

PR #435, opened 2017-09-30
Return both response and error for grpc unary calls

peter-edge:
The gRPC protocol is able to return both a valid response and an error, and `grpc.Invoke` handles this properly.

This is technically a backwards-incompatible change if someone were to rely on the response being nil:

```go
response, _ = foo.Bar(...)
if response == nil {
  ... 
}
```

However users should be checking if `err != nil` and existing behavior will not be effected.


paranoiacblack:
It's not safe to look at response if `Invoke` return a non-nil error. The response proto is most likely only partially filled since most of the errors from Invoke come either  before the response proto is modified or during unmarshal. See also https://github.com/grpc/grpc/issues/12824#issuecomment-335624435.

I think this change also makes the allocated memory for the response proto leak when there is an error. I think protos have a potential to be too large so avoiding allocation, like in the current implementation, seems better in that case, as well. Please re-open if you think there's a good use-case for looking at the response proto on a non-nil error.

Issue #433, opened 2017-09-26
protoc-gen-go: compiling nested extensions generates invalid code


occho:
When trying to use extensions nested in a `message`, a build error happens with a error message like "cannot convert nil to type bool."

In [generator.go][generator-go], I'm expecting the extension fields should reach [needsStar()][generator-go-1678], but actually they won't because [they have a parent and the parent's syntax is considered as proto3][generator-go-1674].

https://github.com/golang/protobuf/blob/130e6b02ab059e7b717a096f397c5b60111cae74/protoc-gen-go/generator/generator.go#L1674-L1680

[generator-go]: https://github.com/golang/protobuf/blob/130e6b02ab059e7b717a096f397c5b60111cae74/protoc-gen-go/generator/generator.go
[generator-go-1678]: https://github.com/golang/protobuf/blob/130e6b02ab059e7b717a096f397c5b60111cae74/protoc-gen-go/generator/generator.go#L1678
[generator-go-1674]: https://github.com/golang/protobuf/blob/130e6b02ab059e7b717a096f397c5b60111cae74/protoc-gen-go/generator/generator.go#L1674

# Versions

protoc version: libprotoc 3.3.2
golang/protobuf: using the current master https://github.com/golang/protobuf/tree/130e6b02ab059e7b717a096f397c5b60111cae74

# Command to reproduce the issue

```
~/go/src/github.com/occho/protobuf-test/test$ pwd
<home>/go/src/github.com/occho/protobuf-test/test
~/go/src/github.com/occho/protobuf-test/test$ ls
test.proto
~/go/src/github.com/occho/protobuf-test/test$ protoc --go_out=${GOPATH}/src test.proto
~/go/src/github.com/occho/protobuf-test/test$ go build
# github.com/occho/protobuf-test/test
./test.pb.go:41:23: cannot convert nil to type bool
./test.pb.go:50:25: cannot convert nil to type string
./test.pb.go:59:24: cannot convert nil to type int32
```

# File test.proto

```protobuf
syntax = "proto3";

package test;

import "google/protobuf/descriptor.proto";

option go_package = "github.com/occho/protobuf-test/test";

message Namespace {
  extend google.protobuf.FieldOptions {
    bool namespace_bool = 50001;
    string namespace_string = 50002;
    int32 namespace_int32 = 50003;
  }
}

extend google.protobuf.FieldOptions {
  bool my_bool = 51001;
  string my_string = 51002;
  int32 my_int32 = 51003;
}
```

[...]

PR #436, opened 2017-10-05
protoc-gen-go: fix field type mapping for extension fields

occho:
Fix passing a wrong argument to `generator.GoType`. `generator.GoType`
should take a message descriptor to which a field descriptor belongs.

Fixes: #433


jhump:
@dsnet, @zombiezen: this is a correctness issue in protoc-gen-go that has a really tiny and simple fix. Could one of you look at this or ping someone who can?

PR #437, opened 2017-10-05
proto: resuse buffers in proto

while-loop:
Completed some TODOs that wanted to reuse buffers. Included the sync.Pool struct to tackle task. I've only replaced buffers where they are used for encoding/decoding byte streams


while-loop:
I signed it!

dsnet:
Thanks for the PR. Unfortunately, the next iteration of protobuf performance in the `dev` branch actually side-steps usage of the `Buffer` type and is demonstrated to be faster.

PR #441, opened 2017-10-14
do not allow files that cross package boundaries

jhump:
protoc-gen-go tries to disallow multiple files if they cross package boundaries -- it wants to be invoked for a single package at a time, and with all files for that package at the same time.

But it doesn't do a very good job checking for this condition. It simply verifies that the *package name* for all files is the same, but it does not check that all files emit code to the same package path (which is what Go really considers to be the identity of a package).

That means I can supply different files that have the same package name but live in different folders, and the generated code is simply wrong. Because it assumes they are the same package (including same package path), references from one to the other will result in code that uses naked symbol references (no package qualifier/alias) and also not produce an import statement.

So this change will cause the plugin to issue an error message if the files it produces would be in different Go packages.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -296,23 +296,25 @@ func (d *FileDescriptor) goPackageOption() (impPath, pkg string, ok bool) {
 	return
 }
 
-// goPackageName returns the Go package name to use in the
-// generated Go file.  The result explicit reports whether the name
-// came from an option go_package statement.  If explicit is false,
-// the name was derived from the protocol buffer's package statement
+// goPackagePathAndName returns the Go import path and package name to use
+// in the generated Go file.  The result explicit reports whether the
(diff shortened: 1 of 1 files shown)


jhump:
@dsnet or @zombiezen: would either of you be the right person to take a look at this?

jhump:
cc: @sanjaybv

tamird:
@jhump I'd imagine this needs to go to the `dev` branch, since that's where all ongoing development is happening.

jhump:
@tamird, yes, I know. I just never bothered rebasing this. It isn't urgent. I left it open as a reminder to myself to rebase it after dev finally gets merged to master.

jhump:
I was pleasantly surprised that this rebased cleanly. But CLA bot got confused when I pushed the changes. Opening a different PR for this...

jhump:
See #527 instead

PR #451, opened 2017-10-30
grpc: generate idempotency level for service description

keelerh:
Protoc now supports an `idempotency_level` method option: https://github.com/google/protobuf/blob/9e745f7/src/google/protobuf/descriptor.proto#L630. The method option can be used to mark the idempotency of a gRPC method.

This change adds the idempotency level information to the generated service descriptions for use in e.g. gRPC server interceptors enforcing idempotency guarantees.

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -393,6 +393,7 @@ func (g *grpc) generateServerMethod(servName, fullServName string, method *pb.Me
 		g.P("info := &", grpcPkg, ".UnaryServerInfo{")
 		g.P("Server: srv,")
 		g.P("FullMethod: ", strconv.Quote(fmt.Sprintf("/%s/%s", fullServName, methName)), ",")
+		g.P("IdempotencyLevel: grpc.", method.GetOptions().GetIdempotencyLevel().String(), ",")
 		g.P("}")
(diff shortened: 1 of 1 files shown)


keelerh:
I signed it!

keelerh:
I signed it!

dsnet:
\cc  @dfawley. Can you comment whether this is desired or not?

@keelerh, all future work is being done now on the `dev` branch. Can you rebase to that branch? Thanks.

paranoiacblack:
Looking back at this, I didn't know about the options, but I also don't see a reason not to support it, here. This PR seems fine if you wanted to proceed @keelerh.

dfawley:
Am I missing something?  `IdempotencyLevel` is not currently a field in [`UnaryServerInfo`](https://godoc.org/google.golang.org/grpc#UnaryServerInfo).

If we're going to do something like this, we should ideally pull in the whole `MethodOptions` message and not just individual fields from it.

paranoiacblack:
So if I understand correctly, from the grpc-Go perspective this is not desired or particularly supported yet. I was reading about this option and see its utility, but the lack of support/uses mean we can wait on this change. I'll close this for now and we can revisit it if/when gRPC starts to expect this field to be set.

PR #453, opened 2017-11-06
fix unnecessary rounding to float64 precision when JSON-marshaling durations

dfawley:


dfawley:
Ping - this should be pretty trivial.

dsnet:
It seems Herbie is OOO.

puellanivis:
“For durations of one second or more, a non-zero value for the nanos field must be of the same sign as the seconds field. Must be from -999,999,999 to +999,999,999 inclusive.” https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration

It looks like you’re right, @awalterschulze . And, the negative duration test is a good idea in all cases. Additionally, a negative less-than-one-second test is also going to be important as this would currently write as “0.-100626394s” for example.

It looks like we will need to extract the sign and absolute value of both of the integers… some quick and dirty code (not intended as complete solution, I just think in code)…

```
sign := ""
if s < 0 {
  s = -s
  sign = "-"
}
if ns < 0 {
  ns = -ns
  sign = "-"
}
x := fmt.Sprintf("%s%d.%09ds", sign, s, ns)
```

Even though it’s more work, in the end, this should still work better (no power-of-ten rounding issues) and be faster than using floating points… (?)

awalterschulze:
I was just trying to merge this change into gogoprotobuf and ran my tests, which found the bug.  I am really glad to hear its not my tests, but an actual bug.  Looking forward to merge a fix when its ready.

PR #458, opened 2017-11-21
grpc: return truely nil interface{} if response payload is nil

jdef:
Related to https://github.com/thecodeteam/gocsi/pull/42

Generated code converts a `*struct` type into an `interface{}` type: when this happens, `nil` pointers are converted to non-`nil` `interface{}` objects leading to subtle errors for interceptors examining the response payload.

This change only sets the value of the returned `interface{}` if the response payload (pointer) is non-`nil`.


puellanivis:
Ah yes, the old “interface{} cast of nil with a concrete type is not == nil” kraken.

jdef:
cc @dsymonds

protoc-gen-go/grpc/grpc.go
```diff
@@ -394,8 +394,12 @@ func (g *grpc) generateServerMethod(servName, fullServName string, method *pb.Me
 		g.P("Server: srv,")
 		g.P("FullMethod: ", strconv.Quote(fmt.Sprintf("/%s/%s", fullServName, methName)), ",")
 		g.P("}")
-		g.P("handler := func(ctx ", contextPkg, ".Context, req interface{}) (interface{}, error) {")
-		g.P("return srv.(", servName, "Server).", methName, "(ctx, req.(*", inType, "))")
+		g.P("handler := func(ctx ", contextPkg, ".Context, req interface{}) (reply interface{}, err error) {")
+		g.P("r, err := srv.(", servName, "Server).", methName, "(ctx, req.(*", inType, "))")
+		g.P("if r != nil {")
+		g.P("reply = r")
+		g.P("}")
```
paranoiacblack: Instead of

```
r, err := srv.(GeneratedServer).Method(ctx, req.(*proto.Message))
if r != nil {
       reply = r
}
return
```

explicitly return `nil`

```
if r, err := srv.(GeneratedServer).Method(ctx, req.(*proto.Message)); r != nil {
        return r, err
} else {
        return nil, err
}
```

protoc-gen-go/grpc/grpc.go
```diff
@@ -394,8 +394,12 @@ func (g *grpc) generateServerMethod(servName, fullServName string, method *pb.Me
 		g.P("Server: srv,")
 		g.P("FullMethod: ", strconv.Quote(fmt.Sprintf("/%s/%s", fullServName, methName)), ",")
 		g.P("}")
-		g.P("handler := func(ctx ", contextPkg, ".Context, req interface{}) (interface{}, error) {")
-		g.P("return srv.(", servName, "Server).", methName, "(ctx, req.(*", inType, "))")
+		g.P("handler := func(ctx ", contextPkg, ".Context, req interface{}) (reply interface{}, err error) {")
```
paranoiacblack: Don't used a named return value here.

paranoiacblack:
This seems useful BTW, and I might take this over myself in another PR since the author hasn't gotten back to it.

paranoiacblack:
Oh, I misread this because of the change to generated code. My mistake. I'm sure this PR is an attempt to address the same issues described in grpc/grpc#12824 and #435. I looked at the original issue in rexray/gocsi#42 and it seems the code there is just printing the error and then attempting to log the response regardless. At the very least, it seems that code should assign this to nil, if there is an error. Please reopen if I miscategorized this.

jdef:
I noticed this when digging through gRPC code while researching some of the issues you cited, but this PR is NOT attempt to fix those issues. This PR is an attempt to fix an unintentional conversion from a nil pointer to a non-nil `interface{}`. If such a conversion is intentional then the generator should probably have a comment that indicates such (as well, perhaps, should the generated code). Otherwise it looks like a bug.

I'd be happy to re-open this but I don't have ops.

PR #459, opened 2017-11-22
JSONPb check enum int values

afking:
This checks enum integer values for JSONPb. Breaks one test that says this behavior is acceptable. However this caused a bug for me as I expected JSONPb to handled serialization to a know enum type.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -834,27 +834,6 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 		}
 	}
 
-	// Handle enums, which have an underlying type of int32,
-	// and may appear as strings.
-	// The case of an enum appearing as a number is handled
-	// at the bottom of this function.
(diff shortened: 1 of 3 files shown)


dsnet:
\cc @cybrcodr

cybrcodr:
If I understand what you are attempting to do in this PR is that you'd like for the Unmarshaler to return error if enum int value is not within the defined set of enums, is that correct?

Per https://developers.google.com/protocol-buffers/docs/proto3

> During deserialization, unrecognized enum values will be preserved in the message, though how this is represented when the message is deserialized is language-dependent. In languages that support open enum types with values outside the range of specified symbols, such as C++ and Go, the unknown enum value is simply stored as its underlying integer representation. In languages with closed enum types such as Java, a case in the enum is used to represent an unrecognized value, and the underlying integer can be accessed with special accessors. In either case, if the message is serialized the unrecognized value will still be serialized with the message.

I believe this is consistent with `proto.Unmarshal`.

PR #469, opened 2017-12-09
jsonpb: check for nil in Marshal and return error to avoid panic.

cybrcodr:
Add check for nil in Marshal and return error to avoid panic.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -118,7 +118,8 @@ type JSONPBUnmarshaler interface {
 
 // Marshal marshals a protocol buffer into JSON.
 func (m *Marshaler) Marshal(out io.Writer, pb proto.Message) error {
-	if pb == nil || reflect.ValueOf(pb).IsNil() {
+	v := reflect.ValueOf(pb)
+	if pb == nil || (v.Kind() == reflect.Ptr && v.IsNil()) {
 		return errors.New("Marshal called with nil")
 	}
 	writer := &errWriter{writer: out}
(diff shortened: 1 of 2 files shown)


jsonpb/jsonpb_test.go
```diff
@@ -462,6 +462,14 @@ func TestMarshaling(t *testing.T) {
 	}
 }
 
+func TestMarshalingNil(t *testing.T) {
+	var msg *pb.Simple
+	m := &Marshaler{}
+	if _, err := m.MarshalToString(msg); err == nil {
+		t.Errorf("mashalling nil returned no error")
```
dsnet: Nit: It seems the convention is usually one l.

s/marshalling/marshaling/

jsonpb/jsonpb.go
```diff
@@ -118,6 +118,9 @@ type JSONPBUnmarshaler interface {
 
 // Marshal marshals a protocol buffer into JSON.
 func (m *Marshaler) Marshal(out io.Writer, pb proto.Message) error {
+	if pb == nil || reflect.ValueOf(pb).IsNil() {
```
dsnet: This assumes that pb is a pointer, which is not always true.

cybrcodr: Ah yes.  Switched to checking for pointer only.  I'm guessing I shouldn't worry about nil maps, slices, etc.

jsonpb/jsonpb_test.go
cybrcodr: Done.

Issue #440, opened 2017-10-11
jsonpb Marshal and Unmarshal should return error on missing required field


mkevac:
If I have a message in *.proto file with required field and I try to convert json representation of this message to message itself, I would expect my Unmarshaller to check for required fields and to give an error if required fields are not in json representation. But jsonpb does not check that.

```
package main

import (
	gsp "lala/lala/proto"
	"log"
	"github.com/golang/protobuf/jsonpb"
)

func main() {
	jsonMsg := "{\"age_from\": 18}"
	msg := gsp.RequestSearch{}
	if err := jsonpb.UnmarshalString(jsonMsg, &msg); err != nil {
		log.Printf("error unmarshalling json: %s", err)
	}
	log.Println(msg)
}
```

```
type RequestSearch struct {
	UserId                *uint32     `protobuf:"varint,1,req,name=user_id" json:"user_id,omitempty"`
	AgeFrom               *uint32     `protobuf:"varint,2,opt,name=age_from" json:"age_from,omitempty"`
	AgeTo                 *uint32     `protobuf:"varint,3,opt,name=age_to" json:"age_to,omitempty"`
	Gender                *UserGender `protobuf:"varint,4,opt,name=gender,enum=badoo.gophersearch.UserGender" json:"gender,omitempty"`
	Lon                   *float32    `protobuf:"fixed32,5,opt,name=lon" json:"lon,omitempty"`
	Lat                   *float32    `protobuf:"fixed32,6,opt,name=lat" json:"lat,omitempty"`
	ExcludeList           []uint32    `protobuf:"fixed32,7,rep,name=exclude_list" json:"exclude_list,omitempty"`
	MainVectorNeeded      *uint32     `protobuf:"varint,8,opt,name=main_vector_needed" json:"main_vector_needed,omitempty"`
	BeautifulVectorNeeded *uint32     `protobuf:"varint,9,opt,name=beautiful_vector_needed" json:"beautiful_vector_needed,omitempty"`
	SearchDistance        *uint32     `protobuf:"varint,10,opt,name=search_distance" json:"search_distance,omitempty"`
	XXX_unrecognized      []byte      `json:"-"`
}
```

dsnet:
\cc @cybrcodr

cybrcodr:
JSON serialization is only described in proto3 page --
https://developers.google.com/protocol-buffers/docs/proto3#json

[...]

PR #472, opened 2017-12-13
jsonpb: change Marshal/Unmarshal to return error if any required field is not set

cybrcodr:
Change Marshal/Unmarshal to return error if any required field is not set.

For Unmarshal, this means JSON is either missing any required field or has required field set to null.

This is an initial attempt at fixing #440.  While I've modified Marshal to produce an error for required fields that are not set, I simply reuse this in Unmarshal to check for required fields which is probably not as efficient.

Anyways, I won't have time to improve on this this week and am simply putting this PR out in case someone would like to try it out and/or feel free to pick it up from here.

I've also yet to test the impact of this change on our existing codebase.


jsonpb/jsonpb.go
```diff
@@ -644,7 +649,19 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
 // permutations of the related Marshaler.
 func (u *Unmarshaler) Unmarshal(r io.Reader, pb proto.Message) error {
 	dec := json.NewDecoder(r)
-	return u.UnmarshalNext(dec, pb)
+	if err := u.UnmarshalNext(dec, pb); err != nil {
+		return err
+	}
+
+	// Marshal it back to check for missing required field error.
+	// TODO: Parse through the message and check for any missing required field w/o having to
```
puellanivis: This TODO will need to be resolved before this could be considered for merging. Marshaling the JSON just after Unmarshaling is going to build in a lot of overhead that wouldn’t be a good idea to pass around. (Especially for everyone using proto3, where required fields have been entirely removed.)

cybrcodr: Agree that this needs to be improved.  Sorry if I wasn't clear in my initial posting.  I didn't intend to merge this yet.  Since I didn't have time this week to continue doing this, I decided to post this PR for others to try it out and see if anyone may be interested in improving it further.

Note though that a proto3 message can contain a proto2 message and vice-versa.  So, as far as I know, the code will still need to traverse all the fields and subfields of the message looking for required ones that are not set.

puellanivis:
Hm… you are correct, [the best kind of correct](https://morbotron.com/caption/S02E15/1266340).

Though, I think anyone using proto3 might have entirely forgotten that fields were even allowed to be required… or maybe just those of us who got the Google “required considered harmful” meme drilled forcefully into our head by SREs…

jsonpb/jsonpb.go
```diff
@@ -257,12 +257,17 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
...
+		prop := jsonProperties(valueField, m.OrigName)
 		// IsNil will panic on most value kinds.
 		switch value.Kind() {
 		case reflect.Chan, reflect.Func, reflect.Interface:
 			if value.IsNil() {
 				continue
 			}
+		case reflect.Ptr:
+			if prop.Required && value.IsNil() {
+				return fmt.Errorf("required field %q is not set", prop.Name)
```
cybrcodr: Curious, for error messages like this, should I use Name or OrigName?  Or is it dependent on whether message is for marshaling or unmarshaling?

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+	}
+
+	return checkRequiredFields(pb)
+}
+
+// checkRequiredFields returns an error if any required field in the given proto message is not set.
+// This function is called from Unmarshal and assumes certain actions are done ahead.
+func checkRequiredFields(pb proto.Message) error {
+	// Most well-known type messages do not contain any required field.  The "Any" type may have
+	// required fields, however the Unmarshaler should have invoked proto.Marshal on the value field
```
puellanivis: Do you mean proto.Unmarshal here?

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func checkRequiredFieldsInValue(v reflect.Value) error {
+	if pm, ok := v.Interface().(proto.Message); ok {
+		if err := checkRequiredFields(pm); err != nil {
```
puellanivis: If this isn’t in a loop, then it’s not necessary to check the error value before returning it. i.e. you can just use `return checkRequiredFields(pm)`

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+			field = v.Field(0)
+			sfield = v.Type().Field(0)
+		}
+
+		var prop proto.Properties
+		prop.Init(sfield.Type, sfield.Name, sfield.Tag.Get("protobuf"), &sfield)
+
+		// IsNil will panic on most value kinds.
+		switch field.Kind() {
+		case reflect.Map:
```
puellanivis: I think we discussed this before, and you told me how a proto3 could contain a proto2 message, but maps are exclusive to proto3, so do they really need to be checked?

Consider adding some code comments explaining the rational of checking maps despite their appearance being expected to be mutually exclusive with required fields.

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
 // permutations of the related Marshaler.
 func (u *Unmarshaler) Unmarshal(r io.Reader, pb proto.Message) error {
 	dec := json.NewDecoder(r)
-	return u.UnmarshalNext(dec, pb)
+	if err := u.UnmarshalNext(dec, pb); err != nil {
+		return err
+	}
+
+	return checkRequiredFields(pb)
```
puellanivis: Would it be possible to work the code into UnmarshalNext, rather than iterate over all the fields a second time? (albeit, yes, fairly fast code.)

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+		field := v.Field(i)
+		sfield := v.Type().Field(i)
+		if strings.HasPrefix(sfield.Name, "XXX_") {
+			continue
+		}
+
+		// Oneof fields need special handling.
+		if sfield.Tag.Get("protobuf_oneof") != "" {
+			// field is an interface containing &T{real_value}.
+			v := field.Elem().Elem() // interface -> *T -> T
```
puellanivis: Have you considered the case where a oneof contains a simple scalar (string, bytes, varint), which in proto3 is not a pointer?

What will this code do in that case?

cybrcodr: Current proto compiler will always produce an interface field for a oneof field.  It will also produce wrapper structs, one for each oneof field.  And hence, this block will grab the oneof field inside the wrapper struct which can be any type.

jsonpb/jsonpb.go
cybrcodr: No, maps are allowed in proto2.  I have a test case for this.

jsonpb/jsonpb.go
cybrcodr: Possibly.  The current unmarshaling logic does rely on json.Unmarshal to populate fields.  To embed the check for required in there means that it will still need to check for all fields that have been unmarshaled.

But you are correct that doing the check w/in the unmarshaling code may be more efficient.  The unmarshaling code is quite complicated though that I prefer to avoid mixing this logic in right now.  A refactor/rewrite of the unmarshaling code should account for this logic.

jsonpb/jsonpb.go
cybrcodr: Nope.  During unmarshaling of an "Any" JSON to a Go proto message, after it resolves and constructs the embedded message, it will actual do a proto marshal to a serialized form and store the bytes into Any.Value field.  See https://github.com/golang/protobuf/blob/dev/jsonpb/jsonpb.go#L751.

I've updated the comment to explain this better.

jsonpb/jsonpb.go
cybrcodr: Good catch.  I previously have this block inlined above multiple times and then pulled it out to this helper func but forgot that this added check is no longer required.  Fixed.

jsonpb/jsonpb.go
dsnet: Even though `UnmarshalNext` is used by almost nobody, why is this check in `Unmarshal` as opposed to `UnmarshalNext`?

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+	// Most well-known type messages do not contain required fields.  The "Any" type may contain
+	// a message that has required fields.  When an Any message is being unmarshaled, the code will
+	// have invoked proto.Marshal on the embedded message to store the serialized message in
+	// Any.Value field, and that should have returned an error if a required field is not set.
+	// Hence skipping well-known types here.
+	if _, ok := pb.(wkt); ok {
+		return nil
+	}
+
+	v := reflect.ValueOf(pb).Elem()
```
dsnet: You will need to verify that pb is a pointer to a struct. If it isn't, then return.

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+		field := v.Field(i)
+		sfield := v.Type().Field(i)
+		if strings.HasPrefix(sfield.Name, "XXX_") {
+			continue
+		}
+
+		// A oneof field is an interface implemented by wrapper structs containing the actual oneof
+		// fields.  Field is an interface containing &T{real_value}.
+		if sfield.Tag.Get("protobuf_oneof") != "" {
+			v := field.Elem().Elem() // interface -> *T -> T
```
dsnet: Need to check:
1. for nil on both the interface.
2. that the interface element is a pointer to a struct
3. that the pointer is not nil

jsonpb/jsonpb.go
```diff
@@ -644,7 +649,105 @@ func (u *Unmarshaler) UnmarshalNext(dec *json.Decoder, pb proto.Message) error {
...
+			}
+			// Check each map value.
+			keys := field.MapKeys()
+			for _, k := range keys {
+				v := field.MapIndex(k)
+				if err := checkRequiredFieldsInValue(v); err != nil {
+					return err
+				}
+			}
+		case reflect.Slice:
```
dsnet: The "bytes" scalar for proto will fall under this case and will need to be specially handled.
IIRC, `[]byte{}` means that a zero-length bytes is set, while `[]byte(nil)` means that it was never set.

jsonpb/jsonpb.go
cybrcodr: Oops, thanks for catching.  Fixed.

jsonpb/jsonpb.go
cybrcodr: Done.  I'm guessing we'll simply ignore message if not pointer to struct.

jsonpb/jsonpb.go
cybrcodr: Good catch, I forgot about that.  Added handling of non-repeated slice.

Yes, I'm going with []byte{} to mean it is set.

jsonpb/jsonpb.go
cybrcodr: Done.

cybrcodr:
Thanks for the review, @dsnet.

@puellanivis do you have more feedback/comments on this PR before I merged it in?  Thanks.

jsonpb/jsonpb.go
```diff
@@ -1080,3 +1087,130 @@ func (s mapKeys) Less(i, j int) bool {
...
+		if sfield.Tag.Get("protobuf_oneof") != "" {
+			if field.Kind() != reflect.Interface {
+				continue
+			}
+			v := field.Elem()
+			if v.Kind() != reflect.Ptr || v.IsNil() {
+				continue
+			}
+			v = v.Elem()
+			if v.Kind() != reflect.Struct {
```
puellanivis: `&& v.NumFields() > 0` maybe, just to be sure?

cybrcodr: I think you meant to check for NumField() == 0 so that Field(0) below won't panic.  That would be a bug in protoc-gen-go, but I went ahead and added the check.

puellanivis: Right, but any value of NumFields() less than 0 would be an error state. And checking for greater than zero, and “is not zero” is either: identical with unsigned ints, or just extra CYA for signed ints.

Sure NumField _shouldn’t_ ever return a negative number, but iI find it to generally be a good idea to remember that _technically_ it might… even if only erroneously. So, I typically prefer to avoid weak/narrow tests, when a stronger/broader test is available for the same cost. (I also usually test if `len(array) < 1` rather than just `len(array) == 0` for the same reasons.)

It could be viewed as an example of “overengineering” but I prefer to cover all load-scenarios rather than only the specific load-scenarios that one thinks might cause a problem.

I mean, if you can order 100 nuts for $100 in either 3 lbs load-hanlding, or 1 lb load-handling, even if you expect that you will only need at most 1 lb of load-handling, it doesn’t cost anything extra to get the higher load handling ones.

cybrcodr: Done.

jhump:
FYI, this change effectively breaks the support for dynamic/custom messages that was added in #325. The way in which required fields are checked herein assumes that the message is normal generated message, without providing a way to override/customize.

In particular, if a given struct has any unexported field that contains a pointer value (including a slice or map with an element that contains a pointer value) and that pointer value is not nil, then this will cause a panic [here](https://github.com/golang/protobuf/pull/472/files#diff-ef290093d80b46fd7f16c53d7b68b1e4R1212):

```
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method 
```

I'll open an issue for it.

PR #473, opened 2017-12-13
Fix Import

florianl:
Signed-off-by: Lehner Florian <dev@der-flo.net>

As of Go 1.7 this package is available in the standard library under the name context.


dsnet:
Thanks for the PR. Unfortunately, we cannot do this yet because the `Context` in the stdlib is not the same as the one in the x/net repo unless type aliases are used. There are still enough users on pre-go1.9 and don't have access to type aliases.

I'll make the transition myself when we can bump the minimum version of protobufs (and clean up a bunch of old Go code).

PR #477, opened 2017-12-22
make json tag the same name as protobuf json_name

Gnouc:
Currently, the generator uses field name as json name, so set the `json_name` of field make none sense.

```
syntax = "proto3";

message Foo {
    string id = 1 [json_name = "_id"];
}
```

will generate go struct:

```
type Foo struct {
	Id string `protobuf:"bytes,1,opt,name=id,json=_id" json:"id,omitempty"`
}
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1781,7 +1781,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		ns := allocNames(base, "Get"+base)
 		fieldName, fieldGetterName := ns[0], ns[1]
 		typename, wiretype := g.GoType(message, field)
-		jsonName := *field.Name
+		jsonName := field.GetJsonName()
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for the PR. 

While protoc-gen-go does emit `json` tags for the standard library JSON encoder, this is problematic because the standard library cannot possibly emit JSON that follows the PB<->JSON specification. Instead, users are to use the `jsonpb` package. The fact that `json` tags were emitted in the past was a mistake and it gives the mistaken perception that the two are inter-operable. 

I would like to move to a world where `json` tags are not emitted at all. As such, I don't think making them more similar is a good thing, since it again encourages the false perception that `jsonpb` and `json` are identical.

Gnouc:
@dsnet thanks for infomation about jsonpb.

I'm using grpc-gateway to serve my grpc service over REST, is there easy way to integrate jsonpb to it?

dsnet:
I'm not sure. That's a question for the maintainers of grpc-gateway.

PR #482, opened 2018-01-11
Remove raw interface

### LINKED ISSUE #311: Remove references to RawMessage
The code contains a few references to this type, including `type raw interface { Bytes() []byte }` which is supposed to be implemented by `RawMessage`. Unfortunately, that method signature is not as uncommon as one would hope, which can cause bad interactions when that function is implemented on generated structs or their constituent fields (e.g. via github.com/gogo/protobuf's `customtype` option).

My question is: is RawMessage still a thing? If yes, what is it? If no, would you accept a PR removing "support" for it?

cc @dsymonds

dsnet:
There is an internal type to Google called RawMessage that is similar
to json.RawMessage. Since there is no proper proto reflection API,
we special-cased the Bytes method of RawMessage to access the raw bytes.
This is a gross hack since Bytes() []byte is such a common method signature.

Remove this hack.

Fixes #311

### DIFF
--- a/proto/equal.go
+++ b/proto/equal.go
@@ -109,15 +109,6 @@ func equalStruct(v1, v2 reflect.Value) bool {
 				// set/unset mismatch
 				return false
 			}
-			b1, ok := f1.Interface().(raw)
-			if ok {
-				b2 := f2.Interface().(raw)
-				// RawMessage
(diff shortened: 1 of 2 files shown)


dsnet:
\cc @tamird

PR #483, opened 2018-01-11
expose accessor for value of unregistered extensions [dev]

### LINKED ISSUE #385: no way to access data for unrecognized extensions

jhump:
This is the same as #420, but for the `dev` branch. This would resolve #385.

### DIFF
--- a/proto/extensions.go
+++ b/proto/extensions.go
@@ -291,13 +291,15 @@ func ClearExtension(pb Message, extension *ExtensionDesc) {
 	delete(extmap, extension.Field)
 }
 
-// GetExtension parses and returns the given extension of pb.
-// If the extension is not present and has no default value it returns ErrMissingExtension.
-// If an incomplete descriptor is used, the returned value will be the raw underlying
-// bytes since the extension value cannot not be decoded. An incomplete extension is one
-// returned from the ExtensionDescs function whose ExtensionType field is nil. An incomplete
-// extension has no default value, so querying for an incomplete extension that is not
-// present will cause this function to return ErrMissingExtension.
+// GetExtension retrieves a proto2 extended field from pb.
+//
+// If the descriptor is type complete (i.e., ExtensionDesc.ExtensionType is non-nil),
+// then GetExtension parses the encoded field and returns a Go value of the specified type.
+// If the field is not present, then the default value is returned (if one is specified),
+// otherwise ErrMissingExtension is reported.
+//
+// If the descriptor is not type complete (i.e., ExtensionDesc.ExtensionType is nil),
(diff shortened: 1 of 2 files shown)


dsnet:
Rather than adding a new function, can we change `GetExtension` to return the raw bytes as a `[]byte` if the `ExtensionDesc` passed in is "partial"; where we define a partial `ExtensionDesc` as one where `ExtensionDesc.ExtensionType` is nil.

jhump:
@dsnet, done. PTAL

proto/extensions.go
```diff
@@ -293,14 +293,22 @@ func ClearExtension(pb Message, extension *ExtensionDesc) {
 
 // GetExtension parses and returns the given extension of pb.
```
dsnet: // GetExtension retrieves a proto2 extended field from pb.
//
// If the descriptor is type complete (i.e., ExtensionDesc.ExtensionType is non-nil),
// then GetExtension parses the encoded field and returns a Go value of the specified type.
// If the field is not present, then the default value is returned (if one is specified),
// otherwise ErrMissingExtension is reported. 
//
// If the descriptor is not type complete (i.e., ExtensionDesc.ExtensionType is nil),
// then GetExtension returns the raw encoded bytes of the field extension.

jhump: Done

jhump:
@dsnet: I addressed your last comment. Is this ready to merge now?

PR #492, opened 2018-01-22
jsonpb: fix handling of illegal and negative nanoseconds

dfawley:
Fixes regression caused by #453.

@dsnet

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -56,6 +56,8 @@ import (
 	stpb "github.com/golang/protobuf/ptypes/struct"
 )
 
+const secondInNanos = int64(time.Second / time.Nanosecond)
+
 // Marshaler is a configurable object for converting between
 // protocol buffer objects and a JSON representation for them.
 type Marshaler struct {
@@ -201,6 +203,9 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
 			// "Generated output always contains 3, 6, or 9 fractional digits,
 			//  depending on required precision."
 			s, ns := s.Field(0).Int(), s.Field(1).Int()
+			if ns <= -secondInNanos || ns >= secondInNanos {
+				return fmt.Errorf("ns out of range (%v, %v)", -secondInNanos, secondInNanos)
+			}
 			if (s > 0 && ns < 0) || (s < 0 && ns > 0) {
 				return errors.New("signs of seconds and nanos do not match")
 			}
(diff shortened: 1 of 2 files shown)


dfawley:
Note that the regression is present on `master`, and the original breaking change #453 was not on `dev`.

dsnet:
Argh... I regret merging the last change into master since the two branches differ by that change now.

jsonpb/jsonpb.go
```diff
@@ -201,8 +201,13 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
 			// "Generated output always contains 3, 6, or 9 fractional digits,
 			//  depending on required precision."
 			s, ns := s.Field(0).Int(), s.Field(1).Int()
-			d := time.Duration(s)*time.Second + time.Duration(ns)*time.Nanosecond
-			x := fmt.Sprintf("%.9f", d.Seconds())
+			if (s > 0 && ns < 0) || (s < 0 && ns > 0) {
```
dsnet: This check seems wrong. The documentation for `Timestamp.nanos` says:
> Non-negative fractions of a second at nanosecond resolution. Negative second values with fractions must still have non-negative nanos values that count forward in time. Must be from 0 to 999,999,999 inclusive.

Which seems to indicate a negative value for `ns` is invalid regardless of `s` value.

dfawley: Sigh.  The spec here says the opposite:

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration

“For durations of one second or more, a non-zero value for the nanos field must be of the same sign as the seconds field. Must be from -999,999,999 to +999,999,999 inclusive.”

dfawley: Oh wait.  This is `Duration`, not `Timestamp`....

dfawley:
> Argh... I regret merging the last change into master since the two branches differ by that change now.

Shall we roll that one back then?

jsonpb/jsonpb.go
dsnet: Oh, my mistake. I'm looking at `Timestamp`.

Also, to what degree should we validate durations?
* Should we also verify that they are within -999999999 to +999999999 inclusive? Note that the logic below breaks if the value exceeds those bounds.
* If we're going to invalidate bad Durations here, should we be consistent and apply similar checks to Timestamps down below?

dfawley: > Also, to what degree should we validate durations?

I'll leave that up to you.  If you want this CL to be consistent with the rest of the code, I can remove the validation and testing of illegal inputs here.  If you want validation elsewhere, I would rather change the other places in a different commit, as this commit fixes a very real regression and it would be best to get a fix in ASAP.

dfawley:
ns range validation (and tests) added to Duration and Timestamp handling.

jsonpb/jsonpb_test.go
```diff
@@ -473,6 +476,33 @@ func TestMarshalingNil(t *testing.T) {
 	}
 }
 
+func TestMarshalIllegalTime(t *testing.T) {
+	tests := []struct {
+		pb      proto.Message
+		errWant error
```
dsnet: I don't think we care that much about the exact error message. Let's just change this to a "fail bool" that checks whether it passes or not.

dsnet:
Overall change LGTM. Just a nit with regards to the test.

jsonpb/jsonpb_test.go
dfawley: Done.

PR #496, opened 2018-01-25
proto: add logic to handle legacy message

dsnet:
Not all proto.Message implementations will be updated to be
using the most recent protoc-gen-go. Thus, they will lack an
XXX_DiscardUnknown method. Add logic to handle older protobufs.

### DIFF
--- a/proto/discard.go
+++ b/proto/discard.go
@@ -59,16 +59,10 @@ func DiscardUnknown(m Message) {
 		m.XXX_DiscardUnknown()
 		return
 	}
-	if m == nil {
-		return
-	}
-
-	// The Message interface really needs to provide some form of reflection
-	// API that can be used to implement the fallback if someone is using
-	// a custom protobuf message implementation.
-	//
-	// See https://github.com/golang/protobuf/issues/364
-	panic(fmt.Sprintf("cannot discard unknown fields on %T", m))
+	// TODO: Dynamically populate a InternalMessageInfo for legacy messages,
+	// but the master branch has no implementation for InternalMessageInfo,
+	// so it would be more work to replicate that approach.
+	discardLegacy(m)
 }
 
 // DiscardUnknown recursively discards all unknown fields.
@@ -172,14 +166,14 @@ func (di *discardInfo) computeDiscardInfo() {
 			tf = tf.Elem()
 		}
 		if isPointer && isSlice && tf.Kind() != reflect.Struct {
-			panic("both pointer and slice for basic type in " + tf.Name())
+			panic(fmt.Sprintf("%v.%s cannot be a slice of pointers to primitive types", t, f.Name))
 		}
 
 		switch tf.Kind() {
 		case reflect.Struct:
 			switch {
 			case !isPointer:
-				panic(fmt.Sprintf("message field %s without pointer", tf))
+				panic(fmt.Sprintf("%v.%s cannot be a direct struct value", t, f.Name))
 			case isSlice: // E.g., []*pb.T
 				di := getDiscardInfo(tf)
 				dfi.discard = func(src pointer) {
@@ -202,7 +196,7 @@ func (di *discardInfo) computeDiscardInfo() {
 		case reflect.Map:
 			switch {
 			case isPointer || isSlice:
-				panic("bad pointer or slice in map case in " + tf.Name())
+				panic(fmt.Sprintf("%v.%s cannot be a pointer to a map or a slice of map values", t, f.Name))
 			default: // E.g., map[K]V
 				if tf.Elem().Kind() == reflect.Ptr { // Proto struct (e.g., *T)
 					dfi.discard = func(src pointer) {
@@ -223,7 +217,7 @@ func (di *discardInfo) computeDiscardInfo() {
 			// Must be oneof field.
 			switch {
 			case isPointer || isSlice:
-				panic("bad pointer or slice in interface case in " + tf.Name())
+				panic(fmt.Sprintf("%v.%s cannot be a pointer to a interface or a slice of interface values", t, f.Name))
 			default: // E.g., interface{}
 				// TODO: Make this faster?
 				dfi.discard = func(src pointer) {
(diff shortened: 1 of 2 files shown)


dsnet:
@dfawley

dsnet:
The added logic is a perfect mirror of discardInfo.computeDiscardInfo.

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
 
 	atomic.StoreInt32(&di.initialized, 1)
 }
+
+func discardLegacy(m Message) {
+	v := reflect.ValueOf(m)
+	if v.Kind() != reflect.Ptr || v.IsNil() || v.Elem().Kind() != reflect.Struct {
```
cybrcodr: Optional.  Can move check v.Elem().Kind() != reflect.Struct after line 259.

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+		return
+	}
+	v = v.Elem() // Kind is Struct
+
+	for i := 0; i < v.NumField(); i++ {
+		f := v.Type().Field(i)
```
cybrcodr: Optional.  Can hoist v.Type() outside of loop.

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+		if strings.HasPrefix(f.Name, "XXX_") {
+			continue
+		}
+		vf := v.Field(i)
+		tf := f.Type
+
+		// Unwrap tf to get its most basic type.
+		var isPointer, isSlice bool
+		if tf.Kind() == reflect.Slice && tf.Elem().Kind() != reflect.Uint8 {
```
cybrcodr: Optional.  Capture tf.Kind() into a variable and use variable here and below.

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+			isSlice = true
+			tf = tf.Elem()
+		}
+		if tf.Kind() == reflect.Ptr {
+			isPointer = true
+			tf = tf.Elem()
+		}
+		if isPointer && isSlice && tf.Kind() != reflect.Struct {
+			panic("both pointer and slice for basic type in " + tf.Name())
```
cybrcodr: Not sure I understand the error message.  I think you meant "slice of pointers to non-struct type", right?

Also, should this be ignored/skipped instead of panic'ing?

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+		}
+
+		switch tf.Kind() {
+		case reflect.Struct:
+			switch {
+			case !isPointer:
+				panic(fmt.Sprintf("message field %s without pointer", f.Name))
```
cybrcodr: Same here, should this be ignored instead of panic'ing?

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+			}
+		case reflect.Map:
+			switch {
+			case isPointer || isSlice:
+				panic("bad pointer or slice in map case in " + f.Name)
+			default: // E.g., map[K]V
+				if vf.Type().Elem().Kind() == reflect.Ptr { // Proto struct (e.g., *T)
+					for _, key := range vf.MapKeys() {
+						val := vf.MapIndex(key)
+						discardLegacy(val.Interface().(Message))
```
cybrcodr: Is it possible for val to not be a Message here?  perhaps use the , ok check before calling discardLegacy and skip ones that are not?

proto/discard.go
```diff
@@ -256,3 +250,96 @@ func (di *discardInfo) computeDiscardInfo() {
...
+					}
+				}
+			}
+		case reflect.Interface:
+			// Must be oneof field.
```
cybrcodr: I take it that legacy messages do not use tags and may have oneof fields?  i.e. if there's protobuf_oneof tag, maybe better to rely on that?  Just a thought.

proto/discard.go
dsnet: I used your wording, which is more clear.
Let's keep the panic, since that's what the new logic (that this legacy fallback is based on) does.

proto/discard.go
dsnet: The value of tf is changing as the logic goes on, so this will actually end up be more confusing.

proto/discard.go
dsnet: Done.

proto/discard.go
dsnet: Keeping the panic, here and elsewhere.

proto/discard.go
dsnet: As far as I can tell, `golang/protoc-gen-go` nor `gogo/protoc-gen-go` ever allow this situation. If it's a pointer, then it's always a pointer to a sub-message.

I added logic to verify that the value type implements proto.Message.

proto/discard.go
dsnet: Done.

proto/discard.go
```diff
@@ -256,3 +250,101 @@ func (di *discardInfo) computeDiscardInfo() {
...
+			case isSlice: // E.g., []*pb.T
+				for j := 0; j < vf.Len(); j++ {
+					discardLegacy(vf.Index(j).Interface().(Message))
+				}
+			default: // E.g., *pb.T
+				discardLegacy(vf.Interface().(Message))
+			}
+		case reflect.Map:
+			switch {
+			case isPointer || isSlice:
```
cybrcodr: Totally optional as I just realized that most of these code is simply following the structure in computeDiscardInfo above so may make more sense to keep as is.

Seems easier to read just using if block on this instead of a switch with single case and a default.  Same for the reflect.Interface case below.

PR #504, opened 2018-02-01
Add functions for Timestamp manipulation

cjongseok:
Thanks for the great library!
I made some time.Time style Timestamp manipulation functions:
TimestampAdd(), TimestampSub(), TimestampSince(), TimestampUntil(), TimestampAfter(), TimestampBefore(), and TimestampEqual().
I hope it would be merged :)

### DIFF
--- a/ptypes/timestamp.go
+++ b/ptypes/timestamp.go
@@ -132,3 +132,68 @@ func TimestampString(ts *tspb.Timestamp) string {
 	}
 	return t.Format(time.RFC3339Nano)
 }
+
+// TimestampAdd returns the timestamp ts+d.
+// It returns an error if the resulting Timestamp is invalid.
+func TimestampAdd(ts *tspb.Timestamp, d time.Duration) (*tspb.Timestamp, error) {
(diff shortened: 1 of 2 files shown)


cjongseok:
I signed CLA. Please check it.

paranoiacblack:
Hi @cjongseok, we appreciate the effort you spent on this change. However, I don't think we'd like to add many more utilities to the current proto libraries unless they are absolutely necessary. It seems that once you have helpers to go from `timestamp_proto` to `time.Time` and vice-versa, implementing `Add`, `Sub`, `Until`, `After`, `Before` and `Equal` should be straightforward.

From a code review standpoint, your implementation of these functions and the associated tests are fairly reasonable so if you think this would be useful to many others, perhaps this would be better as a complementary library that exposes these helpers. If you'd like to discuss this decision more, please open up an issue and elaborate on the necessity of these helpers and we'll come up with some way to split what is needed from what is helpful to make the API boundary clear.

cjongseok:
As you said, helpers as a separate complementary library looks way better in the API boundary. Thanks for your good review & advice, @paranoiacblack :)

PR #506, opened 2018-02-02
protoc-gen-go: indicate deprecated fields in documentation

### LINKED ISSUE #396: Missing support for the deprecated option
**Syntax:**

```protobuf
int32 old_field = 6 [deprecated=true];
```

https://blog.golang.org/godoc-documenting-go-code

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation.

paranoiacblack:
Tools in the Go ecosystem treat code documented with "Deprecated:"
as special for purposes of warning users not to use them.  When a
field is marked as "[deprecated=true]", generator will produce a
comment indicating the deprecated status of the field.

Fixes #396.


paranoiacblack:
Not really familiar with how these are processed by the tooling. Adding a comment to the getters might be redundant if using the underlying deprecated field in any call path triggers `vet` and other analyses to fail.

paranoiacblack:
@tsl0922 sorry for the noise on the original issue. Thoughts on this?

protoc-gen-go/generator/generator.go
```diff
@@ -2252,6 +2258,12 @@ func (g *Generator) generateMessage(message *Descriptor) {
 			})
 		}
 
+		if field.GetOptions().GetDeprecated() {
+			// Generate deprecation comments for getters as well.
+			g.P("// Deprecated: ", mname, " should not be used by new code.")
+			g.P("// Underlying field is marked as [deprecated=true] in ", g.file.Name)
+			g.P("//")
```
tsl0922: Is the 2 lines above needed? The end users may not care about the proto definition here.

paranoiacblack: Is it needed? I don't know. I thought it would be useful to explain why a field is deprecated, but the only reason that it is deprecated in because of the proto definition. I don't have a strong opinion either way.

dsnet:
According to descriptor.proto, deprecation can be applied at multiple levels: file, message, field, enum, enum value, service, and method.

Where to put the comment is clear for most of these:
* message: just above the type declaration.
* field: inline with the field declaration (e.g., `MyField *string // Deprecated: ...`. The associated getters and setters will need to be tagged as well above the func declaration.
* enum: inline with the type declaration (e.g., `MyEnum int32 // Deprecated: ...`.
* enum value: inline with the const declaration (e.g., `FooValue MyEnum = 5 // Deprecated: ...`.
* service: just above the type declaration for the service.
* method: just above the func declaration for the method.

The difficult one is file. Since multiple proto files can go into a single Go package, you cant simply put the "Deprecated" tag in the package comment header.

Given the verbosity. I'm okay if all it said was "Deprecated: Do not use."

protoc-gen-go/generator/generator.go
paranoiacblack: Done.

paranoiacblack:
Please take another look. Newest commit handles all possible deprecated elements in a proto file as you described above. For services, this PR also emits deprecation comments on the `NewGeneratedServiceClient` and `RegisterGeneratedServiceServer` exported functions as well as the generated interface types.

protoc-gen-go/testdata/deprecated/deprecated.proto
```diff
@@ -0,0 +1,58 @@
+// Go support for Protocol Buffers - Google's data interchange format
+//
+// Copyright 2010 The Go Authors.  All rights reserved.
```
dsnet: Wheeeee... Time travel. Let's party like its 2010.

protoc-gen-go/testdata/deprecated/deprecated.proto
```diff
@@ -0,0 +1,58 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+
+// This package contains only deprecated messages and services.
+package deprecated;
+
+option deprecated = true; // file-level deprecation
+
+message DeprecatedRequest {
```
dsnet: Add comment above each message, field, enum, etc to test the ordering of whether "Deprecated: " appears above, below, or inline relative to the comment.

protoc-gen-go/generator/generator.go
```diff
@@ -1330,7 +1330,11 @@ func (g *Generator) generate(file *FileDescriptor) {
 // Generate the header, including package definition
 func (g *Generator) generateHeader() {
 	g.P("// Code generated by protoc-gen-go. DO NOT EDIT.")
-	g.P("// source: ", g.file.Name)
+	if g.file.GetOptions().GetDeprecated() {
+		g.P("// ", g.file.Name, " is deprecated: Do not use.")
```
dsnet: " is a deprecated file."

protoc-gen-go/generator/generator.go
```diff
@@ -1488,17 +1492,27 @@ func (g *Generator) generateEnum(enum *EnumDescriptor) {
 	ccTypeName := CamelCaseSlice(typeName)
 	ccPrefix := enum.prefix()
 
+	deprecatedEnum := ""
+	if enum.GetOptions().GetDeprecated() {
+		// Generate inline enum deprecation comment.
+		deprecatedEnum = "// Deprecated: Do not use."
```
dsnet: Make this an unexported constant?

protoc-gen-go/testdata/deprecated/deprecated.proto
dsnet: For the record, the standard library places the deprecation warning under the comment:
https://golang.org/pkg/compress/flate/#WriteError

paranoiacblack: Done, and added test to guarantee superfluous comments are emitted when there aren't any original ones.

protoc-gen-go/testdata/deprecated/deprecated.proto
paranoiacblack: =P Fixed.

protoc-gen-go/generator/generator.go
paranoiacblack: Done.

protoc-gen-go/generator/generator.go
paranoiacblack: Done

PR #511, opened 2018-02-07
proto: treat bad wire types as unknown fields

dsnet:
Previously, an error was returned during unmarshal when a wiretype
was encountered that did not match the expected wiretype.

In order to match the behavior of the C++ and Python implementations,
we no longer return an error and instead store the bad wire fragment
as an unknown field (or skip them if unknown field preservation is disabled).

The generator still produces code that references ErrInternalBadWireType
for the unmarshal logic for oneof fields.
However, the current proto package does not use the generated unmarshalers
for oneofs, so their existence has no bearing on unmarshal semantics.

Cleaning up the generator to stop producing these is future work.

### DIFF
--- a/proto/all_test.go
+++ b/proto/all_test.go
@@ -1131,21 +1131,25 @@ func TestBigRepeated(t *testing.T) {
 	}
 }
 
-// Verify we give a useful message when decoding to the wrong structure type.
-func TestTypeMismatch(t *testing.T) {
-	pb1 := initGoTest(true)
+func TestBadWireTypeUnknown(t *testing.T) {
+	var b []byte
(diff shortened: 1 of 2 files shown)


rickypai:
@dsnet just curious. Do you know why the original protobuf implementation treats incompatible wiretypes as unknowns? is it to help preserve compatibility between messages that might change wiretypes?

dsnet:
This is the behavior of the C++ implementation, which has become the de-facto "standard" of "correct" behavior. The rationale for this property is not documented to my knowledge.

rickypai:
thank you for the prompt response!

PR #516, opened 2018-02-19
grpc: expose ServiceDesc and ServiceDescriptor

mightyguava:
https://github.com/golang/protobuf/issues/489


mightyguava:
I signed it!

protoc-gen-go/testdata/servicedescriptor_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if desc.ServiceName != "testdata.Test" {
+		t.Error("Service name does not match")
+	}
+	if len(desc.Methods) != 1 {
+		t.Error("Expected 1 method")
+	}
+	if len(desc.Streams) != 3 {
+		t.Error("Expected 3 streams")
+	}
+	if expectedDescString != ListMethodsAndStreams(desc) {
```
puellanivis: Marshaling the descriptor into a string and then testing is not testing the actual public API or contract.

One should likely check that the ServiceName, MethodName(s), and []interface{}{ StreamName, ClientStream, ServerStream }... are all correct directly (since the func values preclude reflect.DeepEqual from working). Oh, and checking the Handlers to be non-nil would also be a good idea.

As a trivial (and dumb) bad-data-passes-test example, what if the Descriptor returned with two StreamNames, "Downstream\n\ttestdata.Test/Upstream" and "Bidi"?

“That would never happen!” But this is what testing is here for. We want to test the contract, (i.e. that the proper number and value of Descriptors is returned) not that marshaling it together in this specific way produces the same string. (n.b.: this example is ridiculous because it is a trivial set of bad data that produces a passing test… there could be further non-trivial bad data that would still pass, but are harder to think of.)

protoc-gen-go/testdata/grpc.pb.go
```diff
@@ -0,0 +1,370 @@
...
+		return nil, err
+	}
+	return m, nil
+}
+
+func GetTestServiceDescriptor() ([]byte, []int) { return fileDescriptor0, []int{0} }
+
+var _Test_serviceDesc = GetTestServiceDesc()
+
+func GetTestServiceDesc() grpc.ServiceDesc {
```
puellanivis: grpc.ServiceDesc is intended to be passed around by reference. Current design uses a non-pointer _ServiceName_serviceDesc because of a complicated argument about about behind the scenes allocations.

However, here, we are definitely passing around each individual element directly on the stack every time it is called or used in a function. That’s a lot of copying data around, especially for a particularly large Service.

protoc-gen-go/testdata/servicedescriptor_test.go
paranoiacblack: I agree. Furthermore, this test would be more appropriate if it was table-driven. Provide expected service descriptors and check to see if the one returned from Get*ServiceDesc are equivalent. This will avoid this test getting out of date as Get*ServiceDesc hides or changes internals.

Some other tests that could be useful here:
- Checking that an imported proto package returns valid ServiceDescriptor objects
- Checking that modifying a given ServiceDescriptor before registering it affects the related service's provided API. I.E., can you call the renamed methods if you get a service descriptor and modify the names/positions/add a method?

puellanivis: "table-driven" is overrated.  In a table-driven test you get all of the pass tests and all of the fail tests in the same *_test.go files and line reference.

This is solely and only useful when a piece of code has such an unreasonably wide scope that it has to be tested by algorithmic production. i.e. metaprogramming. Sometimes this is unavoidable, but when a specific file is supposed to produce a singular canonical output, that output should be known ahead of time, and should be human generatable, or the results themselves become themselves subject to testing, lest we become like Windows' "the implementation defines the behavior".

protoc-gen-go/testdata/grpc.pb.go
mightyguava: Is this a comment on the current implementation or this PR?

protoc-gen-go/testdata/servicedescriptor_test.go
mightyguava: I agree table-driven tests are overrated. I also use table-drive tests a lot.

I'm not really testing anything here because @paranoiacblack asked me to provide a test case as an example of usage. This is an example of usage. I'd like to be able to generate a list of methods that a protobuf service provides as a string, to display to a command-line user.

If you are looking unit tests for regressions, for documentation, for corner cases, etc, I'm happy to spend the time to do that, _after_ the discussion https://github.com/golang/protobuf/issues/489 finishes and I can be sure I'm not wasting my time here.

protoc-gen-go/testdata/grpc.pb.go
```diff
@@ -0,0 +1,370 @@
...
+	StreamMsg2
+*/
+package testdata
+
+import proto "github.com/golang/protobuf/proto"
+import fmt "fmt"
+import math "math"
+
+import (
+	context "golang.org/x/net/context"
```
awalterschulze: Committing this will cause golang/protobuf to have extra dependencies.
I created a [grpctests](https://github.com/gogo/grpctests) repo to keep these things separate and to avoid having any dependencies.

protoc-gen-go/testdata/servicedescriptor_test.go
paranoiacblack: Let's move forward with some sort of unit testing, even if it isn't very good. I don't get the segue into table-drive tests and it doesn't matter to me. Write a good test that isn't table-driven or write a good test that is table-driven. It's not very interesting to bikeshed this little thing but if you write a unit test that repeats a bunch of logic, I'll probably make you rewrite it.

protoc-gen-go/testdata/grpc.pb.go
paranoiacblack: Make this return a pointer. I don't know if it particularly matters that this copies data. It's not good, for sure, but ServiceDescs are always referenced for the intents of using the grpc package.

protoc-gen-go/testdata/grpc.pb.go
mightyguava: This is in a `testdata` folder. It shouldn't matter right?

protoc-gen-go/testdata/grpc.pb.go
mightyguava: Generated grpc code that use the ServiceDesc still use the var up above. It does not get copied. Only callers outside of grpc would get a copy. If this doesn't matter too much, I'd rather not change this, and avoid the awkwardness of doing something like

```
var _Test_serviceDescCopy = GetTestServiceDesc()
var _Test_serviceDesc = &_Test_serviceDescCopy
```

mightyguava: @paranoiacblack after reading your comment in the other file again, did you actually want the code to look like this?

```
var _Test_serviceDesc = &grpc.ServiceDesc{
  ...
}

func GetTestServiceDesc() *grpc.ServiceDesc {
	return _Test_serviceDesc
}
```

This would make usages able to mutate the `ServiceDesc` and affect how the service is registered. Not sure that would be desired behavior...

protoc-gen-go/testdata/grpc.pb.go
mightyguava: Also, did you want me to move the tests into your grpctest package? Or for the repo owners here to create a test repo to mirror to gogo one?

mightyguava:
Related to @awalterschulze's comment, tests are failing because of a missing `google.golang.org/grpc` package. Should I add the dependency here or should this test get moved elsewhere?

protoc-gen-go/testdata/grpc.pb.go
paranoiacblack: What I wanted, explicitly was for the original definition of `_Test_serviceDesc` to remain untouched:

```
var _Test_serviceDesc = grpc.ServiceDesc
```

And for `GetTestServiceDesc` to do the following:

```
func GetTestServiceDesc() *grpc.ServiceDesc {
        return &_Test_serviceDesc
}
```

You are correct that this will allow service descriptors to be modified by the server/clients that wish to register and use them. However, I think that might have been some of the intention behind this change as well. Your use-case involves only introspecting these objects so you can make observations about which methods exists, but another use-case might be adding ad-hoc methods or removing them ad-hoc from a given service before registration (like some sort of middleware interception).

mightyguava: That would be asking the user to change global state in order to change how a server is registered. If that's something you want to support, it actually sounds like the support should be built into service registration in `grpc`. cc: @dfawley

protoc-gen-go/testdata/grpc.pb.go
paranoiacblack: I'm not sure if I understand the intention behind separating the grpc dependency in tests to a different repository. Can you elaborate @awalterschulze? It seems like the transitive dependencies of gRPC aren't large in and of themselves, either.

paranoiacblack:
You could probably move these tests under the gRPC directory is `go get ./...` isn't working.

protoc-gen-go/testdata/grpc.pb.go
paranoiacblack: I think that the way the `grpc` package will support this is by using `GetServiceDesc` and modifying it, probably. Either way, Service Descriptors are meant to be passed around by reference.

protoc-gen-go/testdata/grpc.pb.go
awalterschulze: Currently golang/protobuf has one dependency in a test, which I specifically remove in gogo/protobuf.
It is nicer to depend on something that does not depend on anything except the standard library.
Any dependency is an extra dependency that needs to be managed.
Also even though this won't cause a circular dependency between packages, it does cause a circular dependency between repositories.
Its not the end of the world, but if we can avoid it, it does make things simpler.

protoc-gen-go/testdata/grpc.pb.go
puellanivis: I would much prefer keeping the _Test_serviceDesc definition alone, but make a copy right before returning the pointer-to.

```
var _Test_serviceDesc = grpc.ServiceDesc{
  ...
}

func GetTestServiceDesc() *grpc.ServiceDesc {
	desc := _Test_serviceDesc
	return &desc
}
```

Yes, this makes a copy, but a) it doesn’t make a copy everytime, b) it’s unavoidable if one wishes to avoid mutability of state, c) should be the uncommon case, and so optimizing it for performance would be the wrong way to do optimization anyways.

However, considering the structure could be quite large, let’s say 64 bytes isn’t unreasonable, then just _using_ it incurs this copy cost every time. So, we copy once as it is leaving our implementation, and then usage continues with pointers.

dfawley: @puellanivis, This doesn't make a deep copy, so if the user changed the slices' contents, that would impact the global state.

If we are doing this (which I'm still not totally convinced is worth doing yet), I think the way it's written is fine, except change the return type of this function to `*grpc.ServiceDesc`.  Then it will be passed around more naturally, but everyone that asks for it gets their own version.  This really should not be mutable.

puellanivis: Ugh… oh yeah.

I had looked at an earlier version that was returning a complex literal, and so the the only change would have been adding a "&" instead… I think things have shifted since then.

I’ll just walk away again… :/ (not enough coffee)

protoc-gen-go/testdata/grpc.pb.go
mightyguava: @paranoiacblack does your 👍 mean we should proceed to add this dependency, or that this test belongs in a different repo? I'm waiting to proceed on this.

mightyguava: Oh actually, the 👍 comes from @puellanivis, so I guess there's some disagreement here on your side?

mightyguava:
https://github.com/golang/protobuf/pull/516/commits/9972431b5c5f6998a6c64d970634fab4436edc5b changes `GetServiceDesc()` returns a pointer.

mightyguava:
8f794fd fixes the test by explicitly getting the `google.golang.org/grpc` dependency. I think this is fine for now.

awalterschulze:
My concerns have not been mitigated. To me this is still an extra dependency.

My reasoning is that there is no consensus on how go developers manage dependencies, yet.
One possible method, still used, is to pull in the whole git repo using git subtree or git submodule.
Running go test over your whole repo is also still not unpopular and this will then require an extra dependency to be pulled in as well.

It is not the end of the world, but it is inconvenient.

mightyguava:
@awalterschulze `go test ./...` will not run these tests or build these dependencies because they are in `testdata`. You'd have to do something pretty weird that explicitly references the `testdata` package by relative path to get it to run. I think the risk is pretty minimal.

That said though, I'm happy to move this test elsewhere, if the protobuf owners want to make a new repo for me to move it to.

awalterschulze:
I didn't know that go test skips folders named `testdata` folders, very interesting.

protoc-gen-go/grpc/grpc.go
```diff
@@ -256,6 +260,7 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	g.P("},")
 	g.P("Metadata: \"", file.GetName(), "\",")
```
dfawley: I have a concern about this field.  This is supposed to be extensible, for gRPC internal use.  Exporting this struct means users may begin to rely upon this behavior.  Is there a way to prevent that?

mightyguava:
@dfawley ok. No longer exposing servicedesc.

Rebased and squashed the rest of the changes.

protoc-gen-go/grpc/grpc.go
```diff
@@ -223,6 +223,7 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	}
 
 	// Service descriptor.
+	g.P("func Get", servName, "ServiceDescriptor() ([]byte, []int) { return ", file.VarName(), ", []int{", strconv.Itoa(index), "} }")
```
dfawley: Go style: is "Get" OK?

Also: people sometimes name their services something like "FooService"; This would result in `GetFooServiceServiceDescriptor`.  Messages have a `Descriptor` method.  Should this be `"func ", servName, "Descriptor()"`?

dfawley: Observation: it's technically possible for this to collide with another symbol in the proto file that ends up defined as a top-level type.  This is also true for the `New` functions we generate.  Is that OK?

Note that if we use `<MyService>Descriptor()`, we're probably more likely to collide than prepending the "Get".  Another option: `<MyService>_Descriptor()`.

mightyguava: I don't have an opinion here. Collision is possible but unlikely because I don't see people naming services FooServiceServiceDescriptor.

paranoiacblack: I think the `Get` prefix is fine. In the case you mentioned where a proto `service` has a redundant name, I think it's okay for the generator to include the redundancy.

paranoiacblack: I don't see collision as highly possible, either. It seems difficult to define a go proto package that defines two similarly named proto `service`s. And then I also think it would be difficult to make the go compiler happy. It's unfortunate and I think it would be a problem better resolved by the code author. I really don't expect there to be many situations where this collides.

dfawley: > I don't see people naming services FooServiceServiceDescriptor.

A message named `FooServiceDescriptor` doesn't seem too absurd.  But if we're going with `Get`, then it's much more unlikely.  And `NewFooService` also seems plausible.

@dsnet & @paranoiacblack, my question remains: can we not find a way for the main Go code generator to support more easily accessing ServiceDescriptorProtos?  It seems unusual to me for the grpc plugin to be the thing that provides this capability since it is a staadard proto feature.

paranoiacblack: I'm not sure what you mean by supporting more easy access to service descriptor protos; this change seems to be the easiest access I've heard suggested so far. The other approach we discussed was exporting file descriptors, which would allow users to get to the nested service descriptors. While it might be nice to do that, I don't think it would be easier for users.

I don't think it's odd for the gRPC plugin to generate this code, currently, since it is the defacto proto service implementation for RPCs. If there were another popular implementation, we could move this logic to the main protoc-gen-go to generate generic service descriptors. It seems okay for the gRPC plugin to do this because there's no particular reason to generate this unless grpc is invoked, anyways.

mightyguava: If popularity is measured by github stars, twirp is super popular: https://github.com/twitchtv/twirp. I wouldn't be surprised if it overtakes grpc by the end of the year.

dfawley: > this change seems to be the easiest access I've heard suggested so far

These same functions could just as easily be part of the main Go code generator.  Service descriptor protos are not grpc-specific; they're "services" specific.  I see this as a part of basic introspection of proto definitions.

If you consider this idea and still believe this belongs in only the grpc plugin, then let's just merge this and move forward.

paranoiacblack: I don't think this only belongs in the gRPC plugin. On the contrary, I'm only stating that it is not unusual, as you originally claimed. This PR is particularly about being able to do this with gRPC. If y'all really foresee this being required/useful for twirp or some other RPC implementation, then let's go ahead and export this from the protoc plugin, directly, I suppose.

mightyguava: I'd like to merge this PR if possible. @paranoiacblack I don't think there any remaining blockers.

dfawley: Created #579 for doing this in the main generator instead.

Note that this PR is into the master branch, which the proto team does not allow any longer.

dsnet:
See my comment in https://github.com/golang/protobuf/issues/364#issuecomment-384123374 for a more comprehensive plan to add this functionality.

Issue #141, opened 2016-03-03
Proposal: generator.WriteImport()


zellyn:
### Goal
per-file import uniquification

### Why?
1. It makes output deterministic. Currently, changing the order of input files, or even adding a new import to one file can result in changes to other files. 
2. It's a big step towards being able to generate files one at a time, or multiple packages at a time, as protoc does for other languages. Numerous people have requested this, the ability to generate file-at-a-time was only [recently broken](https://github.com/golang/protobuf/commit/4a63085a88#commitcomment-14418464), and it's otherwise easy to fix.
3. It cleans up the generated code when multiple plugin-plugins use the same packages.

I am happy to write this code if there's general support, and I don't believe the changes would be intrusive.

### Current operation

Currently, plugins call `RegisterUniquePackageName` once per imported name. This happens globally (ie. once per run).

### Proposed addition

Add a method `WriteImport(f *FileDescriptor, desiredName string, importPath string) (prefix string)`

This would generate an `import` statement in the output, as well as returning a prefix string to use when referring to objects from that package.

- If `importPath` is the same package as the FileDescriptor, no import is generated, and an empty prefix is returned.
- If `importPath` has already been imported in the file, no import is generated, and the previous prefix is returned.
- If `desiredName` has already been used by a different import in this file, or reserved by `RegisterUniquePackageName`, a unique import alias is generated for the import. The prefix returned is the alias followed by `.` The alias is added to the blacklist used by `RegisterUniquePackageName`.

This change would allow incremental movement from `RegisterUniquePackageName` towards `WriteImport`, and both could coexist.

@dsymonds @jhump @tamird @awalterschulze

PR #527, opened 2018-02-22
do not allow files that cross package boundaries

jhump:
protoc-gen-go tries to disallow multiple files if they cross package boundaries -- it wants to be invoked for a single package at a time, and with all files for that package at the same time.

But it doesn't do a very good job checking for this condition. It simply verifies that the package name for all files is the same, but it does not check that all files emit code to the same package path (which is what Go really considers to be the identity of a package).

That means I can supply different files that have the same package name but live in different folders, and the generated code is simply wrong. Because it assumes they are the same package (including same package path), references from one to the other will result in code that uses naked symbol references (no package qualifier/alias) and also not produce an import statement.

So this change will cause the plugin to issue an error message if the files it produces would be in different Go packages.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -296,23 +296,25 @@ func (d *FileDescriptor) goPackageOption() (impPath, pkg string, ok bool) {
 	return
 }
 
-// goPackageName returns the Go package name to use in the
-// generated Go file.  The result explicit reports whether the name
-// came from an option go_package statement.  If explicit is false,
-// the name was derived from the protocol buffer's package statement
+// goPackagePathAndName returns the Go import path and package name to use
+// in the generated Go file.  The result explicit reports whether the
+// package name came from an option go_package statement.  If explicit is
+// false, the name was derived from the protocol buffer's package statement
 // or the input file name.
-func (d *FileDescriptor) goPackageName() (name string, explicit bool) {
+func (d *FileDescriptor) goPackagePathAndName() (impPath, pkg string, explicit bool) {
 	// Does the file have a "go_package" option?
-	if _, pkg, ok := d.goPackageOption(); ok {
-		return pkg, true
+	if impPath, pkg, ok := d.goPackageOption(); ok {
+		return impPath, pkg, true
 	}
+
+	dir, base := splitName(d.GetName())
 
 	// Does the file have a package clause?
 	if pkg := d.GetPackage(); pkg != "" {
-		return pkg, false
+		return dir, pkg, false
 	}
-	// Use the file base name.
-	return baseName(d.GetName()), false
+	// Use the file name.
+	return dir, base, false
 }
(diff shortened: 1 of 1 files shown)


jhump:
Ping?

neild:
This is reasonable, but I'm in the middle of some significant reworking of this part of the code. (I'd really like to make it possible to just generate output for multiple packages at once, but no promises.) Mind holding off for a bit?

jhump:
@neild, I am happy to never merge this. I think `protoc-gen-go`'s behavior that requires it to compile only one package and a complete package (e.g. can't produce output for single proto file at a time, if the target package results from more than one source file) is actually terrible. It is very different from how other language output works, where protoc happily transforms any number of files, regardless of package.

This patch was just to make that behavior actually work, since we ran into a situation where files had different packages, but the same name last path element. Instead of spitting out an error, the plugin just generated non-working code.

When I worked at Square, we (and by we I mean @zellyn) tried to submit issues/patches to fix that IIRC, but they weren't accepted. It required us to maintain a fork of this repo since we had build scripts that would generate code for lots of proto files (crossing packages) all at once, and generate output for multiple languages, not just Go (only Go has this quirk).

zellyn:
We use https://github.com/square/goprotowrap. It doesn't let you generate one file at a time, but it does help you generate all the files you need at once without having to think about it.

Back when I was working on Go at Square (I got reassigned to Java-land 😢), it seemed that a few fairly simple changes to how go protobufs generate code would fairly easily clean up the file-at-a-time restriction. In particular, uniquifying symbols better within each file would probably do it. (While you're at it, you might want to make package import names sane too…)

However, getting even simple patches (or indeed even ideas for changes: I'm fine if the maintainers don't like my implementation) against the Go protobuf code base accepted proved completely futile, and we gave up.

awalterschulze:
@zellyn Please bring your ideas.  This is a new team maintaining golang/protobuf.  They listen to ideas and are really trying to fix things.

jhump:
@awalterschulze, I remember a couple of them. One of them is written up in #141.

The other was embedding the proto source file name into the `fileDescriptor*` names, instead of just a numeric suffix that always started at 0. That would allow compiling two source files for the same package in *separate invocations* of protoc and still getting deterministic output that compiles. (Currently, that produces code that won't compile because each output file will have a package-level var with the same name, like `fileDescriptor0`.)

awalterschulze:
Oh wow I remember that the fileDescriptor one.
I think gogoprotobuf ended up going for the filename instead of the number.
That was quite a while ago.

zellyn:
Heh. I forgot about #141. That is a pretty good writeup. Good work, past self! :-)

neild:
FYI, #549 will change protoc-gen-go to enforce that all generated files are under the same import path.

(I'm also making progress towards removing the need for that requirement entirely, but it's not there yet.)

PR #528, opened 2018-02-22
no more generated package doc

### LINKED ISSUE #413: protoc-gen-go: improve package doc
The files generated by protoc-gen-go include a package doc comment that is usually inappropriate. 

```go
/*
Package <pkg> is a generated protocol buffer package.

It is generated from these files:
	<some-file>.proto
	<some-other-file>.proto

It has these top-level messages:
	<SomeMessage>
	<SomeOtherMessage>
*/
```

While this information makes for a useful comment, it not very good godoc. It can't, for example, explain the purpose of the package nor can it provide an overview for any non-generated code in the package. It also makes no reference to GRPC services, which for some packages will actually be the purpose of the package.

Ideally, the tool would not output any package doc comments at all, and just allow developer to provide their own package doc in a separate file.

Also, because the tool emits the above comment as package doc, it only generates a single such comment and puts it (arbitrarily) in the first generated file. It would be actually be more useful if this were just a file comment that appeared in *every* file (not package doc) and showed the name of the one source file and top-level messages for that single file.

jhump:
This would resolve #413. It relies on a flag that can be passed to protoc-gen-go.


dsnet:
\cc @neild. We were discussing just removing this all together.

neild:
The package doc comment is pretty useless at the moment. I can see a case for just removing it. I don't really have a strong opinion one way or the other, although I think I'd rather get rid of it always than add another knob.

jhump:
@neild, how does it look now?

paranoiacblack:
+1 to getting rid of this

neild:
CL looks fine to me, I'll leave it up to @dsnet to decide whether we actually want to do this or not.

dsnet:
I vote to remove it as well.

As it currently stands, it:
* provides no useful information. The list of `.proto` files is not particularly useful for the end user and the list of top-level messages is redundant with the godoc index.
  * One could argue that the [index is overly-verbose](https://godoc.org/google.golang.org/genproto/googleapis/bigtable/v2#pkg-index) because it contains other identifiers like: sub-messages (these sort nicely as they have the parent as a prefix), enums, grpc stubs, and oneof wrappers (it's unfortunate that we generate so many of these). If we did end up wanting to clean this up, perhaps [use of per-file godoc sections](https://github.com/golang/go/issues/18342) is probably better to segregate the index so that all the oneof wrappers appeared at the end.
* actually conflicts with users who desire to put their own hand-written package documentation (which is almost always going to be more informative than what we can generate).

PR #529, opened 2018-02-22
proto: avoid pointer arithmetic with invalidField

dsnet:
In the case of a tag of zero, we perform some pointer arithmetic
where we add -1 to the pointer. Even though we do not end up dereferencing
the pointer, this still causes issues with the Go GC, which expects that
every Go pointer is a valid pointer.

This causes crashes on Go1.7 and below because the function arguments
are considered live throughout the entire function causing the GC to
scan this invalid pointer. On Go1.8 and above, this is not an issue
since the argument is not considered live and is not scanned.

To be on the safer side, perform pointer arithmetic that is effectively
a noop, so that we don't cause a valid pointer to become invalid.

### DIFF
--- a/proto/pointer_reflect.go
+++ b/proto/pointer_reflect.go
@@ -57,6 +57,9 @@ func toField(f *reflect.StructField) field {
 // invalidField is an invalid field identifier.
 var invalidField = field(nil)
 
+// zeroField is a noop when calling pointer.offset.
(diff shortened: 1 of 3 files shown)


dsnet:
@cherrymui, can you review this?

cherrymui:
Did I do it right? I have never done review on GitHub. Let me know if there is anything needed.

PR #533, opened 2018-02-26
protoc-gen-go: add paths=source_relative option

### LINKED ISSUE #515: protoc-gen-go: better support for controlling output filenames
# Background
There are lots of options controlling the package name and/out filename of generated files. Most of them are some combination of not useful or difficult to use correctly. Some of this is a documentation problem, some of this is not.

Related issues, probably not exhaustive: #39, #63, #64, #167, #181, #239, #240

When processing a `.proto` file, we need to know several things:

## The Go package name
Needed so we can write a `package foo` line in the generated code. Set by the `go_package` option if there is one, otherwise derived from the proto package name. This isn't usually a problem.

You can also pass `--go_out=import_path=x` to specify a package name. This is strictly worse than specifying the name in the source file. Note that despite the name, this doesn't set the import path--only the package name. (This is relevant when generating output files; see below.)

## The Go import path
Needed so we can write `import` statements when importing this source file. Set by the `go_package` option if there is one, otherwise guessed by some process I'm not bothering to look up right now.

I'm going to make a possibly controversial assertion here: Every `.proto` file that is used to generate Go code should have a `go_package` option that specifies the import path. It's too late to change now, but we really should just refuse to compile anything that doesn't have a `go_package` option. It's better than trying to guess and *much* better than taking a flag on the command line.

You can pass `--go_out=M=foo.proto=import/path` to specify the import path associated with a particular source file. I cannot imagine ever wanting to do this instead of including a `go_package` option. Nobody should ever need to do this, except possibly for authors of build systems in some strange cases.

## The output file
This is where things get interesting.

Right now, `protoc-gen-go` assumes that it's been called with `--go_out=$GOPATH/src`, and writes output files into the subdirectory corresponding to the generated package's import path. So if you're compiling `foo.proto` which contains a `go_package` option saying that the import path is `github.com/neild/quux`, the generated file goes in `github.com/neild/quux/foo.pb.go`.

You might think that if you passed `--go_out=import_path=some/import/path:. foo.proto` the output file would go in `some/import/path/foo.pb.go`, but no: it goes in `foo.pb.go` instead. That's probably a bug, but it's not really very important (I think) because there's no reason anyone should ever need to specify the import path on the command line--just put it in a `go_package` option.

This all works some of the time, but it breaks down if you don't want to put the generated file in a canonical location relative to `$GOPATH`. What if you're in a vendor directory, for example?

# Proposal(s)

I'm in some directory I have `foo.proto`, the source file contains `option go_package = "github.com/neild/mypackage";`. I want to generate `foo.pb.go` in the current directory. Here are a few possible options.

## Strip a prefix from the path of generated files

```sh
protoc --go_out=trim_import_prefix=github.com/neild/mypackage:. foo.proto
```

The `trim_output_prefix` flag will remove the provided prefix from any generated files. It will generate an error if any files don't start with that prefix.

## Generate files in the current directory

```sh
protoc --go_out=flat:. foo.proto
```

Compared to the previous option, this has the advantage of not requiring the `protoc` line to contain any reference to the import path *if* the `.proto` files are in the same directory that the Go code should go into.

This will not work if you're compiling more than one package at a time. You can't do that today, but we could (and probably should) relax that restriction (#39, #240).

I don't really like this option; it's really easy in the simple case (source and output file in the same directory) but I don't think it generalizes as well.

# RFC

Would either of these features be useful to you? Both? Neither? Something else entirely? Tell me what you think!

neild:
When --go_out=paths=source_relative:., output filenames are always
derived from the input filenames. For example, the output file
for a/b/c.proto will always be a/b/c.pb.go, regardless of the
presence or absence of a go_package option in the input file.

Fixes #515.

Change-Id: I7483449e68f65757bc459233b1644636b98885b9


README.md
```diff
@@ -56,13 +56,43 @@ parameter set to the directory you want to output the Go code to.
...
+name and import path for this package is specified with the `go_package`
+proto option:
+
+	option go_package = "github.com/golang/protobuf/ptypes/any";
+
+The protocol buffer compiler will attempt to derive a package name and
+import path if when a `go_package` option is not present, but it is
+best to always specify one explicitly.
+
+The filename of generated files may be produced in one of two ways.
```
dsnet: Perhaps:
> The output name of the generated file is always produced the same way where the `.proto` suffix is replaced with `.pb.go` (e.g., "foo.proto" is named "foo.pb.go"). 
However, the output directory is selected in one of two ways.

README.md
```diff
@@ -56,13 +56,43 @@ parameter set to the directory you want to output the Go code to.
...
+
+	protoc --go_out=. inputs/x.proto
+	# writes ./github.com/golang/protobuf/p/x.pb.go
+
+  (This can work well with `--go_out=$GOPATH`.)
+
+- Relative to the input file:
+
+	protoc --go_out=paths=source_relative:. inputs/x.proto
+	# generate ./inputs/x.proto
```
dsnet: Don't you mean `./inputs/x.pb.go`?

README.md
```diff
@@ -170,16 +200,17 @@ To create and play with a Test object from the example package,
...
 parameter list separated from the output directory by a colon:
 
-
 	protoc --go_out=plugins=grpc,import_path=mypackage:. *.proto
 
-
 - `import_prefix=xxx` - a prefix that is added onto the beginning of
-  all imports. Useful for things like generating protos in a
-  subdirectory, or regenerating vendored protobufs in-place.
+  all imports. Supported for backwards compatibility, but rarely useful.
```
dsnet: Perhaps use the word "deprecated" somewhere in here?

protoc-gen-go/generator/generator.go
```diff
@@ -625,6 +637,15 @@ func (g *Generator) CommandLineParameters(parameter string) {
 			g.ImportPrefix = v
 		case "import_path":
 			g.PackageImportPath = v
+		case "paths":
+			switch v {
+			case "import":
```
dsnet: Why is this undocumented?

protoc-gen-go/golden_test.go
```diff
@@ -56,7 +56,7 @@ func TestGolden(t *testing.T) {
 			"protoc",
 			"--plugin=protoc-gen-go="+os.Args[0],
 			"-Itestdata",
-			"--go_out=plugins=grpc:"+workdir,
+			"--go_out=paths=source_relative,plugins=grpc:"+workdir,
```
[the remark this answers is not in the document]
neild: Added a couple test cases to the parameters test.

README.md
```diff
@@ -56,13 +56,43 @@ parameter set to the directory you want to output the Go code to.
 The generated files will be suffixed .pb.go.  See the Test code below
 for an example using such a file.
 
+## Packages and input paths ##
+
+The protocol buffer language has a concept of "packages" which does not
+correspond well to the Go notion of packages. In generated Go code,
+each source `.proto` file is associated with a single Go package. The
```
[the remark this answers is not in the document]
neild: I don't think "can only be associated with" is any less 1-1. Reworked this section a bit; PTAL.

README.md
```diff
@@ -56,13 +56,43 @@ parameter set to the directory you want to output the Go code to.
...
+name and import path for this package is specified with the `go_package`
+proto option:
+
+	option go_package = "github.com/golang/protobuf/ptypes/any";
+
+The protocol buffer compiler will attempt to derive a package name and
+import path if when a `go_package` option is not present, but it is
```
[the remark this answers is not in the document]
neild: done

README.md
neild: Phrased a bit differently; PTAL.

dsnet:
Yay... git pull does a merge instead of rebase.

README.md
neild: fixed

README.md
neild: Moved to a "deprecated" section.

protoc-gen-go/generator/generator.go
neild: fixed

neild:
I have absolutely no idea what's gone wrong here. Is this fixable, or should I just start a new PR?

dsnet:
You could do some "git rebase" magic and then do a "git push -f"...

I would just create a new PR.

neild:
https://github.com/golang/protobuf/pull/544

PR #536, opened 2018-02-27
Rename generated Marshal and Unmarshal protobuf method

### LINKED ISSUE #530: [dev] protoc-gen-go: use different name for newly generated Marshal and Unmarshal
The current golang/protobuf API allows the user to write their own Marshal and Unmarshal methods.
https://github.com/golang/protobuf/blob/0a4f71a498b7c4812f64969510bcb4eca251e33a/proto/encode.go#L232
This allows them to write efficient versions of these methods.
Yes, you can write even more efficient marshaling code than the code that gogoprotobuf generates using the specifics of a message.

The new `dev` branch generates Marshal and Unmarshal methods for all messages, that refer back to the library.  Although this can be nice and solves some other problems, this will break other users' code and give the following compile error:

```
./pop.go:36:6: method redeclared: MyType.Marshal
	method(*MyType) func([]byte, bool) ([]byte, error)
	method(MyType) func() ([]byte, error)
```

I would like to not remove this option from users.
I propose that these method names be changed to not break users' code.

I discovered this while merging the dev branch into gogoprotobuf.
I can easily check whether someone generates a Marshal method and then not generate the table Marshal method, but I cannot check whether someone has manually written an even more efficient Marshal method.

dsnet:
As part of the recent protobuf optimization work, we started generating
Unmarshal and Marshal methods on generated messages. These were done to
support the table-driven implementation within the proto package.

We are renaming Marshal as XXX_Marshal and Unmarshal as XXX_Unmarshal
to make it clear that users are not supposed to call these directly.
There are several reasons for the rename:
* The Marshal method assumes that deterministic is the only option we
would ever support for protos, and is not very forward compatible.
* The presence of a deterministic boolean is confusing for users,
where many set it without considering whether it is necessary.
* The Unmarshal method has a slightly different semantic than the previously
documented proto.Unmarshaler interface. The documented Unmarshal specifies
that Reset is called, while the new method does no such thing. The semantic
difference warrants a rename of the method.
* Some users in the Go community depend on these methods not being
generated by default.

Fixes #530

### DIFF
--- a/_conformance/conformance.pb.go
+++ b/_conformance/conformance.pb.go
@@ -123,10 +123,10 @@ func (m *ConformanceRequest) Reset()                    { *m = ConformanceReques
 func (m *ConformanceRequest) String() string            { return proto.CompactTextString(m) }
 func (*ConformanceRequest) ProtoMessage()               {}
 func (*ConformanceRequest) Descriptor() ([]byte, []int) { return fileDescriptor0, []int{0} }
-func (m *ConformanceRequest) Unmarshal(b []byte) error {
+func (m *ConformanceRequest) XXX_Unmarshal(b []byte) error {
 	return xxx_messageInfo_ConformanceRequest.Unmarshal(m, b)
 }
-func (m *ConformanceRequest) Marshal(b []byte, deterministic bool) ([]byte, error) {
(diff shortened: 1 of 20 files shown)


dsnet:
\cc @awalterschulze

awalterschulze:
At first glance this looks great :)
Thank you so much.
So quick :)

PR #542, opened 2018-03-01
protoc-gen-go: don't depend on input file ordering

neild:
Remove generate code dependencies on the order in which source files were
provided to the compiler. In other words, "protoc a.proto b.proto" produces
the same output as "protoc b.proto a.proto".

- Include the proto package version assertion in all files.

- Use the source file name and contents to generate unique var names for
  file descriptors, rather than the file index. In other words,
  "fileDescriptor_imp_81275c260ac30f8b" rather than "fileDescriptor0".

Makes the generated code more stable in the face of unrelated changes
(i.e., adding a new file to a package won't cause generated code for
other files in irrelevant ways), as well as trivial changes in the
protoc command line.

Removes the requirement that all files in a package be compiled at
the same time. Compiling each file individually will produce the same
results as compiling all at once.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -40,6 +40,8 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"go/parser"
 	"go/printer"
@@ -258,7 +260,7 @@ type FileDescriptor struct {
 	// This is used for supporting public imports.
 	exported map[Object][]symbol
 
-	index int // The index of this file in the list of files to generate code for
+	fingerprint string // Fingerprint of this file's contents.
 
 	proto3 bool // whether to generate proto3 code for this file
 }
@@ -269,7 +271,10 @@ func (d *FileDescriptor) PackageName() string { return uniquePackageOf(d.FileDes
 // VarName is the variable name we'll use in the generated code to refer
 // to the compressed bytes of this descriptor. It is not exported, so
 // it is only valid inside the generated package.
-func (d *FileDescriptor) VarName() string { return fmt.Sprintf("fileDescriptor%d", d.index) }
+func (d *FileDescriptor) VarName() string {
+	name := strings.Map(badToUnderscore, baseName(d.GetName()))
+	return fmt.Sprintf("fileDescriptor_%s_%s", name, d.fingerprint)
+}
 
 // goPackageOption interprets the file's go_package option.
 // If there is no go_package, it returns ("", "", false).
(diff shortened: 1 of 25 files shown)


awalterschulze:
Sounds good. As long as the code generation also stays deterministic for
repeated runs.

neild:
Yes, the code generation is deterministic so long as the FileDescriptorProto is. It uses the filename and a hash of the descriptor to ensure uniqueness.

dsnet:
Yay, gofmt changes break tests! :)

dsnet:
> Yes, the code generation is deterministic so long as the FileDescriptorProto is.

That's a important assumption. We should get confirmation from the protoc owners.

protoc-gen-go/generator/generator.go
```diff
@@ -847,11 +852,27 @@ func (g *Generator) WrapTypes() {
...
+		fd.fingerprint = fingerprint
 		g.genFiles = append(g.genFiles, fd)
 	}
 }
 
+// fingerprintProto returns a fingerprint for a message.
+// The fingerprint is intended to prevent conflicts between generated fileds,
+// not to provide cryptographic security.
+func fingerprintProto(m proto.Message) (string, error) {
+	b, err := proto.Marshal(m)
```
dsnet: Shouldn't we be setting the deterministic flag? 

Note that this can still be non-deterministic if the `FileDescriptorProto` as given to us has non-deterministic serialization in an unknown field or field extension. However, this is very unlikely to happen.

dsnet: Also, if it is non-deterministic, it just means we create two variables for what are semantically the same proto, right? No a bad failure scenario.

neild: If it's non-deterministic, we'll generate a different var name each time we compile the source file.  Not really the end of the world.

dsnet: Thinking about it more, doesn't that break assumptions about how Bazel works? It relies on a generation step to produce deterministic output, right?

neild: As for setting the deterministic flag...we could, but we're already marshalling this exact proto elsewhere without it, and I'd rather make that change separately if we're going to do it.

dsnet: It seems we're not any more or less deterministic than before. SGTM.

dsymonds:
Why does this fingerprint the entire FileDescriptorProto, rather than just, say, hashing the filename? I'm having a bit of a nuisance due to these hashes changing for apparently no good reason (the .proto doesn't change, and the gzipped FileDescriptorProto bytes don't even change!).

Actually, now that I've written that, I can't understand what I'm seeing. If the FileDescriptorProto bytes don't change, then the computed hash can't change. But the only difference is that the `source_code_info` field is removed before marshaling+gzipping the descriptor, so the difference must be in there.

Actually, I tracked it down. Between protoc 3.5.1 and 3.6.1, the source_code_info data changed somehow. This explains what I was seeing.

Anyway, to return to my initial question:
   * Why hash the encoded FileDescriptorProto instead of just the filename?
   * If that hash remains, can we skip the source_code_info component in the same way that generating the gzipped byte slice does so that protoc changes don't affect the hash?

neild:
> Why hash the encoded FileDescriptorProto instead of just the filename?

I don't remember. Just the filename is probably fine.

neild:
Changed to use a hash of the filename alone.

dsymonds:
Thanks @neild!

PR #543, opened 2018-03-02
protoc-gen-go: handle package import names per-file

neild:
Rework the handling of imported packages.

Do not add imported protobuf packages to a global registry. Instead,
maintain a per-file list of packages and names. This avoids imports
in one file affecting the generated code in a second one.

Consistently deal with imports in terms of import path, rather than
source file. This avoids imports of two .proto files in the same
Go package producing redundant import statements.

Introduce ImportPath and PackageName types to prevent confusion about
which values contain import paths ("github.com/golang/protobuf/proto")
and which contain package names ("proto").

Convert many uses of FileDescriptorProto to FileDescriptor, for
consistency and general convenience.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -92,6 +92,12 @@ func RegisterPlugin(p Plugin) {
 	plugins = append(plugins, p)
 }
 
+// An ImportPath is the import path of a Go package. e.g., "github.com/golang/protobuf".
+type ImportPath string
+
+// A PackageName is the name of a Go package. e.g., "protobuf".
+type PackageName string
+
 // Each type we import as a protocol buffer (other than FileDescriptorProto) needs
 // a pointer to the FileDescriptorProto that represents it.  These types achieve that
 // wrapping by placing each Proto inside a struct with the pointer to its File. The
@@ -100,19 +106,21 @@ func RegisterPlugin(p Plugin) {
 
 // The file and package name method are common to messages and enums.
 type common struct {
-	file *descriptor.FileDescriptorProto // File this object comes from.
+	file *FileDescriptor // File this object comes from.
 }
 
-// PackageName is name in the package clause in the generated file.
-func (c *common) PackageName() string { return uniquePackageOf(c.file) }
+// ImportPath is the import path of the Go package containing the type.
+func (c *common) ImportPath() ImportPath {
+	return c.file.importPath
+}
 
-func (c *common) File() *descriptor.FileDescriptorProto { return c.file }
+func (c *common) File() *FileDescriptor { return c.file }
 
 func fileIsProto3(file *descriptor.FileDescriptorProto) bool {
 	return file.GetSyntax() == "proto3"
 }
 
-func (c *common) proto3() bool { return fileIsProto3(c.file) }
+func (c *common) proto3() bool { return fileIsProto3(c.file.FileDescriptorProto) }
 
 // Descriptor represents a protocol buffer message.
 type Descriptor struct {
@@ -260,13 +268,11 @@ type FileDescriptor struct {
 	// This is used for supporting public imports.
 	exported map[Object][]symbol
 
-	fingerprint string // Fingerprint of this file's contents.
+	fingerprint string     // Fingerprint of this file's contents.
+	importPath  ImportPath // Import path of this file's package.
 
 	proto3 bool // whether to generate proto3 code for this file
 }
-
-// PackageName is the package name we'll use in the generated code to refer to this file.
-func (d *FileDescriptor) PackageName() string { return uniquePackageOf(d.FileDescriptorProto) }
 
 // VarName is the variable name we'll use in the generated code to refer
 // to the compressed bytes of this descriptor. It is not exported, so
@@ -280,25 +286,22 @@ func (d *FileDescriptor) VarName() string {
 // If there is no go_package, it returns ("", "", false).
 // If there's a simple name, it returns ("", pkg, true).
 // If the option implies an import path, it returns (impPath, pkg, true).
-func (d *FileDescriptor) goPackageOption() (impPath, pkg string, ok bool) {
-	pkg = d.GetOptions().GetGoPackage()
-	if pkg == "" {
-		return
+func (d *FileDescriptor) goPackageOption() (impPath ImportPath, pkg PackageName, ok bool) {
+	opt := d.GetOptions().GetGoPackage()
+	if opt == "" {
+		return "", "", false
 	}
-	ok = true
 	// The presence of a slash implies there's an import path.
-	slash := strings.LastIndex(pkg, "/")
+	slash := strings.LastIndex(opt, "/")
 	if slash < 0 {
-		return
+		return "", cleanPackageName(opt), true
 	}
-	impPath, pkg = pkg, pkg[slash+1:]
 	// A semicolon-delimited suffix overrides the package name.
-	sc := strings.IndexByte(impPath, ';')
+	sc := strings.Index(opt, ";")
 	if sc < 0 {
-		return
+		return ImportPath(opt), cleanPackageName(opt[slash+1:]), true
 	}
-	impPath, pkg = impPath[:sc], impPath[sc+1:]
-	return
+	return ImportPath(opt[:sc]), cleanPackageName(opt[sc+1:]), true
 }
 
 // goPackageName returns the Go package name to use in the
@@ -306,18 +309,18 @@ func (d *FileDescriptor) goPackageOption() (impPath, pkg string, ok bool) {
 // came from an option go_package statement.  If explicit is false,
 // the name was derived from the protocol buffer's package statement
 // or the input file name.
-func (d *FileDescriptor) goPackageName() (name string, explicit bool) {
+func (d *FileDescriptor) goPackageName() (name PackageName, explicit bool) {
 	// Does the file have a "go_package" option?
 	if _, pkg, ok := d.goPackageOption(); ok {
 		return pkg, true
 	}
 
 	// Does the file have a package clause?
-	if pkg := d.GetPackage(); pkg != "" {
-		return pkg, false
+	if p := d.GetPackage(); p != "" {
+		return cleanPackageName(p), false
 	}
 	// Use the file base name.
-	return baseName(d.GetName()), false
+	return cleanPackageName(baseName(d.GetName())), false
 }
 
 // goFileName returns the output name for the generated Go file.
@@ -333,7 +336,7 @@ func (d *FileDescriptor) goFileName() string {
 	if impPath, _, ok := d.goPackageOption(); ok && impPath != "" {
 		// Replace the existing dirname with the declared import path.
 		_, name = path.Split(name)
-		name = path.Join(impPath, name)
+		name = path.Join(string(impPath), name)
 		return name
 	}
 
@@ -348,7 +351,7 @@ func (d *FileDescriptor) addExport(obj Object, sym symbol) {
 type symbol interface {
 	// GenerateAlias should generate an appropriate alias
 	// for the symbol from the named package.
-	GenerateAlias(g *Generator, pkg string)
+	GenerateAlias(g *Generator, pkg PackageName)
 }
 
 type messageSymbol struct {
@@ -365,8 +368,8 @@ type getterSymbol struct {
 	genType  bool   // whether typ contains a generated type (message/group/enum)
 }
 
-func (ms *messageSymbol) GenerateAlias(g *Generator, pkg string) {
-	remoteSym := pkg + "." + ms.sym
+func (ms *messageSymbol) GenerateAlias(g *Generator, pkg PackageName) {
+	remoteSym := string(pkg) + "." + ms.sym
 
 	g.P("type ", ms.sym, " ", remoteSym)
 	g.P("func (m *", ms.sym, ") Reset() { (*", remoteSym, ")(m).Reset() }")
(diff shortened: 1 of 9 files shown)


neild:
The .pb.go diffs should give a sense as to what this is changing.

neild:
Sigh, forgot that "make regenerate" doesn't actually do anything unless you install protoc-gen-go. Need to fix that.

protoc-gen-go/generator/generator.go
```diff
@@ -92,6 +92,12 @@ func RegisterPlugin(p Plugin) {
 	plugins = append(plugins, p)
 }
 
+// An ImportPath is the import path of a Go package. e.g., "github.com/golang/protobuf".
```
dsnet: Nit: "google.golang.org/genproto/protobuf"

since "github.com/golang/protobuf" is not a go package.

protoc-gen-go/generator/generator.go
```diff
@@ -280,44 +286,41 @@ func (d *FileDescriptor) VarName() string {
...
+	if opt == "" {
+		return "", "", false
 	}
-	ok = true
 	// The presence of a slash implies there's an import path.
-	slash := strings.LastIndex(pkg, "/")
+	slash := strings.LastIndex(opt, "/")
 	if slash < 0 {
-		return
+		return "", cleanPackageName(opt), true
```
dsnet: I'm surprised we even need to clean this. I expect `protoc-gen-go` to report an error if the package name supplied by the user is invalid.

protoc-gen-go/generator/generator.go
```diff
@@ -280,44 +286,41 @@ func (d *FileDescriptor) VarName() string {
...
 
 	// Does the file have a package clause?
-	if pkg := d.GetPackage(); pkg != "" {
-		return pkg, false
+	if p := d.GetPackage(); p != "" {
+		return cleanPackageName(p), false
 	}
 	// Use the file base name.
-	return baseName(d.GetName()), false
+	return cleanPackageName(baseName(d.GetName())), false
```
dsnet: ... really... a proto file may be missing a package statement!?

Interestingly, the BNF even permits there to be [multiple package clauses](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#proto_file):
```
proto = syntax { import | package | option | topLevelDef | emptyStatement }
```

protoc-gen-go/generator/generator.go
```diff
@@ -92,6 +92,12 @@ func RegisterPlugin(p Plugin) {
 	plugins = append(plugins, p)
 }
 
+// An ImportPath is the import path of a Go package. e.g., "github.com/golang/protobuf".
+type ImportPath string
```
dsnet: Rename as `GoImportPath`? Clearly distinguishes it from a proto import path, which is a legitimate concept.

protoc-gen-go/generator/generator.go
```diff
@@ -92,6 +92,12 @@ func RegisterPlugin(p Plugin) {
...
+
+// A PackageName is the name of a Go package. e.g., "protobuf".
+type PackageName string
```
dsnet: Ditto, rename as `GoPackageName` to disambiguate from proto packages?

protoc-gen-go/generator/generator.go
```diff
@@ -662,35 +655,40 @@ func (g *Generator) CommandLineParameters(parameter string) {
...
+	}
+	g.packageNames[importPath] = name
+	g.usedPackageNames[name] = true
+	return name
+}
 
-// Package names already registered.  Key is the name from the .proto file;
-// value is the name that appears in the generated code.
-var pkgNamesInUse = make(map[string]bool)
+var globalPackageNames = map[PackageName]int{
```
dsnet: What does the integer mean?

protoc-gen-go/generator/generator.go
```diff
@@ -662,35 +655,40 @@ func (g *Generator) CommandLineParameters(parameter string) {
...
-	for i, orig := 1, pkg; pkgNamesInUse[pkg]; i++ {
-		// It's a duplicate; must rename.
-		pkg = orig + strconv.Itoa(i)
-	}
-	// Install it.
-	pkgNamesInUse[pkg] = true
-	if f != nil {
-		uniquePackageName[f.FileDescriptorProto] = pkg
+	name := cleanPackageName(pkg)
+	globalPackageNames[name]++
```
dsnet: This seems wrong. Should you only increment if there is a conflict?

protoc-gen-go/generator/generator.go
```diff
@@ -1136,6 +1120,10 @@ func (g *Generator) printAtom(v interface{}) {
 		fmt.Fprint(g, v)
 	case *float64:
 		fmt.Fprint(g, *v)
+	case PackageName:
+		g.WriteString(string(v))
+	case ImportPath:
+		g.WriteString(string(v))
```
dsnet: Shouldn't this be `g.WriteString(strconv.Quote(string(v)))`?
That way you don't need to keep calling strconv.Quote manually below.

protoc-gen-go/generator/generator.go
```diff
@@ -1424,32 +1406,43 @@ func (g *Generator) generateImports() {
...
 			continue
 		}
 		// We need to import all the dependencies, even if we don't reference them,
 		// because other code and tools depend on having the full transitive closure
 		// of protocol buffer types in the binary.
-		pname := fd.PackageName()
-		if _, ok := g.usedPackages[pname]; !ok {
-			pname = "_"
+		if _, ok := g.usedPackages[importPath]; !ok {
+			packageName = "_"
```
dsnet: I'm not sure I understand from the comment why this is necessary, but whatever.

protoc-gen-go/generator/generator.go
neild: done

protoc-gen-go/generator/generator.go
neild: Good point. Removed.

protoc-gen-go/generator/generator.go
neild: done

protoc-gen-go/generator/generator.go
neild: done

protoc-gen-go/generator/generator.go
neild: Me being too clever. Rewrote to not be clever.

protoc-gen-go/generator/generator.go
neild: Good thought. Done.

protoc-gen-go/generator/generator.go
neild: If a `.proto` file imports `a.proto`, we need to import `a.proto`'s Go package for its side effects (i.e., the things it registers) even if we don't reference any of the symbols in it.

In a better world, protos would be IWYU clean and this wouldn't be necessary.

dsnet: That's not exactly what the comment says. Worth updating?

protoc-gen-go/grpc/grpc.go
```diff
@@ -130,8 +130,8 @@ func (g *grpc) GenerateImports(file *generator.FileDescriptor) {
 		return
 	}
 	g.P("import (")
-	g.P(contextPkg, " ", strconv.Quote(path.Join(g.gen.ImportPrefix, contextPkgPath)))
-	g.P(grpcPkg, " ", strconv.Quote(path.Join(g.gen.ImportPrefix, grpcPkgPath)))
+	g.P(contextPkg, " ", strconv.Quote(path.Join(string(g.gen.ImportPrefix), contextPkgPath)))
```
dsnet: Could just do:
```go
generator.GoImportPath(path.Join(string(g.gen.ImportPrefix), context.PkgPath))
```
instead of:
```go
strconv.Quote(path.Join(string(g.gen.ImportPrefix), contextPkgPath))
```
to be type correct.

Perhaps even make Join a method of `ImportPrefix`:
```go
g.gen.ImportPrefix.Join(context.PkgPath)
```

protoc-gen-go/generator/generator.go
```diff
@@ -655,42 +655,41 @@ func (g *Generator) CommandLineParameters(parameter string) {
...
 
 // Create and remember a guaranteed unique package name. Pkg is the candidate name.
 // The FileDescriptor parameter is unused.
 func RegisterUniquePackageName(pkg string, f *FileDescriptor) string {
 	name := cleanPackageName(pkg)
-	globalPackageNames[name]++
-	if n := globalPackageNames[name]; n > 1 {
-		return pkg + strconv.Itoa(n)
+	for i, orig := 1, name; globalPackageNames[name]; i++ {
+		name = orig + GoPackageName(strconv.Itoa(i))
```
dsnet: Don't you still have to mutate the `globalPackageNames` eventually?

protoc-gen-go/generator/generator.go
```diff
@@ -1343,7 +1342,7 @@ func (g *Generator) generateHeader() {
 	if !haveImportPath {
 		g.P("package ", name)
 	} else {
-		g.P("package ", name, " // import ", strconv.Quote(g.ImportPrefix+string(importPath)))
+		g.P("package ", name, " // import ", g.ImportPrefix+importPath)
```
dsnet: Why is this `path.Join` in the `grpc` package but a straight concatenation here?

protoc-gen-go/generator/generator.go
neild: I think the comment is fine as is; if you want to update it in a different cl, feel free.

protoc-gen-go/grpc/grpc.go
neild: Done.

Didn't make Join a method, because ImportPrefix is terrible and doesn't need more support.

protoc-gen-go/generator/generator.go
neild: Ack. Done.

This should have been caught by a test, but takes a lot of infrastructure to test. Do you want one?

protoc-gen-go/generator/generator.go
neild: Because ImportPrefix is completely broken.

protoc-gen-go/generator/generator.go
dsnet: This is only to support plugins, so whatever. We can add a test in the future.

protoc-gen-go/generator/generator.go
neild: That was harsh, so let me expand.

The fact that import_prefix rewrites standard imports (proto, grpc, context) makes no sense. The fact that it rewrites them inconsistently is almost irrelevant, since nobody wanted that rewriting in the first place.

We could try to fix it, but everything that it does is done better and more reasonably by the `go_package` option.

meling:
@dsnet @neild Is this PR discussed in a separate issue, where the rational is better explained? I've had a brief discussion with @awalterschulze related to this change which impacts a change I proposed here and presumably can cause some issues with merging this change.
https://github.com/gogo/protobuf/issues/338
https://github.com/gogo/protobuf/pull/341

As a first insight here, I think the current approach to handle imports is pretty flawed. I would suggest that imports generated from any proto file or plugin should be added to a single global registry, and only emitted to the generated output file when all imports have been collected and made unique. That way there wouldn't be a need generate unique import aliases for the same import.

(Caveat: I'm raising this issue without having had time to investigate this issue in depth, and what might be the problem with such a revised import handling strategy. I'm sure there could be compatibility issues, but there are already many such issues arising from the `dev` branch, so why not fix it properly.)

Secondary, if the import handling cannot be fixed to the preferred approach mentioned above, maybe the following changes be integrated into golang/protobuf:

https://github.com/gogo/protobuf/pull/341/files#diff-c5ab46099bd055be2629de524005c350L717
https://github.com/gogo/protobuf/pull/341/files#diff-4cb022973f0dc3083ae64c102f7e4cc9L132

neild:
This change has two main effects:

- We consistently identify packages by their import paths. A Go package is uniquely identified by its import path. The previous behavior was confusing and produced surprising results in a number of circumstances. e.g., importing the same package several times in the same file under different names.

- The contents of one source file do not affect the generated code for an unrelated file. It is surprising for the contents of two generated files to vary based on what order they are provided to `protoc`, or for a change in one file to produce a change in the generated code for a completely unrelated file. This will be particularly important when we address #39 (compiling multiple Go packages in one `protoc` invocation).

If I understand the gogo/protobuf change you're asking us to import, it doesn't seem related to this PR at all, however. As far as I can tell, it's changing the import aliases generated by the compiler from lengthy ones based on the full import path (e.g., "github_com_gogo_protobuf_jsonpb") to short ones based on the basename of the import path (e.g., "jsonpb"). The golang/protobuf compiler already generates the latter style, and (so far as I know) always has. The former style with the underscores seems to be a change introduced by gogo/protobuf.

meling:
Thanks for your explanation @neild, and I apologize for my ignorance. I've mainly used gogo/protobuf due to other features. Is it correct that the golang/protobuf behavior is to supply only unique import paths also from plugins??

@awalterschulze: Could you weigh in on the current approach with "full import path" as import aliases instead of the short ones that @neild refers to? I don't know the rationale for it; I thought it was from golang/protobuf and not gogo/protobuf. I mean if golang/protobuf doesn't produce these long names, I won't need my fix, unless there is some gogo/protobuf reason for them...

(Apologies for not having had the time to examine code and testing things yet... but wanted to ask anyway.)

neild:
When the proto compiler generates an import statement, we always generate a local alias for the import. This isn't always strictly necessary, but is safest given that there's no guaranteed way to map from import path to package name.

If a single file imports two packages with the same name (e.g., `github.com/golang/protobuf/ptypes/duration` and `github.com/google/go-genproto/duration`), we need to disambiguate somehow. We do that by adding a number to one of them, so `duration` and `duration1`.

This is necessary even without plugins; a single proto file might import two other packages with conflicting names.

awalterschulze:
Please forgive me, but after merging in all the changes from the `dev` branch, I am tired.

This tiredness means I didn't properly look at this, because now I am beyond lazy. It could be gogoprotobuf or golang/protobuf, but after the merge, the generated imports seems to include some long underscorey imports.  I just assumed this was because the part of the code that @meling changed originally, to fix this, was affected by the merge, that it was something that golang/protobuf changed.  So I am sorry, I didn't do my homework.

Sorry to drop the ball :(

PR #546, opened 2018-03-07
 protobuf: Delete makefiles, regenerate protos consistently

neild:
Delete all the Makefiles except for the top-level one.
Recursive make is difficult to understand, keeping all
the Makefiles in sync is difficult (as witness the fact
that there were numerous inconsistencies between them),
and mostly they just ran trivial commands like "go test".

Replace all regeneration of .pb.go files with a single
regenerate.sh shell script.

Stop relying on the protobuf repository being installed
in ~/src/protobuf. The only things we need out of the
repository are standard .proto source files, which
protoc is perfectly capable of discovering on its own.

Regenerate .pb.go files which had been missed by
"make regenerate" before.


paranoiacblack:
LGTM

hitzhangjie:
Hi, why do we delete the Makefile? I think it's easier to install by run `make install`.

Now, how should I install this? `go install`?

hitzhangjie:
The toplevel Makefile doesn't exist!

Issue #479, opened 2018-01-05
process: merge dev branch into master


afiune:
Hi, 👋 

This is a simple **question** in regards the release process of this repository;

> **When does 'dev' branch gets merge to 'master'?**

Thanks in advance.
  
![tenor-174386324](https://user-images.githubusercontent.com/5712253/34618553-1807615a-f20d-11e7-800a-eabc4d0647f8.gif)

dsnet:
Official announcement made: https://groups.google.com/forum/#!topic/golang-nuts/F5xFHTfwRnY

Target date: April 30th.

dsnet:
This happened.

PR #548, opened 2018-03-08
protoc-gen-go: use standard library context (requires Go1.9)

### LINKED ISSUE #537: protoc-gen-go: contextPkgPath should use "context"
[As of Go 1.7 this package(`golang.org/x/net/context`) is available in the standard library under the name context. https://golang.org/pkg/context. ](https://godoc.org/golang.org/x/net/context) So we should replace contextPkgPath  golang.org/x/net/context with context

dsnet:
This change effectively makes Go1.9 the minimum required version of Go
when using gRPC since type aliases are needed to truly unify context.
Even though this only affects those who use gRPC, there are other reasons we
need Go1.9 (such as type aliases for import public).

Fixes #537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 sudo: false
 language: go
 go:
-- 1.6.x
-- 1.7.x
-- 1.8.x
 - 1.9.x
 - 1.10.x
 - 1.x
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Google's data interchange format.
 Copyright 2010 The Go Authors.
 https://github.com/golang/protobuf
 
-This package and the code it generates requires at least Go 1.6.
+This package and the code it generates requires at least Go 1.9.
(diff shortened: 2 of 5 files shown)


dsnet:
Even though AppEngine classic is still not on Go1.9, it should happen by the time we merge `dev` into `master` (#479).

neild:
This doesn't make Go 1.9 the minimum unless you need to mix code which uses "golang.org/x/net/context" and "context". If you consistently use one or the other, there's no need for an alias.

dfawley:
> This doesn't make Go 1.9 the minimum unless you need to mix code which uses
> "golang.org/x/net/context" and "context". If you consistently use one or the
> other, there's no need for an alias.

gRPC-Go imports `x/net/context`, so yes, this does make Go 1.9 the minimum version supportable by gRPC-Go.  This is not OK for us.

Issue #397, opened 2017-07-25
conformance: document how to run tests and add in failure whitelist


cybrcodr:
It would be nice to add more documentation on how to run the conformance tests or even perhaps have a script/Makefile target for it.

Conformance test runner is at https://github.com/google/protobuf/tree/master/conformance.

It'll be good to check in the failure whitelist as well for tracking.

PR #552, opened 2018-03-09
conformance: clean up, fix conformance tests

neild:
Move _conformance directory to conformance. (The _ was apparently there
to prevent "go install ./..." from touching that directory, which
doesn't seem like a sufficient reason to have a weird path name.)

Delete conformance_proto directory, regenerate the proto at test time.
(This proto is used for communications between the test runner and test
process; nothing else needs it.)

Add a list of known failures.

Add a Makefile so that "make test" in the conformance directory does the
right thing. (Requires that github.com/google/protobuf be installed
somewhere.) Include documentation in "make all" output. Skip tests if
the necessary dependencies don't exist.


neild:
I have no idea why CI is failing on this; it seems to be hitting a problem while running `go get`.

```
# cd /home/travis/gopath/src/github.com/golang/protobuf; git pull --ff-only
fatal: Not possible to fast-forward, aborting.
```

cybrcodr:
This PR should address issue #397.

cybrcodr:
I think the issue is with ...
package github.com/golang/protobuf/conformance/internal/conformance_proto: exit status 128

That is a generated file that conformance.go depends on but does not exist?

neild:
> That is a generated file that conformance.go depends on but does not exist yet?

Ah, that's it. I thought I could get away without checking in the `.pb.go` and just generating it at build time, but that breaks too many things. Thanks.

_conformance/conformance_proto/conformance.proto
```diff
@@ -1,285 +0,0 @@
-// Protocol Buffers - Google's data interchange format
```
dsnet: IIUC, the removal of this file means golden_test.go won't catch when the generated .pb.go drifts from one generated from the version in the proto repo. It seems to me that we should either:
1) avoid checking in `conformance.pb.go` and rely on the make script to automatically regenerate it upon execution
2) checking in both the `.proto` and `.pb.go` and add a test to ensure that the checked-in and upstream `.proto` files remain the same.
3) add a test to ensure the checked in `.pb.go` doesn't drift when dynamically regenerated from upstream.

conformance/conformance.go
```diff
@@ -124,11 +118,7 @@ func handle(req *pb.ConformanceRequest) *pb.ConformanceResponse {
 	switch req.RequestedOutputFormat {
 	case pb.WireFormat_PROTOBUF:
 		var p []byte
-		if isProto3 {
-			p, err = proto.Marshal(&msg)
-		} else {
-			p, err = proto.Marshal(&msg1)
-		}
+		p, err = proto.Marshal(&msg)
```
dsnet: `p, err :=`

conformance/conformance.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd $(dirname $0)
+exec go run conformance.go $*
```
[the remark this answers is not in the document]
dsnet: Also, what's the purpose of `$*`? The conformance.go program doesn't care about the input args.

conformance/failure_list_go.txt
```diff
@@ -0,0 +1,61 @@
+# This is the list of conformance tests that are known ot fail right now.
+# These should be fixed.
```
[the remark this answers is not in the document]
neild: done

conformance/conformance.sh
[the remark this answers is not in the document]
neild: Habit, mostly: This script is just a proxy for building and running the conformance binary, so it turns itself into that binary.

_conformance/conformance_proto/conformance.proto
neild: Checked in the .proto, made "make test" in the conformance directory warn if its out of date.

conformance/conformance.go
neild: done

dsnet: I don't see any changes. Did you add this file to the latest commit?

neild: I guess not? Fixed now.

PR #556, opened 2018-03-12
protoc-gen-go: statically calculate size of oneof wiretags

dsymonds:
There's no need for oneof sizers to compute the size of the tag/wire
varint, since it can be computed at proto compile time instead (and it
is typically a small number like 1 or 2).

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -2428,58 +2428,49 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				}
 				g.P("case *", oneofTypeName[field], ":")
 				val := "x." + fieldNames[field]
-				var wire, varint, fixed string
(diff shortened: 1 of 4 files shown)


dsnet:
Can you rebase against the `dev` branch?

dsymonds:
I could, but I want this on the master branch (i.e. what people get if they don't do anything specific).

dsymonds:
As in, I can do a follow-up PR to cherry-pick this onto the dev branch if you want.

dsnet:
We're not merging anything in master except critical fixes. `master` will be merged into `dev` on April 30th.

dsnet:
If you need this sooner inside Google, we're already pulling directly from the `dev` branch.

dsymonds:
Fine. I'll open a new PR on dev.

PR #563, opened 2018-03-16
protoc-gen-go: fix generation of proxy getters for distant types

neild:
Skip generating getters for fields that are not defined in the same file
as the publicly imported message.

This is the wrong way to fix this problem. It is, however, the expedient
one to hold us over until (soon, I hope) we completely redo public
imports as type aliases and make all of this moot.

The scenario:

  a.proto publicly imports b.proto
  b.proto has a message M
  M has a field F of type T

We generate forwarding methods for public imports.

  // T is defined in the same Go package as a.proto.
  func (m *M) GetF() T { ... }

Depending on what package T is defined in, we might need to qualify its
name (foo.T), and we might need to add an import for that package.
That's not what we used to do, however: Instead, we'd *only* generate
the GetF forwarder if T is defined in b.proto.

Commit 9d4962b4 made it so that we'd generate the forwarder if it T is
defined in the *same Go package* as b.proto, which seems safe and
reasonable. It turns out that something elsewhere in the generator is
getting confused by this, however. Rather than figuring out what--which
would also change a lot of generated code by adding forwarding methods
that didn't used to be there--this just reverts back to the behavior of
looking at files rather than packages.

All of this goes away completely once we start using type aliases,
because we don't need any forwarding methods at all at that point.


dsnet:
Is this up for review or you still working on it?

neild:
Up for review. Sorry, I thought I'd already added you to it.

PR #565, opened 2018-03-20
proto: do not allow unknown fields to satisfy required field bit

dsnet:
A prior changed (see #511) allows Unmarshal to treat fields with
unknown wire types with known tags as unknown fields.
When doing so, we must be careful not to set the required field bit.

For example, suppose we have:
    message Foo { require fixed32 my_field = 1; }

Then parsing a message with a field of type varint and tag 1 should
not satisfy the required that Foo.my_field be set.

### DIFF
--- a/proto/all_test.go
+++ b/proto/all_test.go
@@ -1860,6 +1860,25 @@ func TestRequiredNotSetError(t *testing.T) {
 	}
 }
 
+func TestRequiredNotSetErrorWithBadWireTypes(t *testing.T) {
+	// Required field expects a varint, and properly found a varint.
+	if err := Unmarshal([]byte{0x08, 0x00}, new(GoEnum)); err != nil {
+		t.Errorf("Unmarshal = %v, want nil", err)
+	}
+	// Required field expects a varint, but found a fixed32 instead.
+	if err := Unmarshal([]byte{0x0d, 0x00, 0x00, 0x00, 0x00}, new(GoEnum)); err == nil {
+		t.Errorf("Unmarshal = nil, want RequiredNotSetError")
+	}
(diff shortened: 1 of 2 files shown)


proto/all_test.go
```diff
@@ -1860,6 +1860,21 @@ func TestRequiredNotSetError(t *testing.T) {
 	}
 }
 
+func TestRequiredNotSetErrorWithBadWireTypes(t *testing.T) {
+	// Required field is properly a varint.
+	if err := Unmarshal([]byte{0x08, 0x00}, new(GoEnum)); err != nil {
+		t.Errorf("Unmarshal = %v, want nil", err)
+	}
+	// Required field is a fixed32 instead of varint.
```
cybrcodr: I got confused a bit on the comment.  How about...

// Bytes contain fixed32 for the required varint field.

proto/all_test.go
```diff
@@ -1860,6 +1860,21 @@ func TestRequiredNotSetError(t *testing.T) {
...
+	if err := Unmarshal([]byte{0x0d, 0x00, 0x00, 0x00, 0x00}, new(GoEnum)); err == nil {
+		t.Errorf("Unmarshal = nil, want RequiredNotSetError")
+	}
+	// Required field has both varint and fixed32 (ignored).
```
cybrcodr: Same here.

// Bytes contain both varint and fixed32 for same required varint field.

Btw, given that both encoded fields are for key 1, shouldn't that supposed to be an error?

dsnet: > Btw, given that both encoded fields are for key 1, shouldn't that supposed to be an error?

Not according to the C++ implementation :)

PR #569, opened 2018-03-28
Fix minor godoc typos in any package

johnmccabe:
An extra space was causing Examples 3 and 4 to render in Example 2's code block. This commit fixes those minor typos.

Also removing the `====` denoting the JSON section will allow godoc to render a heading here.

<details><summary>Before - click to expand</summary><p>
screen shot 2018-03-28 at 16 44 55
</p></details>

<details><summary>After - click to expand</summary><p>
screen shot 2018-03-28 at 16 45 12
</p></details>

PR #570, opened 2018-03-28
Fix godoc examples for Any (#569)

neild:
An extra space was causing Examples 3 and 4 to render in Example 2's
code block. This commit fixes those minor typos.

Also removing the `====` denoting the JSON section will allow godoc to
render a heading here.

Signed-off-by: John McCabe <john@johnmccabe.net>

### DIFF
--- a/ptypes/any/any.pb.go
+++ b/ptypes/any/any.pb.go
@@ -43,7 +43,7 @@ const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
 //       foo = any.unpack(Foo.class);
 //     }
 //
-//  Example 3: Pack and unpack a message in Python.
+// Example 3: Pack and unpack a message in Python.
 //
 //     foo = Foo(...)
 //     any = Any()
@@ -53,7 +53,7 @@ const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
 //       any.Unpack(foo)
 //       ...
 //
-//  Example 4: Pack and unpack a message in Go
+// Example 4: Pack and unpack a message in Go
 //
 //      foo := &pb.Foo{...}
 //      any, err := ptypes.MarshalAny(foo)
@@ -71,7 +71,7 @@ const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
 //
 //
 // JSON
-// ====
+//
 // The JSON representation of an `Any` value uses the regular
 // representation of the deserialized, embedded message, with an
 // additional field `@type` which contains the type URL. Example:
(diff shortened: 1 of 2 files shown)


neild:
Not quite sure what the right way to cherrypick from master onto dev is. clabot doesn't seem to like the obvious one.

dsnet:
:/ We probably should have requested the original author rebase on `dev` and do the cherry pick in the future with all others. I'm not sure how make the legal aspect of CLA bot happy here.

It's tempting to just click "Squash and merge", but I don't know what the lawyer cats would say.

neild:
Ah, right. This also doesn't work because it's modifying the wrong any.proto--the dev branch uses the one from the protoc install.

neild:
As for the CLA, clabot says that it's up to us to verify author consent. Given that this is just a cherrypick from a different branch, that seems pretty clear. I guess we can just merge. Or could, except for the bit about changing the wrong source file.

neild:
And, annoyingly, it now occurs to me that changing the right source file is impractical, because there's nothing wrong with it. The problem is that we're taking a comment clearly written in Markdown and sticking it directly into godoc.

So either we need to manually tweak this comment every time we regenerate this file, live with a lousy godoc, or...any ideas?

dsnet:
To the degree that godoc's light syntax overlaps with markdown, I think we should fix the proto file. It seems that it would be safe to remove the double spaces before "Example 3" and "Example 4" as this matches Example 1 and "Example 2"

neild:
The problem is that we'd need to convince the owners of the proto file that they should change the formatting to conform to godoc's expectations. Maybe that's possible; if I was them, I'd be dubious.

dsnet:
Given that proto is designed to be a cross-language DSL, I don't think it's reasonable to expect people to use godoc syntax (or any programming language's documentation syntax). Even the use of Markdown is dubious to me, but if proto files were specified to always use Markdown, then protoc-gen-go could (in theory) translate Markdown to godoc-like syntax.

dsnet:
I took a peak at the godoc for the other well-known types and they also have weird Markdown-isms leaking through

dsnet:
For example, [timestamp.Timestamp](https://godoc.digitalstatic.net/pkg/github.com/golang/protobuf/ptypes/timestamp/#Timestamp) makes use of the hyperlink syntax in Markdown.

johnmccabe:
@neild / @dsnet apologies, I'd not considered the non-Go case in #569 or realised it was being pulled in... my bad. Agree that avoiding language specific doc syntax in the comments is a good move.

As an FYI regarding the clabot I've got an active agreement, not sure why it wasn't picking it up.

screen shot 2018-03-28 at 21 03 08

Issue #408, opened 2017-08-11
jsonpb: mapped enum cannot marshal/unmarshal string


kazegusuri:
```
enum Numeral {
  UNKNOWN = 0;
  ARABIC = 1;
  ROMAN = 2;
}
message Mappy {
  map<string, Numeral> enumy = 6;
}
```

The `enumy` field in `Mappy` message is always marshaled as int, and cannot be unmarshaled from string  enum like `{"enumy": {"foo": "ARABIC", "bar": "ROMAN"}}`
This is probably because generated message field does not have `enum` property and not handled as Enum in `jsonpb.marshalValue`.

The test case is skipped for now.
https://github.com/golang/protobuf/blob/master/jsonpb/jsonpb_test.go#L388-L389

Tim15:
Same issue here.

PR #574, opened 2018-03-29
jsonpb: fix bug in maps with enum values.

vagruchi:
Fix the #408. Marshal parameters for map values are located under `protobuf_val` tag. Jsonpb marshaler uses a tag for map. Added the `subProp` parameter for `marshalField` and `marshalValue` to provide `protobuf_val` properties.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -158,6 +158,10 @@ type wkt interface {
 	XXX_WellKnownType() string
 }
 
+type enum interface {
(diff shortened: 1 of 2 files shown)


vagruchi:
@neild I agree that introducing new parameter is too broad. 
I researched generated code by proto. Seems like enums have EnumDescriptor method and enums can be detected by check for this method. What do you think about it?

neild:
Checking for EnumDescriptor seems fine.

vagruchi:
@neild I changed implementation. Any notes?

PR #575, opened 2018-03-30
proto: ensure consistency of behavior in Register* functions

namusyaka:
At first, `panic` will not work properly on appengine. In fact, in order to avoid this, [deletion by sed is done in appengine official package](https://github.com/golang/appengine/blob/ad39d7fab7c60b2493fdc318c3d2cdb2128f92a4/internal/regen.sh#L39).

This behavior is not a consistent behavior in `Register*` functions.
I think that it is enough to just log back and do a return like `RegisterType` without generating panic.

In addition, `RegisterFile` also changes to make it behave consistently as well.

### DIFF
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -866,6 +866,10 @@ var (
 // RegisterFile is called from generated code and maps from the
 // full file name of a .proto file to its compressed FileDescriptorProto.
 func RegisterFile(filename string, fileDescriptor []byte) {
+	if _, ok := protoFiles[filename]; ok {
+		log.Printf("proto: duplicate proto file registered: %s", filename)
+		return
+	}
 	protoFiles[filename] = fileDescriptor
 }
 
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -812,7 +812,8 @@ var enumValueMaps = make(map[string]map[string]int32)
 // maps into the global table to aid parsing text format protocol buffers.
 func RegisterEnum(typeName string, unusedNameMap map[int32]string, valueMap map[string]int32) {
 	if _, ok := enumValueMaps[typeName]; ok {
-		panic("proto: duplicate enum registered: " + typeName)
+		log.Printf("proto: duplicate enum registered: %s", typeName)
+		return
 	}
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for the PR. I'm in the process of redesigning how registration works. Let's leave changes like for the broader story around #268.

namusyaka:
@dsnet Thank you for telling me that you are redesigning. I expect it.
However, is there a reason not to accept this change at the moment?

dsnet:
Log printouts get ignored all the time. The panic provided a loud failure; switching a loud failure to a silent one doesn't seem like change we should make now.

dsnet:
On the flip side, if we switch the silent ones to be loud, then people will be broken now when they weren't before. For now, let's just keep the status quo, even if it isn't ideal.

namusyaka:
I will write about the problem I am encountering.

I am currently developing an API using protobuf on appengine. As a matter of course, RegisterEnum needs to be deleted like the official appengine package.
However, in addition to that, I have also developed a CLI corresponding to the API using pb.go, and at that time it has fallen into a condition that can not be developed correctly due to the fact that protoEnum is not registered properly.
I regard this as a fatal defect rather than being not ideal.

dsnet:
Why is the duplicate registration happening in your case?

namusyaka:
This is because duplicate loading of all source code including pb.go generated by appengine and dev_appserver.py as standard is done.
I guess this issue is recognized by official appengine package developers. That's why the official package is doing the same.

dsnet:
Can you be more specific? Which proto is being loaded twice?

namusyaka:
Source code to be deployed for appengine.
For example, RegisterEnum should exist in the following file, but it has been deleted by sed.
https://github.com/golang/appengine/blob/master/internal/datastore/datastore_v3.pb.go

Is this the correct answer for your question?
If you need a minimal source code set corresponding to appengine, please let me know. I am currently in trouble because of in-house development, so I need to create a new repository for reproduction. Probably it will take a few days.

namusyaka:
ping?

PR #577, opened 2018-03-30
protoc-gen-go: revert some API changes

neild:
Introducing the GoImportPath and GoPackageName types to distingish between
these values changed the exported API of the generator package in a few
small ways. While the API of this package comes with no stability
guarantee, the benefit to the exported part of the change is small.
Revert the exported changes to avoid gratuitous breakage.

Specific changes:
  Generator.ImportPrefix reverts to type string
  Generator.ImportMap reverts to type map[string]string

(There are other externally-visible changes to the package API which
are worth making and will not be reverted, most notably the removal of the
PackageName method from FileDescriptor.)


dsnet:
Sad face.

Issue #183, opened 2016-05-11
proto3 jsonpb differs from json field names since protoc-3.0.0-beta2


mwitkow:
Recently `jsonpb` started respecting the [proto3 spec](https://developers.google.com/protocol-buffers/docs/proto3#json) for JSON names in conjunction with protoc `3.0.0 beta2` through (~Feb 2016):
- https://github.com/golang/protobuf/commit/d20896fc31fb0e955e57970d61eb6eea054ff048
- https://github.com/golang/protobuf/commit/001690d39bd620847bb265d93a7c5e1bd3737308

However, the normal `json` package still uses `OrigName` for field name. 

For example:

``` proto
syntax = "proto3";
package exampletest;

message SomeMessage {
  string snake_case = 1;
  string CamelCase = 2;
  string lowerCamelCase = 3;
  string Mixed_Stuff_With_Underscores = 4;  
}
```

produces 

``` go
type SomeMessage struct {
    SnakeCase                    string `protobuf:"bytes,1,opt,name=snake_case,json=snakeCase,proto3" json:"snake_case,omitempty"`
    CamelCase                    string `protobuf:"bytes,2,opt,name=CamelCase,json=camelCase,proto3" json:"CamelCase,omitempty"`
    LowerCamelCase               string `protobuf:"bytes,3,opt,name=lowerCamelCase,proto3" json:"lowerCamelCase,omitempty"`
    Mixed_Stuff_With_Underscores string `protobuf:"bytes,4,opt,name=Mixed_Stuff_With_Underscores,json=mixedStuffWithUnderscores,proto3" json:"Mixed_Stuff_With_Underscores,omitempty"`
}
```

This means that `jsonpb` will serialize `{"snakeCase": "foo"}` while `json` will serialize `{"snake_case": "foo"}`. 
`jsonpb` will accept what `json` serialized `{"snake_case": "foo"}` (thanks to `acceptedJSONFieldNames` method, @dsymonds props for that).
`json` will not read what `jsonpb` serialized.

I agree there is an argument to be made: always use the same deserializer that you used to serialize with... but there is no way for users to easily figure out whether this is `jsonpb` or `json` if it's stored somewhere, leading to massive confusion.  Especially since in our codebase, we use `.proto` as schemas for JSON files that we can't avoid, leading to even more confusion.

Also, this behaviour only exhibits when `protoc-3.0.0-beta-2` is used, not `beta-1`, leading to even more confusion.

Can we either:
[...]

PR #578, opened 2018-04-02
Fix jsonName in generated json tag

tectiv3:
Generator produced json tag with field name instead of jsonName:
e.g. before:
```golang
RunWhen     string `protobuf:"bytes,6,opt,name=run_when,json=runWhen" json:"run_when,omitempty"`
```
after:
```golang
RunWhen     string `protobuf:"bytes,6,opt,name=run_when,json=runWhen" json:"runWhen,omitempty"`
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1781,7 +1781,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 		ns := allocNames(base, "Get"+base)
 		fieldName, fieldGetterName := ns[0], ns[1]
 		typename, wiretype := g.GoType(message, field)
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for your PR, but unfortunately, we are no longer accepting any more changes that alter the `json` tag for several reasons:
1) This is a breaking change as it will cause the way JSON is serialized for existing users to change.
2) We consider it a mistake (see https://github.com/golang/protobuf/issues/183#issuecomment-353693510)

cupen:
Hello @dsnet @tectiv3.  Maybe we can make this as a non-breaking change, `--tagstyle` or `option go_tagstyle="normal"`?

PR #579, opened 2018-04-02
protoc-gen-go: add convenience function for getting ServiceDescriptorProtos

dfawley:
Like #516, but in the main go code generator instead of the grpc plugin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ go:
 - 1.x
 
 install:
-  - go get -v -d -t github.com/golang/protobuf/... google.golang.org/grpc
+  - go get -v -d -t github.com/golang/protobuf/...
   - curl -L https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip -o /tmp/protoc.zip
   - unzip /tmp/protoc.zip -d "$HOME"/protoc
   - mkdir -p "$HOME"/src && ln -s "$HOME"/protoc "$HOME"/src/protobuf
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ go:
 - 1.x
 
 install:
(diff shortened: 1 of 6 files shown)


dfawley:
Sorry for the noise.  I think I see what's happening now.  Because this is a dev branch of protobuf, `go get` doesn't work with it (it attempts to get from the master branch).  Not sure what the best way to deal with this is, but I'll keep trying...

dfawley:
The last error seems to be a context problem.  I'll have to look, but it looks like the grpc plugin on dev is (and was before this PR) importing "context" instead of "x/net/context".  This would be a problem, if true.

dsnet:
I've become increasingly convinced that support for accessing `ServiceDescriptorProto` and `MethodDescriptorProto` should be in the mainline generator. However, I'm not sure this is the right API for this. It should be thought of as a whole with regard to how all other descriptors are retrieved. See https://github.com/golang/protobuf/issues/489#issuecomment-378026018.

dfawley:
@dsnet, that sounds fine to me.

I'd still like to salvage some sort of tests based on this, in light of the current incompatibility of the grpc plugin with Go 1.6 (please see #580).

dsnet:
Given that https://github.com/golang/protobuf/issues/364#issuecomment-384123374 has a more comprehensive plan for adding this functionality, what would you like to salvage from this current PR to add today? The tests you have seem to depend on the functionality that I prefer not adding at the present moment.

dfawley:
@dsnet, a simple "nop" test that imports a `.pb.go` file generated by the grpc plugin would be useful.  Unless I'm missing something, I don't think we have any tests that ensure the generated code will even compile.

dsnet:
Done. Are you okay with closing this PR as stale?

dfawley:
SGTM, thanks.

PR #585, opened 2018-04-09
Remove unused function

tamird:
@dsnet

### DIFF
--- a/proto/text.go
+++ b/proto/text.go
@@ -456,24 +456,6 @@ func (tm *TextMarshaler) writeStruct(w *textWriter, sv reflect.Value) error {
 	return nil
 }
 
-// writeRaw writes an uninterpreted raw message.
-func writeRaw(w *textWriter, b []byte) error {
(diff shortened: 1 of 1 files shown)


proto/message_set.go
```diff
@@ -163,14 +163,8 @@ func marshalMessageSet(exts interface{}, deterministic bool) ([]byte, error) {
...
-		ie := XXX_InternalExtensions{
-			p: &struct {
-				mu           sync.Mutex
-				extensionMap map[int32]Extension
-			}{
-				extensionMap: exts,
-			},
-		}
+		var ie XXX_InternalExtensions
+		ie.p.extensionMap = exts
```
neild: This dereferences ie.p, which is nil.

tamird: Good point, removed that commit.

dsnet:
Ooops, I must have missed this and didn't put it in the internal "Bytes" method patch hack.

LGTM, I'll submit this when I update the internal patch.

PR #596, opened 2018-05-04
jsonpb.Marshaler support for json="-" tag

mcmurdont:
If a pb.go file has a `json="-"` adding by a post-processing tool, jsonpb.Marshaler does not check the tag, unlike "encoding/json".


dsnet:
We can't accept this change. We're trying to phase away from dependence on `json` tags in the `jsonpb` implementation and rely purely on upcoming proto reflection support. Also, we have a policy that we don't accept features that allow users to violate protobuf correctness. Since the protobuf<->JSON spec provides no options of this sort, we can't accept that here.

Issue #598, opened 2018-05-10
protoc-gen-go --version


stevenroose:
For a lot of projects, especially open-source ones, you want all contributors to use the same version of all the protobuf (and grpc) tools. Otherwise pull requests for small changes in proto files could potentially have huge changes in generated code if the contributor has a different version.

A good solution for this, is to provide a generation script within the project that contributors should use to generate the code. In this script, the version of the tools can be checked.

Currently, `protoc --version` gives you the version of protoc. It would be nice if `protoc-gen-go` could also somehow expose its version through a command. Preferably the same as protoc.

dsnet:
Closing as duplicate of #524.

Issue #524, opened 2018-02-21
protoc-gen-go: emit versioning information


dsnet:
The generator currently emits the following:
```
// Code generated by protoc-gen-go. DO NOT EDIT.
// source: extension_user/extension_user.proto
```

We should improve this by adding information about the versions:
1) What `go` toolchain was used (e.g., `go1.9.6`)
2) What version of `protoc` was used (e.g., `libprotoc 3.5.1`)
3) What version of `protoc-gen-go` was used

The 3rd item is harder to do, but the upcoming version control features in the Go toolchain will help with this. My understanding is that one of the features it will provide is a cryptographic signing of all the Go packages that went into building that binary.

For now, I would be happy to just see items 1 and 2 start being encoded in the generated infolog.

dsnet:
Furthermore, any special options passed to protoc-gen-go should probably be provided in that log. For example, whether the `grpc` plugin is used.

dsnet:
Related: #384

neild:
If we do this, we should also make "make regenerate" produce an error when run with the "wrong" (non-canonical) version of the toolchain.

dsnet:
> If we do this, we should also make "make regenerate" produce an error when run with the "wrong" (non-canonical) version of the toolchain.

That sounds fine to me. However, I don't think that's this issue, but what @tamird is trying address in #521.

dsnet:
If it wasn't clear, this issue should address both generated code and also provide a command-line flag to check this information.

dsnet:
This is blocked on https://github.com/golang/go/issues/26404

zeeraw:
What is the status on the command line flag?

dsnet:
Still blocked on golang/go#26404, which is slated for release in Go1.12.

metasyn:
Sounds like it is unblocked now, with go 1.12 released & golang/go#24604 closed?

rocco-haro:
Any updates on this?

saravana-thiyagaraj:
Any updates on this?

thsaravana:
@dsnet Any update on this?

RahulMitra:
@dsnet  any update on this?

dsnet:
This is already resolved in the v2 branch. The v2 release is targeting end of Q1.

dsnet:
This is fixed in the [v1.20.0 release](https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.20.0) of `protoc-gen-go`.

PR #599, opened 2018-05-10
protoc-gen-go: Add -version flag to print version

stevenroose:
Fixes https://github.com/golang/protobuf/issues/598 and partially https://github.com/golang/protobuf/issues/524.

### DIFF
--- a/protoc-gen-go/main.go
+++ b/protoc-gen-go/main.go
@@ -49,6 +49,7 @@
 package main
 
 import (
+	"flag"
 	"io/ioutil"
 	"os"
 
@@ -56,7 +57,20 @@ import (
 	"github.com/golang/protobuf/protoc-gen-go/generator"
 )
 
+const (
+	// versionString holds the current version of protoc-gen-go.
+	versionString = "1.1.0"
+)
+
 func main() {
+	// If the -version flag is set, just print the version and exit.
+	versionFlag := flag.Bool("version", false, "print the version of protoc-gen-go")
+	flag.Parse()
+	if *versionFlag {
+		os.Stdout.WriteString(versionString)
+		os.Exit(0)
+	}
+
 	// Begin by allocating a generator. The request and response structures are stored there
 	// so we can do error handling easily - the response structure contains the field to
 	// report failure.


dsnet:
Unfortunately, v1.1.0 has already been tagged, so we can't retroactively add this functionality without tagging another minor version (for which this is not worth a minor release on its own).

dsnet:
Also, I'd rather not hard-code a version string as it is easy for that to get out of sync with the repository tag. Instead, I rather wait for proper versioning support to come with golang/go#24301. It's not just the version of `protoc-gen-go` you want to encode, but also all of its dependencies (namely the standard library), as that also affects the output.

stevenroose:
> Unfortunately, v1.1.0 has already been tagged, so we can't retroactively add this functionality without tagging another minor version (for which this is not worth a minor release on its own).

Of course. I just added the latest version, so that it could be updated whenever a new release was made.

> Also, I'd rather not hard-code a version string as it is easy for that to get out of sync with the repository tag. Instead, I rather wait for proper versioning support to come with golang/go#24301. It's not just the version of protoc-gen-go you want to encode, but also all of its dependencies (namely the standard library), as that also affects the output.

I agree that an automatic way to include the version into the code would be safer. However I didn't know of any. Waiting for vgo might make sense, but could take more time.

dsnet:
> Waiting for vgo might make sense, but could take more time.

It's certainly the correct path forward to fixing this. Russ is aiming to get vgo support into Go1.11, which is 3 months out. I'd rather wait for that.

Is there something that absolutely requires versioning info now? Or is this just a "nice to have"?

stevenroose:
It's just cumbersome that in open-source project it's virtually impossible
to make changes to proto files and commit the respective changes to the
generated code, since a single update in any of the several used packages
(protoc, protoc-gen-go, protoc-gen-grpc-gateway, protoc-gen-swagger) will
change things that shouldn't be changed. So basically a PR to a proto file
has to be followed by a commit from the maintainer that has a consistent
setup.

dsnet:
Closing this in favor of vgo support, which is moving along.

stevenroose:
vgo doesn't really solve this. This is about external tools being able to determine the version of the executable.

mvdan:
Maybe this should wait until the upcoming `go release` command. It sounds to me like exposing version information for executables is something it would do.

stevenroose:
I'm ok with waiting until a good approach is figured out. But if it's closed, I feel like it'll be forgotten about soon :)

mvdan:
Agreed. I've been struggling with the same problem in some projects of mine, and have temporarily added hard-coded version constants in a couple. I started watching this thread in the hopes that a better idea would come up :)

dsnet:
I do care about having good versioning strings in protoc-gen-go, but I'm not fond of doing extra plumbing work for something that will probably be solved in the coming months. There are many other protobuf issues to address that is probably better to spend the limited time on.

stevenroose:
I totally agree. But again, there's no harm in this issue being open and low priority.

dsnet:
Perhaps, I'm misunderstanding something, but there is an open issue for this? #524

stevenroose:
That issue is about logging the version used in the generated code.
> For now, I would be happy to just see items 1 and 2 start being encoded in the generated infolog.

That's different than having the executable log its version (like almost all GNU executables do, fwiw) so that tooling can check this version *before* generating new code.

dsnet:
That issue is a superset of the functionality requested here. I still want #3 in that list (which is why that issue is still open). Once #524 is achievable, it is trivial to expose a command-line flag that exposes this information. I think we're spending time debating the same thing.

PR #600, opened 2018-05-11
grpc: document ctx closure best practices

jadekler:
Corresponding grpc PR: https://github.com/grpc/grpc-go/pull/2060

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -162,7 +162,7 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	g.P()
 	g.P(fmt.Sprintf(`// Client API for %s service.
 //
-// For semantics around ctx use and closing/ending RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.`, servName))
+// For semantics around ctx use and closing/ending streaming RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.`, servName))
 
 	// Client interface.
 	if deprecated {
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -160,8 +160,9 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	deprecated := service.GetOptions().GetDeprecated()
 
 	g.P()
-	g.P("// Client API for ", servName, " service")
-	g.P()
+	g.P(fmt.Sprintf(`// Client API for %s service.
+//
+// For semantics around ctx use and closing/ending RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.`, servName))
 
 	// Client interface.
 	if deprecated {
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -160,13 +160,13 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	deprecated := service.GetOptions().GetDeprecated()
 
 	g.P()
-	g.P(fmt.Sprintf(`// Client API for %s service.
+	g.P(fmt.Sprintf(`// %sClient is the client API for %s service.
 //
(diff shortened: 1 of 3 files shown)


jadekler:
@dfawley

protoc-gen-go/grpc/grpc.go
```diff
@@ -303,6 +303,19 @@ func (g *grpc) generateClientMethod(servName, fullServName, serviceDescVar strin
 	if method.GetOptions().GetDeprecated() {
 		g.P(deprecationComment)
 	}
+	streamType := unexport(servName) + methName + "Client"
+	g.P(fmt.Sprintf(`// The ctx passed in will be used for the lifetime of this object.
```
dfawley: This is not a godoc-compliant comment.
```go
// <method> initiates a new streaming RPC to the server, yadda yadda.
```
However, the type that owns this method is not exported (`unexport(servName)` below).  This should be documented in the interface instead, or else it will never be seen.  Note that those include the comments from the proto file, so we'll have to be smart about how to keep pulling those in but also document the Go implementation as well.

jadekler:
@dfawley PTAL - moved to interface, use pointer instead of dupe doc

protoc-gen-go/grpc/grpc.go
```diff
@@ -160,8 +160,9 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	deprecated := service.GetOptions().GetDeprecated()
 
 	g.P()
-	g.P("// Client API for ", servName, " service")
-	g.P()
+	g.P(fmt.Sprintf(`// Client API for %s service.
+//
+// For semantics around ctx use and closing/ending RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.`, servName))
```
dfawley: Nit: "closing/ending *streaming* RPCs".

jadekler: Done

jadekler:
@dsnet Could you describe what the problem with travis is? It seems to be stuck on some commit? Should I rebase master?

protoc-gen-go/grpc/grpc.go
```diff
@@ -160,8 +160,9 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 	deprecated := service.GetOptions().GetDeprecated()
 
 	g.P()
-	g.P("// Client API for ", servName, " service")
-	g.P()
+	g.P(fmt.Sprintf(`// Client API for %s service.
```
dsnet: According to https://golang.org/doc/effective_go.html#commentary, this should probably be written as:

```go
// TestClient is the client API for the Test service.
//
// For semantics around resource cleanup, please refer to
// https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.
```

dsnet: Also, shouldn't a similar comment be on the server equivalent of this?

protoc-gen-go/grpc/grpc.go
```diff
@@ -160,8 +160,9 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
...
+//
+// For semantics around ctx use and closing/ending streaming RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.`, servName))
 
 	// Client interface.
 	if deprecated {
```
dsnet: We should put a newline before the deprecation comment:
```go
g.P("\n"+deprecationComment)
```

jadekler:
Aha - thanks, done.

protoc-gen-go/grpc/grpc.go
jadekler: Done

protoc-gen-go/grpc/grpc.go
jadekler: Done

And, although I think you're right, the server's case + recommendations are different and should probably live in a different PR.

dsnet: That's fine with me.

dsnet:
Changes LGTM, but you'll need to regenerate the protos again.

jadekler:
Shoot, good point. Done.

PR #604, opened 2018-05-14
grpc: fix and improve interface comments for client/server types

dsymonds:
32a84b2 added a newline that resulted in the interface doc comment
becoming detached from the interface type definition. However, both it
and the corresponding server comment were inconsistent, so this change
makes them all consistent and well-rendered in godoc.


dsnet:
Thank you for the fix.

PR #612, opened 2018-05-18
jsonpb: fix unmarshaling JSON object with escaped string into Struct type

### LINKED ISSUE #605: jsonpb: Attempting to unmarshal escaped JSON into a protobuf throws an error
In a project of mine, I am fetching JSON from a third-party API and then attempting to unmarshal the response into a protobuf struct, defined with the type `google.protobuf.Struct`. The API response has escaped values in it and is resulting in the following error: `bad value in StructValue for key "image": unrecognized type for Value "\"https:\\/\\/ei.isnooth.com\\/multimedia\\/0\\/2\\/8\\/image_787698_square.jpeg\""`.

As a workaround I have had to use the regular `encoding/json` library to unmarshal and then marshal the response again, before passing the data on to `jsonpb.Unmarshal` with a protobuf.

Further information and relevant code can be found on my posts here:
- [Stack Overflow](https://stackoverflow.com/questions/50314476/why-does-unmarshalling-this-api-response-return-an-unexpected-eof?noredirect=1#comment87649017_50314476)
- [golang-nuts Google Group](https://groups.google.com/forum/#!topic/golang-nuts/fNVZI0mCEkA)

cybrcodr:
Fix unmarshaling JSON object with escaped string into Struct type.

Fixes #605

### DIFF
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -693,11 +693,11 @@ var unmarshalingTests = []struct {
 
 	{"Duration", Unmarshaler{}, `{"dur":"3.000s"}`, &pb.KnownTypes{Dur: &durpb.Duration{Seconds: 3}}},
 	{"Duration", Unmarshaler{}, `{"dur":"4s"}`, &pb.KnownTypes{Dur: &durpb.Duration{Seconds: 4}}},
-	{"Duration with unicode", Unmarshaler{}, "{\"dur\":\"3\\u0073\"}", &pb.KnownTypes{Dur: &durpb.Duration{Seconds: 3}}},
+	{"Duration with unicode", Unmarshaler{}, `{"dur": "3\u0073"}`, &pb.KnownTypes{Dur: &durpb.Duration{Seconds: 3}}},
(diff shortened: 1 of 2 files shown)


jsonpb/jsonpb_test.go
```diff
@@ -752,6 +754,13 @@ var unmarshalingTests = []struct {
 	{"UInt32Value", Unmarshaler{}, `{"u32":4}`, &pb.KnownTypes{U32: &wpb.UInt32Value{Value: 4}}},
 	{"BoolValue", Unmarshaler{}, `{"bool":true}`, &pb.KnownTypes{Bool: &wpb.BoolValue{Value: true}}},
 	{"StringValue", Unmarshaler{}, `{"str":"plush"}`, &pb.KnownTypes{Str: &wpb.StringValue{Value: "plush"}}},
+	{"StructValue containing StringValue's", Unmarshaler{}, "{\"escaped\": \"a\\/b\", \"unicode\": \"\\u00004E16\\u0000754C\"}",
+		&stpb.Struct{
+			Fields: map[string]*stpb.Value{
+				"escaped": {Kind: &stpb.Value_StringValue{"a/b"}},
```
cybrcodr: This testcase exposes issue #605.

Note that the other testcases that I added here are somehow working even before adding the fix in.

PR #627, opened 2018-06-05
jsonpb: error on scalar enum provided for repeated enums instead of panic

eleniums:
Ran into an issue where attempting to unmarshal valid json that contains a field with a single enum into a message that has a repeated enum field causes a panic. For example, try to unmarshal this:
```go
{
    "rFunny":"PUNS"
}
```
into a message that looks like this:
```go
type Message struct {
    RFunny []Message_Humour
}
```
This will cause a panic. I know the json is an invalid representation of the Message struct, but it shouldn't panic. I'm attempting to fix this issue by validating that the targetType is correct before setting the enum.

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -878,9 +878,10 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			target.Set(reflect.New(targetType.Elem()))
 			target = target.Elem()
 		}
-		if targetType.Kind() == reflect.Int32 {
-			target.SetInt(int64(n))
+		if targetType.Kind() != reflect.Int32 {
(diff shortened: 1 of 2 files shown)


jsonpb/jsonpb.go
```diff
@@ -878,7 +878,9 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			target.Set(reflect.New(targetType.Elem()))
 			target = target.Elem()
 		}
-		target.SetInt(int64(n))
+		if targetType.Kind() == reflect.Int32 {
+			target.SetInt(int64(n))
+		}
```
dsnet: The conditional is correct, but the behavior is still off. This simply ignores the invalid input instead of returning an error.

eleniums:
I wasn't sure whether jsonpb typically ignored errors like that or not. I just pushed a commit to instead return an error in this scenario.

PR #628, opened 2018-06-05
proto: revert UTF-8 validation for proto2

### LINKED ISSUE #622: Revert proto2 string UTF-8 validation
#484 introduced UTF-8 validation for string fields into proto2. This has resulted in a number of problems with App Engine (https://github.com/golang/appengine/issues/136, https://github.com/golang/appengine/issues/140, https://github.com/golang/appengine/issues/143) which do not have a workaround other than pinning to an outdated version of proto2.

The introduction of strict UTF-8 validation for string fields in proto2 with no opt-out has broken a critical flow for App Engine users using App Engine datastore. We have always allowed serializing `[]byte` into the payload of a datastore PUT request, and with the latest proto2 validation changes, users will be unable to retrieve any blobs previously stored in datastore and unable to store new blobs.

Strict UTF-8 validation of proto2 string fields may very well be worthwhile, but this change leaves our users without a migration path because datastore lacks the capability to receive blobs as anything other than a serialized string, and adding that feature is not on anyone's timeline.

I have opened this as an issue internally at Google, but I'm duplicating it here for additional input from the community.

dsnet:
The proto specification officially says that proto2 and proto3 strings should be
validated, but pragmatically, compliance with the spec has been poor.
For example, the Go implementation did not validate either and added strict
validation recently to be compliant. However, this caused signficant breakage.

Cases of breakage should change the proto field type from string to the bytes type.
However, this is not always possible, when the field is part of the exposed API.
This tends to be the case for proto2, where some other notable language
implementations (like C++) do not validate proto2 for valid UTF-8.
However, since most language implementations do validate for UTF-8 in proto3,
we keep that behavior.

Making this change for Go is a little tricky since each field does not necessarily
know whether it is operating under the proto2 or proto3 syntax. Thus, we modify
the generator to emit a "utf8" struct field tag for any string fields in proto3.
The implications of this change is that people will need to regenerate their
proto files to have UTF-8 validation.

For completeness, the proto package checks for the utf8 tag in all cases even
if the field is obviously proto2 (i.e., is of the *string type).
However, the generator does not emit the utf8 tag for proto2, so this works fine.
We add support in the proto package because it is more consistent and also
for the event that we ever try to add validation to proto2 again.

We expand UTF-8 validation tests to ensure this works for the cross-product of
(proto2, proto3) and (scalar, vector, oneof, and maps) fields with strings.

Fixes #622


protoc-gen-go/generator/generator.go
```diff
@@ -1573,7 +1573,9 @@ func (g *Generator) goTag(message *Descriptor, field *descriptor.FieldDescriptor
 		if *field.Type == descriptor.FieldDescriptorProto_TYPE_BYTES {
 			name += ",proto3"
 		}
-
+		if *field.Type == descriptor.FieldDescriptorProto_TYPE_STRING {
+			name += ",utf8"
```
neild: Should this use the "proto3" tag instead? We may discover other cases in the future where proto2/proto3 fields need different handling, and it'll be more efficient to have a single tag in that case.

If so, should we preemptively tag all proto3 fields, even if they currently have no semantic difference?

dsnet: The `proto3` tag was already added to distinguish the semantic difference for `[]byte{}` vs `[]byte(nil)` when it comes to the bytes type.

Looking through the code history, the attribute was only generated on the bytes type out of fear for binary size increase. If we want to go with `proto3`, should we just generate it for all fields? That's an increase of 7 bytes for every field. Alternatively, we can only generate it for the `string` field.

dsnet:
Let's avoid exporting the error for now.

Several thoughts:
* As it stands the current logic fails-fast when there is invalid UTF-8. In order for a distinguishable error to be useful, they would still need to fully marshal or unmarshal the message.
* We probably wouldn't want a sentinel error. A useful piece of information is the exact field that had UTF-8 error.
* If we exported a new type, then we have to think carefully about what happens message has both required fields not set and invalid UTF-8. You would probably want to report that both errors occurred. Here's a case where your error tagging idea would be really useful.

protoc-gen-go/generator/generator.go
neild: If we're worried about binary size increase that argues for a proto3 tag rather than utf8, to avoid the possibility of redundant tags in the future.

I don't really have a strong opinion on every field vs. string, although there are a couple arguments for the former: It avoids any possibility of needing to do this again in the future, and I think all the cases where 7 bytes per field might conceivably be an issue are proto2 anyway.

dsnet: I enabled it unilaterally for all fields.

PR #630, opened 2018-06-08
jsonpb: decode int32/uint32/float32/float64 strings

kassiansun:
According to
https://developers.google.com/protocol-buffers/docs/proto3#json,
all integers & floats are accepted as strings.

We still preserve the encoding behavior, to parse int64s to strings and other number to number.


kassiansun:
I signed it!

dsnet:
Looks good. Thank you.

Issue #648, opened 2018-06-29
deterministic marshaling for non regenerated code


awalterschulze:
It seems that the Marshal method is called when deterministic is true or false.
https://github.com/golang/protobuf/blob/master/proto/table_marshal.go#L227-L232

This might be a non deterministic generated Marshal method in a library used by a user pulling in the newer version of golang/protobuf, where they are able to set deterministic to true.

Even if we modify this code to rather be

```go
if u.hasmarshaler && !deterministic {
		...
}
```

We also have to worry about
https://github.com/golang/protobuf/blob/master/proto/table_marshal.go#L310-L314
Which means that if there is a Marshaler the size cache is not set, but the non generated marshal function assumes that it is set.

How do we solve this?

dsnet:
The unfortunate reality is that we can't reconcile this issue in the present API. If deterministic marshaling is enabled, it will just have to be a best effort as it is impossible to know whether custom implementations are deterministic or not.

The design of the [`protoiface` API](https://docs.google.com/document/d/19kfhro7-CnBdFqFk7l4_HmwaH2JT_Rhw5-2FLWLEGGk/edit#heading=h.o9txp09ylzju) is intended to enable to safer way for custom implementations to coexist.

awalterschulze:
@dsnet ^^ ?

dsnet:
Strict erroring for custom unmarshalers sounds good to me. Would you like to submit a PR?

awalterschulze:
That is great to hear :)

Just checking you mean marshalers and not unmarshalers, right?

dsnet:
Correct.

awalterschulze:
@jmarais is working on a pull request.

Thank you so much :)

PR #650, opened 2018-07-10
strict erroring for custom marshalers trying to use a deterministic b…

jmarais:
…uffer as discussed in #648


jmarais:
I signed it!

jmarais:
I am not sure why the following would fail in travis. Any recommendation?

```
go build ./protoc-gen-go/testdata/grpc/grpc.pb.go
# golang.org/x/net/http2
../../../golang.org/x/net/http2/go111.go:10:30: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
../../../golang.org/x/net/http2/go111.go:14:26: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
../../../golang.org/x/net/http2/go111.go:15:8: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
```

proto/table_marshal.go
```diff
@@ -225,6 +225,9 @@ func (u *marshalInfo) marshal(b []byte, ptr pointer, deterministic bool) ([]byte
 	// If the message can marshal itself, let it do it, for compatibility.
 	// NOTE: This is not efficient.
 	if u.hasmarshaler {
+		if deterministic {
+			return nil, errors.New("proto: deterministic not supported by the Marshal method of " + u.typ.Name())
```
dsnet: s/Name/String/

`reflect.Type.Name` only prints the type name, while `reflect.Type.String` provides the package as well.

dsnet:
You can ignore the travis failure. After some digging, it seems that `WroteHeaderField` was added on June 27th in https://go-review.googlesource.com/c/go/+/67430. However, travis is running with go1.11beta1, which was cut on June 26th, so lacks that function.

proto/table_marshal.go
jmarais: Thanks for the suggestion.

dsnet:
Thanks for the change!

PR #654, opened 2018-07-13
jsonpb: avoid copying string-valued map-keys

dhowden:
Encoding large maps with string keys is very expensive.  Profiling indicated that the culprit was `github.com/golang/protobuf/jsonpb.mapKeys.Less` which is used to ensure that the ordering of the map keys are consistent since Go map iteration is random.

By adding a string-specific case we avoid the overhead of going through the fmt.Sprintf calls, which removes a lot of unnecessary allocations.

I suspect that this can be taken further - the requirement of ordered keys does not apply to the JSON rep, so I believe it could be be removed in this case (https://developers.google.com/protocol-buffers/docs/proto3#json).  Note: the "text format" rep does require ordering, so maybe this is why the JSON code inherited it...

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -1116,6 +1116,8 @@ func (s mapKeys) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 func (s mapKeys) Less(i, j int) bool {
 	if k := s[i].Kind(); k == s[j].Kind() {
 		switch k {
+		case reflect.String:
+			return s[i].String() < s[j].String()
 		case reflect.Int32, reflect.Int64:
 			return s[i].Int() < s[j].Int()
 		case reflect.Uint32, reflect.Uint64:


puellanivis:
:+1:

dhowden:
Build is passing for go versions `1.10.x` and `1.6.x`,  but failing on `1.x` due to:
```
go build ./protoc-gen-go/testdata/grpc/grpc.pb.go
# golang.org/x/net/http2
../../../golang.org/x/net/http2/go111.go:12:30: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
../../../golang.org/x/net/http2/go111.go:16:26: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
../../../golang.org/x/net/http2/go111.go:17:8: trace.WroteHeaderField undefined (type *clientTrace has no field or method WroteHeaderField)
make: *** [test] Error 2
```
https://travis-ci.org/golang/protobuf/jobs/403362012#L634

puellanivis:
This is a known break in the travis CI, and not at all the result of this patch.

PR #656, opened 2018-07-20
proto: fix and cleanup test for deterministic marshal with custom marshalers

dsnet:
PR #650 added a check to error out when a custom marshaler was called and
deterministic marshaling was also specified.

That change performed the check in a relatively obscure location that did not
catch all code paths. Since determinism can only be enabled on the Buffer type,
we check it up front in Buffer.Marshal.

Also, change the test to avoid code injection into generated sources.


dsnet:
Done.

PR #657, opened 2018-07-24
Refactor protoc-gen-go generator - split up generateMessage

bufflig:
Split actual text generation passes of generateMessage out by
first building a data structure suitable for code generation
and then generate the output for each field in separate methods.

Partially incorrect and unneccessary usage of g.In() and g.Out()
has als been removed throughout the code generation as the output
is properly formatted at the end of code generation anyway.

The functionality is unchanged except for the order of oneof
methods in the output file.


bufflig:
I've now followed the instructions for someone working on the Go team at google. As far as I understand it, the CLA check should now clear.

bufflig:
The generated json differs, will try again when I've fixed that.

bufflig:
With the proper protoc, I could update the golden files without destroying the json names, trying again...

PR #660, opened 2018-07-27
proto: make invalid UTF-8 errors non-fatal

dsnet:
The current logic currently treats RequredNotSetError as a non-fatal
error such that it continues to proceed with its normal execution, but
returns that non-fatal error at the end.

We now make it such that invalid UTF-8 is also a distinguishable error
that is treated as non-fatal by the execution logic. In the rare event
that both an RequiredNotSet error and InvalidUTF8 error is encountered,
the first one encountered is returned.

In the process of making this change, we also fix a number of cases where
RequiredNotSet was treated as fatal, when it should not have been non-fatal
(notably in the logic for extensions, message sets, and maps).

This change deliberately does not provide an API make it easy to distinguish
invalid UTF-8 errors as this is not a normal behavior we want users to
depend on. We can always expose additional API for this in the future.

Users can test for invalid UTF-8 with:
	re, ok := err.(interface{ InvalidUTF8() bool })
	isInvalidUTF8 := ok && re.InvalidUTF8()

### DIFF
--- a/proto/all_test.go
+++ b/proto/all_test.go
@@ -2256,26 +2256,32 @@ func TestInvalidUTF8(t *testing.T) {
 		label  string
 		proto2 Message
 		proto3 Message
+		want   []byte
 	}{{
 		label:  "Scalar",
 		proto2: &TestUTF8{Scalar: String(invalidUTF8)},
 		proto3: &pb3.TestUTF8{Scalar: invalidUTF8},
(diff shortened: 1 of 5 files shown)


dsnet:
\cc @spenczar

puellanivis:
Just as a note, the example test for invalid utf-8 combines an assignment with a logical operator. I’m sure it’s just a typo.

dsnet:
@puellanivis, I'm not sure I follow, which line are you referring to?

Issue #52, opened 2015-07-19
Custom tags to generated structs


jedborovik:
Is it possible to define a `message` with custom tags? For example defining a json name that isn't just the lowercase field name or adding a bson tag.

dsymonds:
No, there's no support for that, nor any intention to support that.

mwmahlberg:
I'd like to add that this feature would be extremely useful for validation purposes.

jswierad:
+1

Use case: Using protnbuf generated structs along with sqlx package. Would be awesome

mikebell-org:
Likewise the go library for cloud datastore. Basically, increasing numbers of things make use of struct tags, and without support for them you're left manually duplicating structs? If there are no plans to add this functionality, what is the recommended way to handle it?

AlexChen-misfit:
+1

chai2010:
+10086

iain17:
I would love this feature!

Sef1995:
I would like to have this feature aswell!

Sef1995:
I would love this feature aswell!

willks:
This would be handy, as I need to validate JSON payloads for my REST API portion of the service.

awalterschulze:
Since golang/protobuf will probably never support this.  Here are some solutions.

https://github.com/mwitkow/go-proto-validators is useful for validation on proto structs.

https://github.com/gogo/protobuf can be used in two ways to have custom tags:

1) You can use the moretags extension
https://github.com/gogo/protobuf/blob/master/test/tags/tags.proto

2) Or you can use the typedecl extension if you don't want to generate the golang struct.
This allows you declare your own golang struct with all the tags you want and more.

ekiyanov:
Need it as well to automate insert/update into the sql database

qianlnk:
I've add a plugin retag for grpc.

https://github.com/qianlnk/protobuf/tree/master/protoc-gen-go/retag

RajeshKumar1990:
Thanks @qianlnk. This really helped me.But i want to use timestamp in my proto file.When i try to use it like this 
protoc -I src/gatewayapi/rpcdata/rpcdata --proto_path=src/github.com/google/protobuf/src/ --go_out=plugins=grpc+retag:. src/gatewayapi/rpcdata/rpcdata/api.proto
I m not getting the changes in the tag.May i know any way to achieve it.
Thanks

RajeshKumar1990:
[...]

PR #665, opened 2018-08-01
Add tag flag to export message tag comments

lvht:
Some golang packages use struct tag to store infomation. For example, using [go-playground/validator](https://godoc.org/gopkg.in/go-playground/validator.v9), you can define a validate rule like this

```golang
type Test struct {
	Field `validate:"max=10,min=1"`
}
```

It will be great convenient if struct generated by protoc-gen-go according to proto has these tags.

This PR promotes a simple way to achieve this goal.

First, we introduce a so called **struct tag comment** like
```proto
message User {
    int32 uid = 1 // `validator:"gt=0"`
}
```

When we generate the *.pb.go, we enable the tag flag by `--go_out=tag=true:path`, wewill get 
```golang
type User struct {
    Uid int32 `protobuf:"varint,1,opt,name=uid,proto3" json:"mid,omitempty validator:gt=0"
}
```

So other packages could read the right struct tag.

----

There are some other solution like [gogo](https://github.com/gogo/protobuf) using the protobuf extension. But this PR is a more lightweight one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ parameter list separated from the output directory by a colon:
 - `Mfoo/bar.proto=quux/shme` - declares that foo/bar.proto is
   associated with Go package quux/shme.  This is subject to the
   import_prefix parameter.
-- `tag=true` - specifies wheter export tag comment.
+- `tag=true` - specifies whether export tag comment.
 
 The following parameters are deprecated and should not be used:
 
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ parameter list separated from the output directory by a colon:
 - `Mfoo/bar.proto=quux/shme` - declares that foo/bar.proto is
   associated with Go package quux/shme.  This is subject to the
   import_prefix parameter.
+- `tag=true` - specifies wheter export tag comment.
 
 The following parameters are deprecated and should not be used:
 
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -50,6 +50,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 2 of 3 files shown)


meling:
Issue #52 discuss various solutions that are related to this. Personally, I’m not in favor of adding validation tags to message fields, for reasons explained in one of my comments on #52. Especially not as comments in the proto file. I’m not sure what is the status of @dsnet’s proposal for `go_tags`, but seems like it has been put on hold, while the protobuf team resolves other issues.

lvht:
@meling Thank you  for reply. This PR is not about the validation tags but generic tags. If it was accepted, any user defined tags in trailing comment can be added to the message struct tags. And it is the user's code's business to use these tags. And these does not have any meaning to the `protoc-gen-go`. And they are just comments in proto, there is no work to do for any util about proto.

dsnet:
Hi, thanks for the PR, but we cannot accept this change. As @meling mentioned, the proposal to add this is #52. However, we are busy working on proto reflection and are not considering various generator extensions at this time.

If we add #52, it would have to be a proto option similar to what gogo/protobuf does. A magical comment that has special meaning in Go does not scale well with other languages and proto is intended to be a cross-language serialization technology that is mostly agnostic towards the target language.

lvht:
@meling @dsnet 
> A magical comment that has special meaning in Go does not scale well with other languages

[golang/protobuf] is golang only. And the tag comment is golang only. And the tag comment is **just comment**, so other utils can not be confused.

dsnet:
A comment is intended for human consumption and humans should be free to edit them as they please without accidentally breaking semantics. Machine interpretation of comments is a definite no-go. Proto options were invented to avoid hackery around comments.

lvht:
@dsnet thank you

Issue #564, opened 2018-03-20
README.md, proto: remove usage of group from example


wellsjo:
From Google's documentation:
https://developers.google.com/protocol-buffers/docs/proto#groups
```
Groups
Note that this feature is deprecated and should not be used when creating new message types – use nested message types instead.
```

dsnet:
Where are you seeing groups being advertised?

wellsjo:
In the README, right after
> Consider file test.proto, containing

There is a code example that uses a group

dsnet:
Ah. Got it. I was searching for "groups" instead and didn't see that.

tmcrain:
I'll take a look!

PR #673, opened 2018-08-04
README.md, proto: Replaced usage of group in example 

tmcrain:
Fixes: #564

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,9 +161,10 @@ Consider file test.proto, containing
 	  required string label = 1;
 	  optional int32 type = 2 [default=77];
 	  repeated int64 reps = 3;
-	  optional group OptionalGroup = 4 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,9 +161,10 @@ Consider file test.proto, containing
 	  required string label = 1;
 	  optional int32 type = 2 [default=77];
 	  repeated int64 reps = 3;
-	  optional group OptionalGroup = 4 {
-	    required string RequiredField = 5;
+	  message OptionalNest {
```
dsnet: This is not valid proto syntax. You can't inline a message like this.

I would just remove the entire reference to groups.

tmcrain: You're right, made a dumb type there. Will just remove the entire reference.

Issue #682, opened 2018-08-20
Need an option to prevent add ",omit_empty" for json label


sparkthu:
When the json is used to communicate with browser, all field with omit_empty will cause lots of missing for keys.

puellanivis:
Was not sure about how this aligned with the standards, so I looked it up:

> An implementation may provide options to emit fields with default values in the JSON-encoded output.

https://developers.google.com/protocol-buffers/docs/proto3#json

It appears to be a valid option to do. But maybe the option name could be narrowed to be clear that it is json-specific: `json_emit_default_values`.

dsnet:
The generation of the `json` tag predates the finalization of the JSON<->Protobuf specification. When that spec was finalized, it became clear that the `json` tag alone could not be used to implement the specification (unless the `encoding/json` provides some advanced features). In retrospect, I wish we never generated the `json` tags.

That being said, adding a command-line option to remove this is not the right solution for several reasons:
* It's an option that provides fine-grained control of a niche part of the generated output and is not generalized to other use-cases
* We need to think carefully about adding any more commandline flags as that may interact poorly if we ever do https://github.com/golang/protobuf/issues/39, which would allow `protoc-gen-go` to output multiple files at once. In such a case, would the commandline flag affect all outputted files? If yes, is that wise to do?
* It's use case becomes fuzzy if we ever do https://github.com/golang/protobuf/issues/256, which would subvert control of the `json` tags.
* I'd rather see https://github.com/golang/protobuf/issues/52 be supported, which can be used to specify this yourself.

I'm going to close this in favor of #52, which is the closest issue to being able to support this.

PR #683, opened 2018-08-20
add no_omitempty option to remote omitempty label from generated struct

sparkthu:
For #682 
add an new option "no_omitempty", which is default false to keep default appearance.
when call` go_out=no_omitempty=true:./out.go`, structs without omit_empty label for json will be generated.


sparkthu:
I signed it!

PR #684, opened 2018-08-20
add no_omitempty option to remote omitempty label from generated struct

sparkthu:
#682

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -86,7 +86,7 @@ type Plugin interface {
 }
 
 var plugins []Plugin
-
+var noOmitEmpty bool
 // RegisterPlugin installs a (second-order) plugin to be run when the Go output is generated.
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for the pull request, but we're going to decline. See my https://github.com/golang/protobuf/issues/682#issuecomment-414407450

sparkthu:
@cybrcodr  I tried to use jsonpb, but the name style is totally different with json. We already have lots of code using underline name style instead of camel which is generated by jsonpb.
If there is any method to use under line name style in jsonpb, please let me know. Thank you.

cybrcodr:
If you are using the original field name in the .proto file, you can set jsonpb.Marshaler.OrigName to true and that would use the original .proto field name which should be the same as the generated json tag name.

sparkthu:
@cybrcodr Awsome! I love it. Thank you so much

PR #685, opened 2018-08-20
More useful error message on setting extensions

dasmdasm:
Including the types that are mismatched, rather than just stating that there was a type mismatch makes the error easier to debug and fix.

### DIFF
--- a/proto/extensions.go
+++ b/proto/extensions.go
@@ -488,7 +488,7 @@ func SetExtension(pb Message, extension *ExtensionDesc, value interface{}) error
 	}
 	typ := reflect.TypeOf(extension.ExtensionType)
 	if typ != reflect.TypeOf(value) {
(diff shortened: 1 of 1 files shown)


proto/extensions.go
```diff
@@ -488,7 +488,7 @@ func SetExtension(pb Message, extension *ExtensionDesc, value interface{}) error
 	}
 	typ := reflect.TypeOf(extension.ExtensionType)
 	if typ != reflect.TypeOf(value) {
-		return errors.New("proto: bad extension value type")
+		return fmt.Errorf("proto: bad extension value type. want: %T, got: %T", extension.ExtensionType, value)
```
dsnet: Nit: Usually, we mention got before want.

dasmdasm: Done.

dsnet:
Thanks!

PR #696, opened 2018-09-10
proto: uniquify descriptor var based on filename alone

neild:
The generated per-file var containing the file's FileDescriptorProto
needs to be unique across all files in a package.

Change the var name from:

  fileDescriptor_<file basename>_<hash of descriptor>

to:

  fileDescriptor_<hash of filename>

There's no need for the name of the var to change every time the
descriptor contents change (although this is harmless). The filename
alone suffices to make the var unique.


dsnet:
Your commit message looks wonky since Markdown treats the angle brackets speciallly.

PR #705, opened 2018-09-14
proto: replace exprs with matching consts from math pkg

Quasilyte:
Use named constants from math package instead of expressions
that evaluate to the same value.


Quasilyte:
I'll take a look at the build failure in a moment.

Quasilyte:
Sorry, off-by-1 error from me. Should be fixed now.

PR #706, opened 2018-09-14
proto: replace len(b)<=0 with len(b)==0

Quasilyte:
len never returns negative values, so len(b)==0 states
requirements in a more clear way with less "uncertainty".


dsnet:
Thanks.

PR #707, opened 2018-09-14
protoc-gen-go: put imports in a single import() block

neild:
Put all the imports into a single block.

Before:

	import proto "github.com/golang/protobuf/proto"
	import fmt "fmt"
	import math "math"

After:

	import (
		fmt "fmt"
		proto "github.com/golang/protobuf/proto"
		math "math"
	)

Sort imports using ast.SortImports rather than doing our own sorting,
ensuring consistency with gofmt.

This is a trivial change, but allows for identical output from this
version of the generator and one which uses astutil.AddNamedImport to
patch imports into the AST.

### DIFF
--- a/conformance/internal/conformance_proto/conformance.pb.go
+++ b/conformance/internal/conformance_proto/conformance.pb.go
@@ -3,15 +3,17 @@
 
 package conformance
 
-import proto "github.com/golang/protobuf/proto"
-import fmt "fmt"
-import math "math"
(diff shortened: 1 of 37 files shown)


neild:
I think this leaves the order unchanged, actually--we always sorted the imports by import path (other than the fixed ones of fmt, math, and proto). This just delegates the sorting to the ast package.

PR #714, opened 2018-09-17
fix symbol names on public imports

neild:
This PR includes two commits: The first adds new test cases that better demonstrate the effects of the second.


dsnet:
Lovely title. You'll probably want to use the "rebase and merge" feature.

PR #720, opened 2018-09-28
protoc-gen-go: put all imports in one section

neild:
Keep plugin imports and non-plugin imports in the same section, but
don't try to split stdlib and non-stdlib imports into different
sections. Consistent with using astutil to insert all imports from a
blank slate.


dsnet:
Oops. I accidentally merged this instead of approved it. GitHub could benefit from a 1-click approve button.

PR #722, opened 2018-10-03
protoc-gen-go: predeclared identifiers in cleanPackageName

sesmith177:
Don't generate identifiers that conflict with predeclared identifiers, prepend with "_" instead. This matches the behavior for package names that are Go keywords.

We ran into this issue during https://github.com/envoyproxy/envoy/pull/4556. Code with import statement:

```
import string "github.com/envoyproxy/data-plane-api/api/string"
```

was being generated and would not compile.

The list of predeclared indentifiers was taken from: https://golang.org/ref/spec#Predeclared_identifiers

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -567,7 +567,8 @@ func RegisterUniquePackageName(pkg string, f *FileDescriptor) string {
 	return string(name)
 }
 
-var isGoKeyword = map[string]bool{
+var isGoKeywordOrPredeclaredIdentifier = map[string]bool{
+	// Keywords
 	"break":       true,
 	"case":        true,
 	"chan":        true,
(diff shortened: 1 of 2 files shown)


dsnet:
While naming a package name as a built-in identifier is not great, isn't this a breaking change?

neild:
In the case of a package with a name that conflicts with a predefined identifier, that is imported without a local renaming, and that doesn't conflict with any identifier used in the places it is imported, this could be a breaking change.

Testing the change inside Google's codebase isn't likely to turn up anything broken by this, because our local style is to always rename imported proto packages.

I think generating a local rename that conflicts with a predefined identifier (e.g., `import string "..."`) is obviously wrong. We could keep the previous behavior for determining package names (`package string` is probably a bad idea, but it'll compile) and apply the new one for local renames only.

dsnet:
In most cases, the user should specify the `go_package` option, where they manually determine what the Go package path and name should be. If they really wanted to name a package the same as a built-in identifier, we should preserve that behavior. I find it odd that we do magical name cleanups for a string that the user explicitly sets. In my opinion, if they put an invalid package name (perhaps a typo), we should have avoided magically cleaning it up for them, and let the Go compiler complain. It's unfortunate that we sanitize the user-provided string today, but I'm not sure adding more magical sanitization logic is a good idea.

Also, I can understand hardcoding the keywords since they cannot be used as identifiers and it is very unlikely that we'll add new keywords even in Go2 (at least the generics proposal seems to make a conscious effort to avoid a new keyword). However, it seems more plausible that new built-in identifiers may be added to Go over time, which would presumably be added to the backlist. The idea that the output of this sanitization logic can change over time is concerning. 

> apply the new one for local renames only.

SGTM.

PR #725, opened 2018-10-08
protoc-gen-go: don't force space after "//" in comments

neild:
Avoid turning comments like "//////" into "// ////".

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1295,7 +1295,7 @@ func (g *Generator) makeComments(path string) (string, bool) {
 	w := new(bytes.Buffer)
 	nl := ""
 	for _, line := range strings.Split(strings.TrimSuffix(loc.GetLeadingComments(), "\n"), "\n") {
-		fmt.Fprintf(w, "%s// %s", nl, strings.TrimPrefix(line, " "))
+		fmt.Fprintf(w, "%s//%s", nl, line)
 		nl = "\n"
 	}
 	return w.String(), true


dsnet:
Wouldn't this cause the normal case to kind of odd?
```go
//Compute constructs a location mapping from input to output.
//An error is reported if any of the tokens of output cannot be mapped.
```
where the preceding space is elided?

Also, how come no golden files are updated as part of this?

dsnet:
Oh, I see. It seems that by convention our test protos already use a space after the leading "//", which protoc preserves, and we emit that ad-verbatim.

dsnet:
It's not obvious to me what the right behavior here is.

The use of `/*` now produces non-idiomatic comments. Given:
```proto
/*
 Foo is a foobabble.
 This is another line.
*/
message Foo {}
```
the new logic emits:
```go
//Foo is a foobabble.
//This is another line.
type Foo struct { ... }
```
It seems that `protoc` strips preceding whitespace for each line in a `/*` block, so maybe the current behavior is useful for block comments. However, we are unable to differentiate between line and block comments since `protoc` stripped them for us.

Fortunately, it also seems overwhelmingly that the style in `.proto` files is to use line comments with a space afterwards such that this does not matter.

Also, assuming the input were line comments only, this properly preserves the nature line comment exactly.

In general, it seems that there's no good way to convert `.proto` comments to language specific comments. For example, the `.proto` files in the google/protobuf repo assumes comments are markdown, which leads to obscure godoc renderings.

Given the above, I think doing the simplest thing (copying ad-verbatim) seems reasonable.

PR #726, opened 2018-10-10
Add Tars rpc support

sandyskies:
Tars, aka TAF(Total Application Framework), has been used in Tencent since 2008. It supports C++,Java,Nodejs ,php, and golang for now. This framework offers a set of solution for development, maintenance and testing, which making develop, deploy and testing service efficiently. It integrated extensible protocol for encoding/decoding, high-performance RPC communication framework, name service, monitor, statistics and configuration .
project : https://github.com/TarsCloud/Tars 
TarsGo is the Go version of tars.  TarsCpp and TarsJava have all support for  protocol buffer, so TarsGo will support protocol buffer too . In this case , I implement a plugin like grpc plugin in protoc-gen-go . 
See more about TarsGo  : https://github.com/TarsCloud/TarsGo


neild:
Thanks for the PR. I'm afraid, however, that this should go in a repository that you own rather than here. This repo is for general-purpose protocol buffer support. (The gRPC code generator is admittedly in this repo, which is unfortunate. We're working on moving gRPC code generation support out into a gRPC-owned repo instead.)

sandyskies:
@neild  when it's be done?  What if pepole what to use tars rpc plugin ,should I maintain it myself ?

neild:
I would recommend maintaining a separate protoc plugin that generates the tars services. This is how gRPC works for every language other than Go--there are, for example separate `protoc-gen-python` and `protoc-gen-python-grpc` protoc plugins which generate one file with the generic protobuf structures and a second file with the gRPC-specific bits.

We're going to move to a similar model for Go, although I don't have a specific timeline yet. You can see the work in progress on the `api-v2` branch of this repo.

PR #729, opened 2018-10-17
proto: remove panic behavior when registering a duplicate enum

vosst:
This change unifies the behavior when trying to register an already registered type. Instead of panic'ing, a warning is printed and the registration is not altered. 

This solves issues with duplicate registration of enums caused by switching to golang's modules. Prior to modules, packages would vendor protobuf, and registration of common types (like MapBox's vector tile format, for example) happened in the context of the vendored protobuf. When switching on modules, this is no longer the case and pulling in aforementioned packages as dependencies results in broken runtime behavior.

I have been thinking about multiple approaches here, among them:
 * being able to pass in a registration context to `Register*` functions. While certainly interesting, the implications are significant as any code relying on the registered entities would have to be made aware of the original registration context. Probably solvable with the code generator, but it seems like this approach requires further discussion.
 * refactor `Register*` functions to return an `error`  and adjusting the code generator to support a flag for specifying error handling strategies for the case of `Register*` returning an error. Together with something along the lines of `Err{Type/Enum/File}AlreadyRegistered`, this seems like a reasonable way forward.
 * introduce `var EnforceUniqueRegistration = false` in package `proto`. This looked promising at first, but an env var would be needed to be able to adjust the behavior, as calling code would not be able to alter the value prior to `init` of package `proto` being run at startup.

With that, the approach presented here seems to be easy enough and falls in line with existing policies for handling duplicate registration of types.

### DIFF
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -439,14 +439,14 @@ func getPropertiesLocked(t reflect.Type) *StructProperties {
 
 // A global registry of enum types.
 // The generated code will register the generated maps by calling RegisterEnum.
-
 var enumValueMaps = make(map[string]map[string]int32)
 
 // RegisterEnum is called from the generated code to install the enum descriptor
 // maps into the global table to aid parsing text format protocol buffers.
 func RegisterEnum(typeName string, unusedNameMap map[int32]string, valueMap map[string]int32) {
 	if _, ok := enumValueMaps[typeName]; ok {
-		panic("proto: duplicate enum registered: " + typeName)
+		log.Printf("proto: duplicate enum registered: %s", typeName)
+		return
 	}
 	enumValueMaps[typeName] = valueMap
 }


dsnet:
Hi, thanks for the PR.

> This change unifies the behavior when trying to register an already registered type. Instead of panic'ing, a warning is printed and the registration is not altered.

Not really:
* `RegisterExtension`: panics on conflicting entry
* `RegisterType`: first entry takes precedence (log warning on conflicts)
* `RegisterEnum`: panic on conflicting entry
* `RegisterMapType`: first entry takes precedence (log warning on conflicts)
* `RegisterMessageSetType`: last entry takes precedence
* `RegisterFile`: last entry takes precedence

As you can see the situation today is unfortunate and inconsistent. The whole registration system needs to be re-evaluated as a whole with consideration to the totality of the protobuf ecosystem. Some components in Go protobufs make the assumption of global uniqueness and changing these without thinking carefully of the implications is unwise.

> This solves issues with duplicate registration of enums caused by switching to golang's modules.

In general, switching to modules should reduce the registration conflicts, not increase them (although the conflict may have been masked in vendoring if both the generated protobufs and the `proto` runtime were vendored together; but there are other issues with that). Modules ensure that only one version of a given import path is built into the binary. Thus, with modules, registration conflicts can only occur if two proto files are in the same namespace and have the same type names.

In other words, conflicts after using modules implies an issue of conflicts not with how the Go binary is built, but with the set of proto files that is being built into the binary. This suggests a deeper issue where protobufs are being used in a way that wasn't intended or originally envisioned.

> like MapBox's vector tile format, for example

Reference? Examples are helpful if we can study them to understand why they're being problematic.

> being able to pass in a registration context to Register* functions.

This is a reasonable alternative, but requires proper API infrastructure to be built up. The upcoming v2 APIs is capable of supporting this (though we weren't planning on provided generator support for it).

> Probably solvable with the code generator

Rather that adding an option for the generator to select which context to register in, we were considering having an option in the generator to just drop registration all-together.

> refactor `Register*` functions to return an `error`

Knowing whether a registration conflict is problematic is not known at registration time. Registration into a global registry is primarily to support dynamic usages of protobufs. However, dynamic usages by nature implies that use of them is not statically known ahead of time. Furthermore, use of them is usually in code completely unrelated to the generated package.

> introduce `var EnforceUniqueRegistration = false` in package proto

As you noted, a global flag to control has problems. Furthermore, who is responsible for setting this? The author of package `main`? A library that depends on dynamic usages of protobufs and needs registration be internally consistent?

vosst:
@dsnet thanks for the detailed feedback. 

> Hi, thanks for the PR.
> 
> > This change unifies the behavior when trying to register an already registered type. Instead of panic'ing, a warning is printed and the registration is not altered.
> 
> Not really:
> 
> * `RegisterExtension`: panics on conflicting entry
> * `RegisterType`: first entry takes precedence (log warning on conflicts)
> * `RegisterEnum`: panic on conflicting entry
> * `RegisterMapType`: first entry takes precedence (log warning on conflicts)
> * `RegisterMessageSetType`: last entry takes precedence
> * `RegisterFile`: last entry takes precedence
> 

yup, I should have been more careful in terms of describing and naming the change. Thanks for the complete list here.

> As you can see the situation today is unfortunate and inconsistent. The registration system needs to be re-evaluated with consideration to the totality of the protobuf ecosystem. Some components in Go protobufs make the assumption of global uniqueness and breaking these guarantees without thinking carefully about the implications is unwise.
> 
> > This solves issues with duplicate registration of enums caused by switching to golang's modules.
> 
> In general, switching to modules should reduce the registration conflicts, not increase them (although the conflict may have been masked in vendoring if both the generated protobufs and the `proto` runtime were vendored together; but there are other issues with that). Modules ensure that only one version of a given import path is built into the binary. Thus, with modules, registration conflicts can only occur if two proto files are in the same namespace and have the same type names.
> 

I agree in general, but the masking issue (while transient) might warrant a transient escape hook to keep existing setups working while enabling them to transition to modules. I obviously lack numbers on the severity of the problem, though.

> In other words, conflicts after using modules implies an issue of conflicts not with how the Go binary is built, but with the set of proto files that is being built into the binary. This suggests a deeper issue where protobufs are being used in a way that wasn't intended or originally envisioned.
> 

Agreed in general, but fixing the deeper issues might take time (specifically in the case of external dependencies).

> > like MapBox's vector tile format, for example
> 
> Reference? Examples are helpful if we can study them to understand why they're being problematic.
> 

Please see:

 * https://github.com/go-spatial/tegola
 * in particular https://github.com/go-spatial/tegola/tree/master/mvt/vector_tile

In our example, we are using `tegola` and another package (not public on GitHub) that also pulls in MapBox's `vector_tile` proto. Both packages vendor their dependencies. *Without* modules, everything works as expected (types and enums are registered with the package-local protobuf instance). When modules are enabled, the resulting executable panics on startup as all proto registration happens in the context of a single protobuf 'instance'.

> > * being able to pass in a registration context to Register* functions.
> 
> This is a reasonable alternative, but requires proper API infrastructure to be built up. The upcoming v2 APIs is capable of supporting this (though we weren't planning on provided generator support for it).
> 
> > Probably solvable with the code generator
> 
> Rather that adding an option for the generator to select which context to register in, we were considering having an option in the generator to just drop registration all-together.
> 
> > * refactor `Register*` functions to return an `error`
> 
> Knowing whether a registration conflict is problematic is not known at registration time. Registration into a global registry is primarily to support dynamic usages of protobufs. However, dynamic usages by nature implies that use of them is not statically known ahead of time. Furthermore, use of them is usually in code completely unrelated to the generated package.
> 
> > * introduce `var EnforceUniqueRegistration = false` in package proto
> 
> As you noted, a global flag to control has problems. Furthermore, who is responsible for setting this? The author of package `main`? A library that depends on dynamic usages of protobufs and needs registration be internally consistent?

Agreed, having it exported certainly opens up those questions and they are difficult to answer in the general case. Another option would be to have `var enforceUniqueEnumRegistration = true` and an env var to set it to `false` (I dismissed the option a little early). If `false`, log and return early. With that, the caller of a final executable has an escape route available while dependencies are being adjusted.

dsnet:
> Please see:
>
> * https://github.com/go-spatial/tegola
> * in particular https://github.com/go-spatial/tegola/tree/master/mvt/vector_tile
>
>In our example, we are using `tegola` and another package (not public on GitHub) that also pulls in MapBox's `vector_tile` proto. Both packages vendor their dependencies. Without modules, everything works as expected (types and enums are registered with the package-local protobuf instance). When modules are enabled, the resulting executable panics on startup as all proto registration happens in the context of a single protobuf 'instance'.

I'm going to need some more details to understand. So the `tegola` project declares a `vectorTile` package (with an import path of `"github.com/go-spatial/tegola/mvt/vector_tile"`). Why is the "another package" not importing `"github.com/go-spatial/tegola/mvt/vector_tile"`?

vosst:
> > Please see:
> > 
> > * https://github.com/go-spatial/tegola
> > * in particular https://github.com/go-spatial/tegola/tree/master/mvt/vector_tile
> > 
> > In our example, we are using `tegola` and another package (not public on GitHub) that also pulls in MapBox's `vector_tile` proto. Both packages vendor their dependencies. Without modules, everything works as expected (types and enums are registered with the package-local protobuf instance). When modules are enabled, the resulting executable panics on startup as all proto registration happens in the context of a single protobuf 'instance'.
> 
> I'm going to need some more details to understand. So the `tegola` project declares a `vectorTile` package (with an import path of `"github.com/go-spatial/tegola/mvt/vector_tile"`). Why is the "another package" not importing `"github.com/go-spatial/tegola/mvt/vector_tile"`?

"another package" is not aware of `tegola`. More to this, the `tegola` package is not considered the `common` upstream providing pre-generated `pb.go` for MapBox's .proto. Which makes me think: In the case of modules, and with the panic in place, that would pretty much mean that an upstream for a common .proto would have to provide a go package, too, wouldn't it? Is providing pre-generated code considered a good practice?

vosst:
@dsnet ping :smile:

dsnet:
Apologies for not responding. I don't know why GitHub isn't sending me notifications for this even though I'm subscribed to the thread and you even pinged me about it. I happened to come back to this since it was on my TODO list.

---

Let me first describe what I perceive to be the current situation. Correct me if I'm wrong here.

There are at least three parties:
* **SRC**: This is the party that first came up with the `.proto` file. As the author, they control how the file changes. As such, their version is the "canonical" definition. In this specific situation, [MapBox is the source](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto).
* **USR1**: This is a party that wants to use the **SRC**'s version of the `.proto` file, but cannot since they do not have a Go package. So they fork **SRC**'s version and generate their own Go package. In this case, this is [what `github.com/go-spatial/tegola` has done](https://github.com/go-spatial/tegola/blob/master/mvt/vector_tile/vector_tile.proto).
* **USR2**: This is a party that is in the same situation as **USR1** and they do the same thing (fork **SRC**'s `.proto` and generate their own Go package). In this situation, this is your "another package".

In this ecosystem, **USR1** and **USR2** may be completely unrelated to each other and may even be different companies. They may not even know about each other's existence (although in this case you as **USR2** know about **USR1**, but possibly not the other way around).

A problem occurs when some end-user (possibly you) tries to link in code from both **USR1** and **USR2**, causing a registration conflict.

---

Given the problem described above, I see several solutions (with advantages and disadvantages).

#### Solution 1: Petition **SRC** to generate and maintain a Go package

This is the ideal solution since protobufs are built upon assumptions of universal uniqueness. If the **SRC** were to maintain their own Go package and place a `go_package` statement in their `.proto` file, then both **USR1** and **USR2** can use that generated Go package. It avoids registration conflicts since only one version is generated, imported, and thus registered. It means that resolving the name `"vector_tile.Tile"` resolves to the same Go type universally. Resolving types by name is important for the proper functioning of extensions and the `Any` well-known type.

The disadvantage of this solution is that it is a political problem where you need to convince the **SRC** of the importance of generating and maintaining a Go package for all their `.proto` file. Reasonably, the **SRC** may refuse to take on the burden of potentially maintaining generated code for all languages that their `.proto` file may be used in.

#### Solution 1a: Have **USR1** and **USR2** cooperate to maintain a single Go package

If Solution 1 is untenable, then the Go users of a `.proto` file can band together to maintain what is effectively the canonical version of **SRC**'s `.proto` files from the perspective of the Go ecosystem.

For this to work effectively, you still need the cooperation of **SRC** to update the `.proto` file with a `go_package` statement so that all Go users know where to find the Go package for that `.proto` file.

An example of this is [`descriptor.proto`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto), the canonical version of which is owned by the protocol buffer team. However, the canonical `.proto` file has a [`go_package` statement that declares which Go package is considered the canonical version](https://github.com/protocolbuffers/protobuf/blob/0038ff49af882463c2af9049356eed7df45c3e8e/src/google/protobuf/descriptor.proto#L43). You'll note that the Go package lives in this repository, which is different from the mainline protocol buffers repo.

#### Solution 2: vendor `github.com/golang/protobuf/proto`

This is approach taken today. It is a hacky approach to solution 5a using "go build" constructs. This works because it effectively turns each independent `proto` package into a local registry. However, the problem, as you noted, is that vendoring is going away with the advent of modules.

#### Solution 3: allow registration conflicts

This is the solution proposed by this PR. In the event of registration conflicts, the `proto` package just uses one of them arbitrarily. The problem with this approach is that extensions and `Any` may or may not work. Who knows? It depends on the order in which packages are registered. Furthermore, the linking in of **USR2**'s code may suddenly cause **USR1**'s code to stop working. Subtle and spooky affects at a distance is bad. Subtle and spooky effects that silently occurs is even worse.

#### Solution 4: change the proto namespace when forking

When **USR1** forks **SRC**'s `.proto` file, they change the proto namespace. For example, instead of [`package vector_tile`, they use `package user1.vector_tile`](https://github.com/mapbox/vector-tile-spec/blob/26b7823b16df1e15f55adb44243a3896acb7b804/2.1/vector_tile.proto#L1). Likewise, **USR2** does the same thing. (Note: under this approach both **USR1** and **USR2** should change the namespace, but it is possible for only one user to do a namespace change.)

This approach avoids registration conflicts since declarations in the `.proto` file have different names. It ensures the extensions and `Any` continues to work within a given user's ecosystem. However, extensions and `Any` breaks when interacting with **SRC**. For example, if **USR1** were to use gRPC with **SRC**, then extensions and `Any` will not work since the namespace has changed between the two.

#### Solution 5: provide generator option to disable registration

In this approach, there is some way for `protoc-gen-go` to avoid registering. Any usages of extensions or `Any` will require the user to manually resolve protobuf names, which can be more cumbersome. However, when done correctly, it ensures that extensions and `Any` work in all situations.

#### Solution 5a: provide generator option to specify which registry to use

In this approach, an option tells `protoc-gen-go` not to register with the global registry in the Go `proto` package, but possibly some user's local registry. It is still the user's responsibility to pass around local registries if they want extensions and `Any` to work properly, but registering into a local registry makes this work easier since there is only a single variable (representing the local registry) to pass around.

----

Analysis of the above:
* **Solution 1**: this is the ideal solution since it ensures that there is a single canonical Go package that everyone is supposed to use.
* **Solution 1a**: if **SRC** is not willing to maintain a canonical Go package, this is a decent approach as well.
* **Solution 2**: untenable in a post-module world.
* **Solution 3**: untenable since it provides no guarantees to either **USR1** or **USR2** for the use of extensions and `Any`. It's unfortunate that registration conflicts for messages are ignored, but we don't want to exacerbate the problem further. I'm glad that conflicts are surfaced readily for enums and I would rather see it panics on conflict registrations be stricter, rather than looser for messages.
* **Solution 4**: this is a hack, really. However, it does provide a given user, **USR1** the ability to work around the issue on their own regardless of **USR2** or **SRC** does. However, this solution does not work if extensions or `Any` are used. Since this is a course of action that **USR1** is making, they can look at their own codebase and determine whether they care about extensions or `Any`.
* **Solution 5**: this is an acceptable approach as well. However, it will require the design of an API for how disabling registration is specified (whether as a `.proto` file option, a command-line argument to `protoc`, or something else).
* **Solution 5a**: this is a better approach to solution 5, but cannot be done yet since the v2 API is not ready yet.

I suggest trying **solution 1**, then **solution 1a**. If you want something quick and dirty, then **solution 4** could work for you (it does not seem like `vector_tile` is a heavy user of extensions and `Any`; [there are extensions declared, but they don't seem to be used](https://github.com/mapbox/vector-tile-spec/blob/26b7823b16df1e15f55adb44243a3896acb7b804/2.1/vector_tile.proto#L27).).

**Solution 5** requires more planning. It is something we're leaning towards already. However, our attention is current fully devoted to the v2 API, and we'd prefer not addressing **solution 5** at this time since the v2 API is likely to affect what **solution 5** ends up looking like.

dsnet:
Closing this PR as it is not the direction we want to head. Happy to still discuss the general problem though.

PR #739, opened 2018-10-29
protoc-gen-go: include groups in public import forwarders

neild:
Groups were being excluded from the set of messages we generate
forwarding declarations for. Include them.


dsnet:
No tests?

neild:
Tests added.

PR #741, opened 2018-10-31
proto: deprecate {Unm,M}arshalMessageSet{JSON}

dsnet:
Despite the naming, these are "internal-only" functions that are intended to
only be called from generated code for MessageSets.
Furthermore, MessageSets are themselves a deprecated feature from proto1 days,
such that descriptor.proto even warns against their use since the initial
open-source release of protocol buffers in 2008. Within Google,
there are no direct usages of these functions, and all existing
usages of MessageSets go through the new table-driven implementation.

In addition to deprecating the {Unm,M}arshalMessageSet{JSON} top-level functions,
also modify the generator to stop emitting MarshalJSON and UnmarshalJSON methods
for messages sets. The UnmarshalJSON method is not implemented and the
MarshalJSON method does not seem to be called anywhere in Google (verified by
making the method panic and doing a global test). The jsonpb package continues
to work with MessageSets.

I should note that when the table-driven implementation was open sourced
in late 2017 (see 8cc9e46429bfb16289d40d30b2ee3f4923b47345), it accidentally
removed generation of the Marshal and Unmarshal method. However, no one seemed
to have noticed that those methods were no longer generated.

### DIFF
--- a/proto/deprecated.go
+++ b/proto/deprecated.go
@@ -31,8 +31,33 @@
 
 package proto
 
+import "errors"
+
 // Deprecated: do not use.
 type Stats struct{ Emalloc, Dmalloc, Encode, Decode, Chit, Cmiss, Size uint64 }
 
 // Deprecated: do not use.
 func GetStats() Stats { return Stats{} }
+
+// Deprecated: do not use.
+func MarshalMessageSet(interface{}) ([]byte, error) {
+	return nil, errors.New("proto: not implemented")
+}
+
+// Deprecated: do not use.
+func UnmarshalMessageSet([]byte, interface{}) error {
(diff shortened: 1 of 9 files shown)


dsnet:
According to a quick GitHub code search, there are no known external MessageSet messages:

https://github.com/search?l=Go&q=MarshalMessageSet+NOT+extMap+NOT+messageSetDesc+NOT+Request_Color_value+NOT+GenerateAlias&type=Code

dsnet:
\cc @jmarais @donaldgraham (maintainers over at gogo/protobuf).

Let me know if you have reason to believe people depend on these. Messages were already deprecated upon the first open source release of protobufs in 2008 and I'm aware of zero real usages externally.

PR #743, opened 2018-11-05
improve write performance for protobuf (5 times more in throughput, 50% less CPU usage on AIX)

amandeepgautam:
This patch improves the performance of writes. Tested on AIX and successfully ran `go test text_test.go`. On our AIX LPAR, we were seeing very high CPU usage, which eventually lead to very low performance while writing files. This patch improves CPU performance by at least 50% on the LPAR we have and improves the write speed by 5 times in our setup.

I am just creating this pull request for an early review and am attaching CPU profiles from before and after the change.

I have done 2 things:
-- batched writes
-- avoid using fprintf

Before:
[
before
](url)

After:
[
after
](url)

Please let me know if a more detailed performance analysis is needed to get this merged in.

### DIFF
--- a/proto/text.go
+++ b/proto/text.go
@@ -561,34 +561,43 @@ func writeString(w *textWriter, s string) error {
 	if err := w.WriteByte('"'); err != nil {
 		return err
 	}
+
+	// Allocate the slice. This should reduce few allocations.
+	aSlice := make([]byte, 0, len(s))
 	// Loop over the bytes, not the runes.
 	for i := 0; i < len(s); i++ {
-		var err error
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thank you for the PR. At the moment, the text proto implementation is in a frozen state (only bug fixes accepted). We're working on v2 of this code which will completely replace this implementation. In fact, it [currently takes an `append` approach to constructing strings](https://github.com/golang/protobuf/blob/95b0290ea849bd21f9b747291b4b16a223354fd4/internal/encoding/text/string.go#L26-L81) similar to this PR.

PR #746, opened 2018-11-08
proto: store extension values according to protobuf data model

dsnet:
The current API represents scalar extension fields as *T and
repeated extension fields as []T.
However, this is not an accurate reflection of the protobuf data model.

For scalars, pointers are usually used to represent nullability. However,
in the case of extension scalars, there is no need to do so since presence
information is captured by checking whether the field is in the extension map.
For this reason, presence on extension scalars is not determined by checking
whether the returned pointer is nil, but whether HasExtension reports true.

For repeated fields, using []T means that the returned value is only a partially
mutable value. You can swap out elements, but you cannot change the length of
the original field value. On the other hand, the reflective API provides methods
on repeated field values that permit operations that do change the length.
Thus, using *[]T is a closer match to the protobuf data model.

This CL changes the implementation of extension fields to always store T
for scalars, and *[]T for repeated fields. However, for backwards compatibility,
the API continues to provide *T and []T. In theory, this could break anyone
that relies on memory aliasing for *T. However, this is unlikely since:
* use of extensions themselves are relatively rare
* if extensions are used, it is recommended practice to use a message as the
field extension and not a scalar
* relying on memory aliasing is generally not a good idiom to follow.
The expected pattern is to call SetExtension to make it explicit that a mutation
is happening on the message.
* analysis within Google demonstrates that no one is relying on this behavior.

### DIFF
--- a/proto/clone_test.go
+++ b/proto/clone_test.go
@@ -67,34 +67,50 @@ func init() {
 	if err := proto.SetExtension(cloneTestMessage, pb.E_Ext_More, ext); err != nil {
 		panic("SetExtension: " + err.Error())
 	}
+	if err := proto.SetExtension(cloneTestMessage, pb.E_Ext_Text, proto.String("hello")); err != nil {
+		panic("SetExtension: " + err.Error())
+	}
+	if err := proto.SetExtension(cloneTestMessage, pb.E_Greeting, []string{"one", "two"}); err != nil {
+		panic("SetExtension: " + err.Error())
+	}
 }
 
 func TestClone(t *testing.T) {
+	// Create a clone using a marshal/unmarshal roundtrip.
+	vanilla := new(pb.MyMessage)
+	b, err := proto.Marshal(cloneTestMessage)
+	if err != nil {
+		t.Errorf("unexpected Marshal error: %v", err)
+	}
+	if err := proto.Unmarshal(b, vanilla); err != nil {
+		t.Errorf("unexpected Unarshal error: %v", err)
+	}
+
+	// Create a clone using Clone and verify that it is equal to the original.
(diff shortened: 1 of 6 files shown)


proto/clone_test.go
```diff
@@ -67,6 +67,12 @@ func init() {
 	if err := proto.SetExtension(cloneTestMessage, pb.E_Ext_More, ext); err != nil {
 		panic("SetExtension: " + err.Error())
 	}
+	if err := proto.SetExtension(cloneTestMessage, pb.E_Ext_Text, proto.String("hello")); err != nil {
+		panic("SetExtension: " + err.Error())
+	}
+	if err := proto.SetExtension(cloneTestMessage, pb.E_Greeting, []string{"one", "two"}); err != nil {
```
cybrcodr: Do you also need to verify that clone makes a deep copy on this?

proto/extensions.go
```diff
@@ -541,3 +541,47 @@ func RegisterExtension(desc *ExtensionDesc) {
 func RegisteredExtensions(pb Message) map[int32]*ExtensionDesc {
 	return extensionMaps[reflect.TypeOf(pb).Elem()]
 }
+
+func extensionAsLegacyType(v interface{}) interface{} {
```
neild: This function and extensionAsStorageType need doc comments explaining what they do.

proto/clone_test.go
dsnet: Done.

proto/extensions.go
dsnet: Done. I commented on `Extension.value` had referenced it from `extensionAsLegacyType` and `extensionAsStorageType`.

proto/clone_test.go
dsnet: Done

PR #753, opened 2018-11-16
jsonpb: add support for type aliased slices to implement JSONPBMarshaler

joshlreese:
Added support to the marshaling process to check if a slice implements
JSONPBMarshaler - this is useful in situations where one wants to encode
a slice as a JSON object.


dsnet:
Hi, thanks for the PR.

Unfortunately the v1 `jsonpb` package is frozen and being completely rewritten for v2. Also, I regret accepting the `JSONPBMarshaler` in the first place since it was a hack around the fact that we do not have protobuf reflection. The v2 API forgoes the need for that interface since it requires all messages to properly implement protobuf reflection. As such, I do not want to see this interface be expanded in it's use-cases.

joshlreese:
Fair enough! I appreciate the rapid response, and I'll keep an eye on that v2 work.

Issue #406, opened 2017-08-08
jsonpb: with EmitDefaults=true, nil map/slice should marshal to "null"


cybrcodr:
Current implementation with EmitDefaults=true, marshals map to JSON {} and slice to JSON [] regardless of whether map/slice is nil or empty instance.  Spec does not dictate any rules on this afaik.  Note that if EmitDefaults=false, empty or nil is treated as "default" value and hence is omitted in the JSON data, which I think is appropriate.

With commit 748d386b5c1ea99658fd69fe9f03991ce86a90c1, library unmarshals JSON "null" to nil map/slice instead of empty instance for the same rationale as mentioned in PR #394.  So perhaps it would be more intuitive/natural to make the distinction in the reverse direction to distinguish nil vs empty, i.e. empty map marshals to {}, nil map marshals to null, empty slice marshals to [], nil slice marshals to null.

The other part that comes with the above suggestion is what to do with WKT Struct and ListValue.  Struct wraps a map field "fields" and ListValue wraps a repeated field "values".  However, marshaling these types are typically based on marshaling the inner fields directly if not empty/nil.  If one sets a field with empty &Struct{}, the inner map "fields" is still nil.  Current implementation marshals it into {} regardless of EmitDefaults.

@zombiezen -- thoughts on this?

dsnet:
The current behavior is correct. In the protobuf data model, there is no distinction between an "unpopulated" list or map and an "empty" list or map. Therefore, a pure protobuf reflection implementation of JSON cannot distinguish between these two possible outputs.

PR #754, opened 2018-11-16
jsonpb: Add SuppressCompositeDefaults flag to suppress composite fields with zero values when EmitDefaults is true.

joshlreese:
Allow a bit of flexibility in emitting default values for maps, slices,
and nested structs.


dsnet:
As mentioned in #753, the v1 `jsonpb` is being frozen in terms of feature-set. Can you file an issue about this and I'll think about the implications of it for v2? Thanks.

dsnet:
You'll probably want to reference #406.

joshlreese:
Will do.

Issue #708, opened 2018-09-15
protoc-gen-go: simplify generated oneof implementation


dsnet:
The logic to generate support for oneofs are a significant source of complexity:
* It is [complexity to maintain in the generator](https://golang.org/cl/135535).
* It is [complexity in the generated code](https://github.com/golang/protobuf/blob/a47340a8e42c4ceb44c00c2a05199e24b24fcaef/proto/test_proto/test.pb.go#L3365-L3679).
* It is complexity in the `proto` package as the generated code depends on the following types, functions, variables, and constants: `Buffer`, `SizeVarint`, `ErrInternalBadWireType`, `WireBytes`, `WireEndGroup`, `WireFixed`, `WireStartGroup`, and `WireVarint`.
* It exposes a pseudo-internal method `XXX_OneofFuncs`.

However, this does not need to be the case:
* The `proto` runtime, since switching to the table-driven approach, uses `XXX_OneofFuncs` in [marshal](https://github.com/golang/protobuf/blob/master/proto/table_marshal.go#L323-L325) and [unmarshal](https://github.com/golang/protobuf/blob/master/proto/table_unmarshal.go#L365-L368) only to obtain type information about the wrappers. 

Since the type information is the only information needed, we can accomplish this in a cleaner way. Instead of adding the method, we add an unexported, zero-length field with all the wrapper types:
```go
type FooMessage struct {
    // other fields of FooMessage as usual
    xxx_oneofWrappers [0]struct{
        Oneof_F_Bool
        Oneof_F_Int32
        Oneof_F_Int64
        Oneof_F_Fixed32
        Oneof_F_Fixed64
        Oneof_F_Uint32
        Oneof_F_Uint64
        Oneof_F_Float
        Oneof_F_Double
        Oneof_F_String
        Oneof_F_Bytes
        Oneof_F_Sint32
        Oneof_F_Sint64
        Oneof_F_Enum
        Oneof_F_Message
        Oneof_FGroup
        Oneof_F_Largest_Tag
        Oneof_Value
    }
}
```

This has the following properties:
* It removes the need for the exported XXX method (see #276). In fact, this approach keeps all of the type information hidden. Using Go reflection, you can obtain a list of all the oneof wrappers types from the type information in the message alone (no need to dynamically call methods).
* It simplifies the implementation of `proto` since we don't have to type assert for a method with a relatively complex signature.
* `unsafe.Sizeof(FooMessage{})` remains unchanged
* The generated code is significantly simpler.

[...]

PR #760, opened 2018-11-26
protoc-gen-go: generate XXX_OneofWrappers instead of XXX_OneofFuncs

dsnet:
The marshaler, unmarshaler, and sizer functions are unused ever since
the underlying implementation was switched to be table-driven.
Change the function to only return the wrapper structs.

This change:
* enables generated protos to drop dependencies on certain proto types
* reduces the size of generated protos
* simplifies the implementation of oneofs in protoc-gen-go

Updates #708

### DIFF
--- a/conformance/internal/conformance_proto/conformance.pb.go
+++ b/conformance/internal/conformance_proto/conformance.pb.go
@@ -24,7 +24,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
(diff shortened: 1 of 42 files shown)


dsnet:
I realized that this PR is near-impossible to review. The relevant code changes are in:
* proto/lib.go proto/lib.go
* proto/properties.go proto/properties.go
* proto/table_marshal.go proto/table_marshal.go
* proto/table_unmarshal.go proto/table_unmarshal.go
* protoc-gen-go/generator/generator.go protoc-gen-go/generator/generator.go

dsnet:
This change broke a few targets inside Google that were directly calling XXX_OneofFuncs. Since the compatibility agreement explicitly gives us the right to add and remove XXX fields and methods, this is clearly their fault.

To be nice, however, I went and fixed those broken cases for them by calling `proto.GetProperties`.

PR #761, opened 2018-11-26
protoapi: new package for types that generated messages depend on

dsnet:
Create a new package protoapi that contains the minimum set of types depended
upon by generated messages so that a generated message implemented the v2
APIs do not need to import the v1 proto runtime.

Some types like proto.Buffer are not included here since those dependencies
are going away with PR#760.

Explicitly document that no one should import this package directly,
so we have the flexibility to remove this package if necessary.


dsnet:
Disregard this PR. I'm going to be making these changes on v1 branch.

PR #766, opened 2018-11-29
Update README.md

shanghai-Jerry:
add more details about tracing errors while using protoc-gen-go plugin to get `.pb.go` file


dsnet:
I'm not sure this adds anything. Saying `$GOPATH/bin` must be in `$PATH` is not strictly correct, since it could be in `$GOBIN`. The important property is that the `protoc-gen-go` binary be resolvable when searching `$PATH`. I think this is a common enough Unix and Windows idiom that it doesn't need to be expanded upon, lest the README becomes overly verbose.

shanghai-Jerry:
i am not sure if i set $GOBIN, it will work or not. i'll check it, perhaps it's not must be, it's just one way to avoid that error

shanghai-Jerry:
`
➜  bin echo $GOBIN
/Users/devops/workspace/gopath/bin
➜  bin echo $GOPATH
/Users/devops/workspace/gopath
➜  bin protoc --go_out=. ai_call_operation.proto
protoc-gen-go: program not found or is not executable
--go_out: protoc-gen-go: Plugin failed with status code 1.
➜  bin PATH=/Users/devops/workspace/gopath/bin:$PATH
➜  bin protoc --go_out=. ai_call_operation.proto
`
it seems if i set $GOBIN, but it still has the error.  Once i added $GOBIN to $PATH, it works
you  can check this by yourself

shanghai-Jerry:
`Thanks, your CLA has been submitted.`

I signed it!

PR #771, opened 2018-12-05
gRPC client service info

piotrkowalczuk:
This PR introduces new function in gRPC specific generated code `ClientServiceInfo`. It accepts an `interface{}` and returns a `grpc.ServiceInfo`.

The rationale for this feature is to expose more detailed information about the service on the client side. It can be used for example to preallocate metrics. Something that is already possible on a server side.


dsnet:
\cc @dfawley 

The v2 API is aiming to make protobuf reflection a first-class feature. This PR adds overlap with that work and I'm hesitant to see this be merged.

dfawley:
I would also prefer to accomplish this with support in the primary proto package/generator, and not in the grpc codegen.  (See also this related issue: https://github.com/grpc/grpc-go/issues/1914)

PR #782, opened 2019-01-08
protoc-gen-go: fix oneof name mangling regression

### LINKED ISSUE #780: Unexpected trailing underscore in oneof field name
**What version of protobuf and what language are you using?**
Version: go1.11.4 darwin/amd64

**What did you do?**
```proto
syntax = "proto2";

package oneoftest;

// If you rename this to `PutBar`, the problem goes away.
message GetBar {
    optional string a_string = 1;
}

message Foo {
  /**
   * This field is required.
   */
  oneof bar {
    GetBar get_bar = 1;
  }
}
```

**What did you expect to see?**
The oneof field on the struct should be named `Bar`, since there's nothing else for it to conflict with.

**What did you see instead?**
The oneof field is named `Bar_`.

```proto
type Foo struct {
	//*
	// This field is required.
	//
	// Types that are valid to be assigned to Bar_:
	//	*Foo_GetBar
	Bar_                 isFoo_Bar_ `protobuf_oneof:"bar"`
	XXX_NoUnkeyedLiteral struct{}   `json:"-"`
	XXX_unrecognized     []byte     `json:"-"`
	XXX_sizecache        int32      `json:"-"`
}
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment?**

IRL, we have things named `GetRequest` and `PutRequest`, and oneof fields named `Request`. I'm trying to upgrade our version of golang/protobuf, and this is causing widespread breakage.

I *think* it's happening because of the conflict with the `GetBar()` method on the struct, but the renaming of the field itself was unexpected, and caused breakage.

dsnet:
The generator currently uses an unintuitive and stateful algorithm
for name generation where it "fixes" name conflicts by appending "_"
to the end of the new name.

PR#657 refactored the generator code and noticed that the above
algorithm was not properly taking into account that a Get method is
generated for parent oneofs, fixing it in the same PR. While this is
more correct, this breaks users (see #780) since it means that the
generation of names can change.

This PR changes the name mangling logic to be as it was previously.
This does mean that some new proto files may be unbuildable,
but that is arguably better than breaking existing proto files.

Fixes #780


dsnet:
Before submitting, I'll run this through code at Google to make sure nothing depends on the new mangling semantics.

PR #789, opened 2019-01-17
doc: fix indentation on Any examples

bitspill:
Examples 3 and 4 had an extra leading space causing GoDoc to treat them as part of Example 2 rather than distinct snippets

https://i.imgur.com/FSwehb1.png
https://godoc.org/github.com/golang/protobuf/ptypes/any#Any

### DIFF
--- a/ptypes/any/any.pb.go
+++ b/ptypes/any/any.pb.go
@@ -45,7 +45,7 @@ const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
 //       foo = any.unpack(Foo.class);
 //     }
 //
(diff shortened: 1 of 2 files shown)


dsnet:
Hi, thanks for the PR, but this is a wrong repo to make the change.

This file is pulled in ad-verbatim from https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto, so a PR should made upstream.

bitspill:
Ah, sure enough. An upstream PR has been created https://github.com/protocolbuffers/protobuf/pull/5597

PR #793, opened 2019-01-25
adds check for ignore tag

### LINKED ISSUE #617: proto: upstream changes breaking compatibility with Kubernetes
https://github.com/golang/protobuf/commit/8cc9e46429bfb16289d40d30b2ee3f4923b47345 breaks compatibility with kubernetes. 

https://github.com/golang/protobuf/blob/master/proto/table_unmarshal.go#L331 mandates that all struct fields have proto tags. The `TypeMeta` fields in kubernetes structs do not contain a proto tag, https://github.com/kubernetes/apiserver/blob/master/pkg/apis/audit/v1beta1/types.go#L145 is a typical example.

If this change is necessitated we should work to put it in a major revision and allow `v1` to be compatible without tags.

pbarker:
From the discussions in #617 allows field to be ignored with the use of a "-" tag

closes #617


pbarker:
I signed it!

neild:
(I'm going to leave this open for @dsnet to look at and weigh in on, but he's out on paternity leave so it may take a while for him to get a chance to look at it.)

pbarker:
@neild we discussed the v2 api in #617, but its still a ways out and this is a somewhat pressing issue. Adding an ignore tag seems like a reasonable accommodation and provides utility for real world use cases. This is more of a stop gap to get us to v2, and would be surprised if it set any sort of detrimental precedent on the v1 api.

PR #805, opened 2019-02-28
proto: remove test dependency on experimental packages

AlexRudd:
Removes two go mod dependencies on `golang.org/x/sync` and indirectly on `golang.org/x/net`. These were being used to provide an `errgroup.Group` to collate errors in a test.

This PR provides the same functionality but using `sync.WaitGroup`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,3 @@
 module github.com/golang/protobuf
 
-require (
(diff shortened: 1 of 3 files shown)


dsnet:
Thanks for the cleanup.

PR #807, opened 2019-02-28
conformance: split into separate module

### LINKED ISSUE #806: remove inappropriate transitive dependencies by splitting `conformance` into a separate module
https://github.com/golang/protobuf/issues/804 reports an error with `go get -u` for this module.

The problem arises through a large transitive dependency graph, introduced via a dependency on `google.golang.org/genproto` (and eventually `google.golang.org/grpc` and its wide net of dependencies) in the `conformance/internal/conformance_proto` package:

```
~/src/github.com/golang/protobuf$ go mod why -m google.golang.org/genproto
# google.golang.org/genproto
github.com/golang/protobuf/conformance/internal/conformance_proto
google.golang.org/genproto/protobuf/field_mask
```

However, the `conformance` subtree doesn't even build using `cmd/go`: it has its own `Makefile` and `.sh` test driver. It should not be included in the main module, and then the inappropriate `grpc` dependency will be removed.

That will also help to address https://github.com/golang/lint/issues/436.

CC @dmitshur @thepudds

bcmills:
Fixes #806

### DIFF
--- a/conformance/go.mod
+++ b/conformance/go.mod
@@ -1,0 +1,10 @@
+module github.com/golang/protobuf/conformance
+
+go 1.11
+
+require (
+	github.com/golang/protobuf v1.3.0
+	google.golang.org/genproto v0.0.0-20180831171423-11092d34479b
+)
+
+replace github.com/golang/protobuf => ../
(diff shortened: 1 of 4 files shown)


dsnet:
I prefer #808 over this.

PR #818, opened 2019-03-16
protoc-gen-go/grpc: inline errUnimplemented function

### LINKED ISSUE #817: errUnimplemented redeclared in this block
**What version of protobuf and what language are you using?**
Version: (e.g., `v1.1.0`, `89a0c16f`, etc)
v3.6.0, Golang

**What did you do?**
If possible, provide a recipe for reproducing the error.
A complete runnable program is good with `.proto` and `.go` source code.
We have multiple proto files in the same go package.

**What did you expect to see?**
Compiles are failing.

**What did you see instead?**
We use go get for the following packages.
        "google.golang.org/grpc"
        "github.com/golang/protobuf/protoc-gen-go"
        "github.com/golang/protobuf/ptypes"
        "github.com/golang/protobuf/jsonpb"
        "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway"
        "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"

For the last 3 hours our builds failing due to the following reason:
errUnimplemented redeclared in this block

This seems like a issue if anyone has more than 1 proto file in the same package. 

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

**Anything else we should know about your project / environment?**

dsnet:
Avoid using an errUnimplemented function, which requires us to
give a name to it, causing possible conflicts.
Instead, just inline it's functionality.

Fixes #817

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -118,10 +118,6 @@ func (g *grpc) Generate(file *generator.FileDescriptor) {
 	g.P("var _ ", contextPkg, ".Context")
 	g.P("var _ ", grpcPkg, ".ClientConn")
 	g.P()
-	g.P("func errUnimplemented(methodName string) error {")
-	g.P("\treturn ", errorPkg, ".Errorf(codes.Unimplemented, \"method %s not implemented\", methodName)")
-	g.P("}")
-	g.P()
 
 	// Assert version compatibility.
 	g.P("// This is a compile-time assertion to ensure that this generated file")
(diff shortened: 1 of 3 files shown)


solodynamo:
For folks using this repo using go get, their go builds will fail as It would throw errors like `imported "google.golang.org/grpc/codes but not used`. Its because their .proto files services defination not uses error codes. To fix use the older release of this repo ie. 1.3.1 by compiling using source instead of go get.
In case of docker : 
`
RUN mkdir -p ${GOPATH}/src/github.com/golang/protobuf && \
    curl -sSL https://github.com/golang/protobuf/archive/v${PROTOC_GEN_GO_VER}.tar.gz | tar -xz --strip 1 -C ${GOPATH}/src/github.com/golang/protobuf &&\
    cd ${GOPATH}/src/github.com/golang/protobuf && \
    go build -ldflags '-w -s' -o /golang-protobuf-out/protoc-gen-go ./protoc-gen-go && \
    install -Ds /golang-protobuf-out/protoc-gen-go ${OUTDIR}/usr/bin/
`
where PROTOC_GEN_GO_VER = 1.3.1

Issue #814, opened 2019-03-08
proto: textproto marshal writes comments with invalid syntax


darkfeline:
The textproto marshaller writes various comments using `/* */` syntax.  Example:

https://github.com/golang/protobuf/blob/f05648d464991ab1aa8cf6a499122c56f0f50f2f/proto/text.go#L597

This is invalid textproto format.  The parser respects comments starting with `#`

https://github.com/golang/protobuf/blob/f05648d464991ab1aa8cf6a499122c56f0f50f2f/proto/text_parser.go#L136

I couldn't find a spec for textproto, but the Python parser agress that comments start with `#`.

https://github.com/protocolbuffers/protobuf/blob/624a40a387e1b2da4c6bbde44ad3d0a5efb9879b/python/google/protobuf/text_format.py#L1060

dsnet:
\cc @cybrcodr

darkfeline:
(I'd be happy to help fix it if a dev agrees this makes sense, since it's easy.)

dsnet:
The current `master` branch is more or less in maintenance mode as all effort is going towards the v2 implementation, which actually doesn't have this bug.

My concern with making this change is the amount of people it breaks that depend on the output of text proto to be stable. These are clearly invalid usages, but fixing them is not something we want to deal with right now.

@cybrcodr can decide whether it's worth changing in v1.

puellanivis:
It seems like you’re right and the C and Python implementations only accept `#` as a comment style. However, the reason you cannot find any standard for the plain text representation of protobufs is because there is intentionally no standard for the simple text format. It was always intended that this text format would only be used for debugging and quick-and-dirty editing, and not for any form of actual interchange. It was decided that the best way to ensure this was to not actually define any standard at all, and if someone objected to a lack of interoperability, one would then jsut point to the intentionally blank space where a standard should be.

Once a JSON encoding was established and standardized, this only further eliminated any need to standardize the simple text format.

So, general rule of thumb is the plain text formatting `SHOULD` attempt to conform to a common implementation, but that it `CANNOT` be relied upon to actually be compatible as an interchange format, because there is no actual standard. If one needs a guarantee of interoperability, one `MUST` rely upon one of the two standardized encodings: binary or JSON.

cybrcodr:
That comment is only produced when there are unknown fields.  And afaik, unknown fields cannot be parsed back by either current Go or C++ parsers.  Hence, this hasn't been much of a priority.

[...]

PR #824, opened 2019-03-20
Fix textproto comment syntax

darkfeline:
https://github.com/golang/protobuf/issues/814

### DIFF
--- a/proto/text.go
+++ b/proto/text.go
@@ -420,7 +420,7 @@ func (tm *TextMarshaler) writeStruct(w *textWriter, sv reflect.Value) error {
 				// in writeAny.
 				if fv.Kind() == reflect.Ptr && fv.IsNil() {
 					// Use errors.New so writeAny won't render quotes.
-					msg := errors.New("/* nil */")
+					msg := errors.New("# nil")
 					fv = reflect.ValueOf(&msg).Elem()
 				}
 			}
@@ -594,7 +594,7 @@ func writeString(w *textWriter, s string) error {
 
 func writeUnknownStruct(w *textWriter, data []byte) (err error) {
 	if !w.compact {
-		if _, err := fmt.Fprintf(w, "/* %d unknown bytes */\n", len(data)); err != nil {
+		if _, err := fmt.Fprintf(w, "# %d unknown bytes\n", len(data)); err != nil {
 			return err
 		}
 	}
@@ -602,7 +602,7 @@ func writeUnknownStruct(w *textWriter, data []byte) (err error) {
 	for b.index < len(b.buf) {
 		x, err := b.DecodeVarint()
 		if err != nil {
-			_, err := fmt.Fprintf(w, "/* %v */\n", err)
+			_, err := fmt.Fprintf(w, "# %v\n", err)
 			return err
 		}
 		wire, tag := x&7, x>>3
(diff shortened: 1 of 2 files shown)


proto/text.go
```diff
@@ -420,7 +420,7 @@ func (tm *TextMarshaler) writeStruct(w *textWriter, sv reflect.Value) error {
 				// in writeAny.
 				if fv.Kind() == reflect.Ptr && fv.IsNil() {
 					// Use errors.New so writeAny won't render quotes.
-					msg := errors.New("/* nil */")
+					msg := errors.New("# nil")
```
cybrcodr: This is actually a bug.  As shown in the testcase, the output is not a valid textproto, i.e. field name without value.  V2 will output an empty message for this as the oneof has been set to one of the fields.  I'd prefer to follow the same as v2.

proto/text.go
```diff
@@ -594,15 +594,15 @@ func writeString(w *textWriter, s string) error {
 
 func writeUnknownStruct(w *textWriter, data []byte) (err error) {
 	if !w.compact {
-		if _, err := fmt.Fprintf(w, "/* %d unknown bytes */\n", len(data)); err != nil {
+		if _, err := fmt.Fprintf(w, "# %d unknown bytes\n", len(data)); err != nil {
```
cybrcodr: While I prefer removing this, if you feel strongly that this is beneficial, feel free to keep it.

proto/text.go
```diff
@@ -594,15 +594,15 @@ func writeString(w *textWriter, s string) error {
...
 			return err
 		}
 	}
 	b := NewBuffer(data)
 	for b.index < len(b.buf) {
 		x, err := b.DecodeVarint()
 		if err != nil {
-			_, err := fmt.Fprintf(w, "/* %v */\n", err)
+			_, err := fmt.Fprintf(w, "# %v\n", err)
```
cybrcodr: I don't think the error message should be written to the output as it is already returned anyways.  I'd prefer removing this type of comment, same for the rest below.

proto/text.go
```diff
@@ -632,7 +632,7 @@ func writeUnknownStruct(w *textWriter, data []byte) (err error) {
 			if e == nil {
 				_, err = fmt.Fprintf(w, "%q", buf)
 			} else {
-				_, err = fmt.Fprintf(w, "/* %v */", e)
+				_, err = fmt.Fprintf(w, "# %v", e)
```
cybrcodr: ditto

proto/text.go
```diff
@@ -663,7 +663,7 @@ func writeUnknownInt(w *textWriter, x uint64, err error) error {
 	if err == nil {
 		_, err = fmt.Fprint(w, x)
 	} else {
-		_, err = fmt.Fprintf(w, "/* %v */", err)
+		_, err = fmt.Fprintf(w, "# %v", err)
```
cybrcodr: ditto

proto/text.go
puellanivis: I agree. I think returning any `error` that contains syntactically important parts of the output format is being overly clever, especially using an `error` type to avoid rendering quotes elsewhere.

Basically, the code as was is confusing, which is why it has a comment explaining itself. Since we’re here, we could instead make it less confusing, and much simpler.

darkfeline:
No problem, we'll probably move to JSON soon anyway.  Thanks for your help.

PR #827, opened 2019-03-27
proto: Ensure input to EncodeFixed32 does not exceed max uint32

jonathangold:
See https://github.com/gogo/protobuf/issues/536

### DIFF
--- a/proto/encode.go
+++ b/proto/encode.go
@@ -37,6 +37,7 @@ package proto
 
 import (
 	"errors"
+	"math"
 	"reflect"
 )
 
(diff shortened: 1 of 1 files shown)


jonathangold:
I signed it!

proto/encode.go
```diff
@@ -139,6 +144,9 @@ func (p *Buffer) EncodeFixed64(x uint64) error {
 // This is the format for the
 // fixed32, sfixed32, and float protocol buffer types.
 func (p *Buffer) EncodeFixed32(x uint64) error {
+	if x > math.MaxUint32 {
```
puellanivis: https://play.golang.org/p/xSeqrvM8Bye 😟 I think the naïve use of this function will cause this change to break a lot of code if taken as is.

dsnet:
The current behavior is unfortunate, but the reality is that people have come to depend on the overflow semantics.

jonathangold:
@dsnet Fair enough. Thanks for having a look. Until next time!

PR #829, opened 2019-04-02
Log parsing errors using log pkg.

dilyevsky:
Gets rid of annoying stderr messages. Consistent with the rest of the file.

---
This change is [Reviewable](https://reviewable.io/reviews/golang/protobuf/829)


dsnet:
This API is only intended for use with generated messages from `protoc-gen-go`, how were you able to trigger these failures in the first place?

dilyevsky:
This comes up a lot when trying to render Kubernetes psuedo-protos. Specifically this: https://github.com/kubernetes/apimachinery/blob/7b3d41122501114701f8d013ea1d9533f60c5dd5/pkg/apis/meta/v1/time.go#L34

dilyevsky:
Looks like history got overwritten. @dsnet should i re-open this or send it to some other repo?

neild:
We don't support message types other than the ones produced by `protoc-gen-go`, and I don't think we should set a precedent of accepting changes to support those types.

dilyevsky:
Hi Damien,

Not asking to support my special case - simply correcting what seems like
an inconsistency in this package which just so happens to present itself in
my admittedly non-standard use case.

Thanks,
-Dmitry

neild:
That's fair, and on a second look I do see that there are a lot of `log.Print` calls. We should be consistent. Go ahead and re-open this. (It was accidentally closed due to a config error--sorry!)

FWIW, the in-development revision of the package API neither logs nor prints.

dilyevsky:
Thanks for taking a look, @neild. Re-opened as #851.

PR #831, opened 2019-04-05
proto: always prefer non protoc-gen-go unmarshal functions

katallaxie:
## Details

This PR changes the precedence of the use of the generated methods to unmarshal protobuf data and the explicitly defined ones. This solves the case, where you have a generated go package and overwrite some of the marshaling functions with custom behavior upon need. 

Example: 

```golang
func (m *SomeEvent) XXX_Unmarshal(b []byte) error {
	return xxx_messageInfo_SomeEvent.Unmarshal(m, b)
}

func (m *SomeEvent) Unmarshal(b []byte) error {
     // some other foo
}
```

## Reviewers

* [x] Ready for review :+1:


katallaxie:
I signed it!

PR #834, opened 2019-04-05
fix indentation in jsonpb with Any messages

jhump:
If an Any message contains a message that knows how to marshal itself (e.g implements the `JSONPBMarshaler` interface), the resulting JSON is not correctly indented. This is due to the place where the JSON is re-written to add the `@type` attribute not correctly respecting the `Indent` field.


jhump:
@cybrcodr, @dsnet: It looks like I failed to account for indentation/pretty-printing when I landed #361 way back when. I now have need for this code path to generate properly indented JSON. This PR is the fix.

I know you may be trying to limit the number of patches to the v1 API. But this is a bug-fix, and it's pretty simple and low risk. Hopefully it passes the bar for v1 pull requests.

dsnet:
This change is sufficiently simple that we can probably merge it into v1. I'll let @cybrcodr make the call. He's become the de-facto export on textpb and jsonpb related functionality.

PR #844, opened 2019-05-02
Ability to specify proto_import_prefix for imported .proto file exclusively.

MartinKosicky:
Currently you can use only import_prefix=xxx/ that gets applyied to core imports like: github.com/golang/protobuf/proto

What this commit is about is that it allows you to specify import for .proto imports only (ie. one .proto file is referencing another) without changing the import prefix of core packages (ie. github.com/golang/protobuf/proto)

This commit does not affect previous functionality. If you set only import_prefix it will still work as before.

This commit is helpful if you want to generate into other folders than $GOPATH/src and dont want to use go_package. If you want to generate into $GOPATH/src/my_project/proto_generated for example

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,8 @@ The following parameters are deprecated and should not be used:
 
 - `import_prefix=xxx` - a prefix that is added onto the beginning of
   all imports.
+- `proto_import_prefix=xxx` - a prefix that is added onto the beggining of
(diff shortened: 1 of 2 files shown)


MartinKosicky:
Added CLA

dsnet:
\cc @neild

MartinKosicky:
CLA added

MartinKosicky:
testBot

MartinKosicky:
I signed it!

PR #845, opened 2019-05-02
Ability to specify proto_import_prefix for imported .proto file exclusively.

MartinKosicky:
Currently you can use only import_prefix=xxx/ that gets applyied to core imports like: github.com/golang/protobuf/proto

What this commit is about is that it allows you to specify import for .proto imports only (ie. one .proto file is referencing another) without changing the import prefix of core packages (ie. github.com/golang/protobuf/proto)

This commit does not affect previous functionality. If you set only import_prefix it will still work as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,8 @@ The following parameters are deprecated and should not be used:
 
 - `import_prefix=xxx` - a prefix that is added onto the beginning of
   all imports.
+- `proto_import_prefix=xxx` - a prefix that is added onto the beggining of
+  only those imports that come from one .proto file including another. Useful if you don't want to use go_package option and want to generate all files under a custom location.
 - `import_path=foo/bar` - used as the package if no input files
   declare `go_package`. If it contains slashes, everything up to the
   rightmost slash is ignored.
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -408,6 +408,7 @@ type Generator struct {
 	Param             map[string]string // Command-line parameters.
 	PackageImportPath string            // Go import path of the package we're generating code for
 	ImportPrefix      string            // String to prefix to imported package file names.
+	ProtoImportPrefix string            // String to prefix to imported package file names. This applyies when proto includes another proto file. Not core source code.
 	ImportMap         map[string]string // Mapping from .proto file name to import path
 
 	Pkg map[string]string // The names under which we import support packages
@@ -479,6 +480,11 @@ func (g *Generator) CommandLineParameters(parameter string) {
 		switch k {
 		case "import_prefix":
 			g.ImportPrefix = v
+			if g.ProtoImportPrefix == "" {
+				g.ProtoImportPrefix = v
+			}
+		case "proto_import_prefix":
+			g.ProtoImportPrefix = v
 		case "import_path":
 			g.PackageImportPath = v
 		case "paths":
@@ -1320,7 +1326,7 @@ func (g *Generator) generateImports() {
 	g.P(g.Pkg["math"] + ` "math"`)
 	g.P(g.Pkg["proto"]+" ", GoImportPath(g.ImportPrefix)+"github.com/golang/protobuf/proto")
 	for importPath, packageName := range imports {
-		g.P(packageName, " ", GoImportPath(g.ImportPrefix)+importPath)
+		g.P(packageName, " ", GoImportPath(g.ProtoImportPrefix)+importPath)
 	}
 	g.P(")")
 	g.P()


neild:
What is the motivation for adding a prefix to import paths? What, specifically, are you trying to do?

I don't think I've seen a case where import_prefix was the right answer, as opposed to some combination of paths=source_relative, M, and/or go_package options.

MartinKosicky:
Suppose I have  a large git repository (historically we don't have the 1 repo per project) that was created mainly for C++ projects. Now we wanted to create a microservice using existing protobufs with a structure like this (have the ability to have project sctructure i want is my goal actually):

```  src
      Common
           Definitions
               Tasks
                    task.proto
              Triggers
                    trigger.proto
          Microservices
               Service1
                    request.proto    -- this references some of the trigger.proto task.proto
        Microservices  
            Service1
                 Chart
                 Pipeline
                 src
                      service
                           vendor
                              github.com
                                  golang
                                      protobuf
                                         ... 
                           pkg
                                protobufs
                                    -- I want to generate protobufs here
                           cmd
                           Gopkg.toml
                           Gopkg.lock
```

When we generate protobufs we use the proto_path as the src/Common.  For the sake of having generated files in this new golang project in a single folder (so that writing .gitignore is easier and to have it one place) I want to generate the source code into src/Microservices/Service1/src/service/pkg/protobufs.  Since the generated imports are acting like from proto_path I cannot use this. I have all the protoc-gen-go in vendor folder and also other protobuf core. i can't use multiple GOPATHS because the second GOPATH doesn't search another ones' vendors. I don't want to generate into vendor folder because I can't make a vendor/protobufs because of the same problem and i don't want to generate directly into vendor.  I don't want to come up with a valid package name for DEP packaging so that it would work in vendor and the number of proto files is too large for this.  From user experience allowing user to generate protobufs where he wants and use the standard .proto schema without intrusing the .proto files with go_package options is in my oppinion a lot straight forward.

puellanivis:
> writing .gitignore is easier

All files generated by the protoc Go compiler match `*.pb.go`

> … without intrusing the .proto files with go_package options …

This was by design. The protobuf should specify the necessary options for all intended cross-platform users of the protobufs defined.

MartinKosicky:
> This was by design. The protobuf should specify the necessary options for all intended cross-platform users of the protobufs defined.

Ehmmm, so do you think that having ability to make some global configuration for code generation is a bad thing (Regardless of what the design is) ? Because I read that protobuf is meant as a language neutral format (in terms of wire format and also code generation). Also because there are already ways to make the global configuration as @neild mentioned by using the paths=source_relative, M.  However this thing would require me to make a code generator that would generate a script that would invoke the protoc-gen-go generator (because our protobufs import list may change).
Also please note that not all protobufs I am using are under my control and the authors of those protobufs don't care what language I will be using

puellanivis:
So, there is a complication here that protobufs are not used “in the wild” the same way that they are in Google. In Google, protobuf code generation (except for Java) goes into the same path as the `.proto` file, and it is built new every time. And it is understood that any protobuf supporting language might import your proto.

But I don‘t need to tell you that that just does not work outside of the constrained environment provided by Google’s architecture. There are two solutions of course, rework the way you work, or build the use case into the wider available tools. You‘re clearly working towards the later, so that you can “just get your work done”, but there’s also complications of this complicating use cases, and is now Yet Another Weird Corner Case that has to be accounted for, and carried forward under all cases.

I guess, that I‘m sorry I don’t have good answers. I’m trying to provide information about “how things should be” but reality does not always let that be.

MartinKosicky:
> but there’s also complications of this complicating use cases, and is now Yet Another Weird Corner Case 
Why do you consider this another weird corner case? Protoc allows for --go_out to specify where do you want to build, but if you don't specify the "correct" path it won't work. What I purpose is a generic method to "correct" the generated code so no matter where you generate it would work. 

Maybe we can discuss in what case (where you don't want to configure using the go_package) will not this thing suffice? Or possibly cause complications?

dsnet:
We're moving to a world where either generating Go code from .proto files must either: 1) have a `go_package` option specified or 2) have a set of `M` flags passed to `protoc-gen-go`.

The addition of this option works against that goal. As such, we won't be adding this feature.

MartinKosicky:
Well your goal violates open closed principle and is intrusive (1), which is a sign of bad design. And second goal is user unfriendly because it requires additional code generation to create the protoc-gen-go commands. However if that is the goal then i will live with it.

dsnet:
This is the same approach as used by other language implementations for protobufs. The Go implementation does not aim to be special.

PR #860, opened 2019-05-27
Feature/table unmarshal fix

wpr7280:
Change-Id: I34dceb6109b8ff0893836855760bb8245dd0293f

In table_marshal.go file, 
`	tags := strings.Split(f.Tag.Get("protobuf"), ",")
	if tags[0] == "" {
		return
	}
`
but in table_unmarshal.go, don't consider this condition.

when i use  `protobuf:""`  to ignore some field, serialization is normal, but it will fail when deserialized.


dsnet:
I'm not sure why this PR is surfaced now. Is there something to look at here?

dsnet:
I'm not sure why this PR is surfaced now. Is there something to look at here?

dsnet:
Closing as obsolete.

PR #861, opened 2019-05-27
table unmarshal fix

wpr7280:
Change-Id: I7b2d8e79edf41d4d6085ac3c43a9646b527ee197

In table_marshal.go file,
tags := strings.Split(f.Tag.Get("protobuf"), ",") if tags[0] == "" { return } 
but in table_unmarshal.go, don't consider this condition.

when i use protobuf:"" to ignore some field, serialization is normal, but it will fail when deserialized.


wpr7280:
I signed it!

dsnet:
Closing as obsolete.

Issue #867, opened 2019-06-11
APIv2: track breaking changes


dsnet:
This is a meta-issue for tracking any breaking changes that occur before the final v2 release. Since v2 is not yet released, we reserve the right to make breaking changes to improve the API to be as good as it can be.

The CL (changelist) that introduces a breaking change will be posted here as a comment. Subscribe to this issue to be notified of breakages. We try to batch multiple breaking changes close together in time to reduce user churn.

robbertvanginkel:
Is there any rough idea of when api-v2 will become stable/available/recommended?

dsnet:
We're aspirationally aiming for GopherCon 2019, which would be the end of July. However, this is aspirational. If the v2 is not where we would like it to be (i.e., incomplete, buggy, etc...), then we will delay. We estimate that the amount of work left is on the order of months, not years.

dsnet:
Breaking changes:
* [CL/182537](https://golang.org/cl/182537): reflect/protoregistry: remove Files.RangeFilesByPath method
* [CL/175987](https://golang.org/cl/175987): reflect/protoreflect: remove bool output for Descriptor.Parent method
* [CL/176942](https://golang.org/cl/176942): reflect/protoreflect: remove FieldDescriptor.{Oneof,Extendee} methods
* [CL/174938](https://golang.org/cl/174938): reflect/protoreflect: remove {Enum,Message}.Type methods
* [CL/182597](https://golang.org/cl/182597): reflect/protoreflect: remove KnownFields and UnknownFields interfaces

dsnet:
Minor change:
* [CL/182638](https://go-review.googlesource.com/c/protobuf/+/182638): reflect/protodesc: use Resolver interface

neild:
Exceedingly minor change:

* [CL/182980](https://golang.org/cl/182980): encoding/proto{json,text}.MarshalOptions.Resolver must also resolve extensions

dsnet:
Minor change:
* [CL/183101](https://go-review.googlesource.com/c/protobuf/+/183101): internal/fileinit: remove package

This only affects generated code; the fix is to regenerate the `.pb.go` files.

dsnet:
Breaking change:
* [CL/183700](https://go-review.googlesource.com/c/protobuf/+/183700): remove Files.Find{Enum,Message,Extension,Service}ByName

neild:
Breaking change:

[...]

PR #870, opened 2019-06-17
add Number method to enum

timestee:
related issue: https://github.com/protocolbuffers/protobuf/issues/6104#issuecomment-502423539

if enum has `Number` method. Though we define `ErrorCode ` in different proto files. We could solve the problem like this:

```golang
type ProtoEnum interface {
	fmt.Stringer
	Number() int32
	EnumDescriptor() ([]byte, []int)
}

func Errorf(v ProtoEnum) error {
   // now we could get the code and message desc.
}
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1419,6 +1419,11 @@ func (g *Generator) generateEnum(enum *EnumDescriptor) {
 	g.P("}")
 	g.P()
 
+	g.P("func (x ", ccTypeName, ") Number() int32 {")
+	g.P("return ", "int32(x)")
+	g.P("}")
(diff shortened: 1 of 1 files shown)


dsnet:
This is fixed in v2 with the new [`protoreflect.Enum` interface](https://github.com/golang/protobuf/blob/5b6d0471e5bd463898a8121123ace60c7a0d9ca7/reflect/protoreflect/value.go#L9-L22) where the `Number` method returns a `protoreflect.EnumNumber`.

timestee:
thanks @dsnet  v2 is stable now ? could use it in production?

dsnet:
It is not stable. If you're interested in tracking breaking changes, subscribe to #867.

PR #871, opened 2019-06-17
added v2 suffix to go.mod module name

donaldgraham:
The v2 suffix was dropped in commit 934520f5b.  Was this intentional?


donaldgraham:
I've signed the CLA.

dsnet:
Dropping the v2 was intentional. Go modules does not require a major version change to have a `v2` suffix. The only requirement is that the import path be different from previous major versions.

Issue #745, opened 2018-11-07
jsonpb: does not support FieldMask


nodirt:
Go implementation of JSONPB does not support well-known proto type [FieldMask](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/field_mask.proto), which has [special JSON encoding](https://github.com/protocolbuffers/protobuf/blob/cd95c791497747043200755a4e45e1ead98e85b3/src/google/protobuf/field_mask.proto#L180).

**What version of protobuf and what language are you using?**
Version: ddf22928ea3c56eb4292a0adbbf5001b1e8e7d0d

**What did you do?**

```go
package main

import (
	"fmt"

	"google.golang.org/genproto/protobuf/field_mask"

	"github.com/golang/protobuf/jsonpb"
)

func main() {
	fm := &field_mask.FieldMask{Paths: []string{"a", "b"}}
	m := &jsonpb.Marshaler{}
	marshaled, _ := m.MarshalToString(fm)
	fmt.Println(marshaled)
	// expected: "a,b"
	// actual: {paths: ["a", "b"]}

	fm = &field_mask.FieldMask{}
	marshaled = "a,b"
	err := jsonpb.UnmarshalString(marshaled, fm)
	fmt.Println("error:", err)
}
```
**What did you expect to see?**

```
"a,b"
nil
```

**What did you see instead?**

```
{"paths":["a","b"]}
error: invalid character 'a' looking for beginning of value
```

dsnet:
Thanks for filing. I marked with with "api-v2" since I rewrote the whole `jsonpb` package and the new implementation handles this.

See https://go-review.googlesource.com/c/protobuf/+/134495/1/encoding/jsonpb/known.go#94

AlexAPB:
Can we get a status update on this? I think this may be better suited under a `bug` rather than an `enhancement` request.

cybrcodr:
This is implemented in v2, which we're not quite ready for a release yet.  v1 will get the change when we update it to use v2 underneath.  The release may take a while though.

AlexAPB:
Do you have any public documents discussing v2 timelines/details?

cybrcodr:
https://github.com/golang/protobuf/issues/364#issuecomment-384123374 provides links to v2 design details.  We don't have published timelines.  You can follow updates on our repo at https://go.googlesource.com/protobuf.

dsnet:
[...]

PR #874, opened 2019-06-20
Add optional support for FieldMask wkt

gtungatkar:
This fixes https://github.com/golang/protobuf/issues/745.

Go implementation of JSONPB does not support well-known proto type FieldMask, which has special JSON encoding. 
The spec is defined as:
`
mask {
  paths: "user.display_name"
  paths: "photo"
}
`
JSON:
`
{
  mask: "user.displayName,photo"
}
`
What jsonpb actually serializes it as:
` mask: {paths: ["user.displayName","photo"]}`

While this has been fixed in v2, v1 is still widely in use and v2 does not have published timelines.

**What does this pull request do**
This PR adds support for correctly serializing FieldMask. But since this is a breaking change, this fix is gated by an env variable. 
Protobuf implementations in other languages like python/cpp already support FieldMask. This allows golang jsonpb generated JSON to be compatible.


dsnet:
I appreciate the effort you put into this.

Unfortunately, we cannot accept this change for several reasons:
* As you noted, v2 already implements this and we will be re-writing all of v1 to use v2 under the hood in the near future. We're aspirationally targeting a v2 release by the end of July. We have no intentions of adding new features to v1.
* This re-generates the `FieldMask` proto in this repository, which is going to cause registration issues with the [current location of the proto](https://godoc.org/google.golang.org/genproto/protobuf/field_mask). We intend to move `FieldMask` to be under our ownership, but it's going to require coordination with `google.golang.org/genproto`.

Tatskaari:
Pragmatically speaking protobuf has a well defined specification for serialising to json for a reason. This specification was necessary was to facilitate cross language serialisation. This is the most widely used implementation and is non-compliant. Once V2 is released this will still be the most widely used implementation for a good while. Leaving the v1 version of the API broken doesn't seem like the right thing to do.

I've created a similar PR (#881) that doesn't vendor the FieldMask into the repo. Let me know if there are any reasons blocking that being merged in.

dsnet:
> Leaving the v1 version of the API broken doesn't seem like the right thing to do.

We have no intentions of leaving v1 broken. When v2 is finished, v1 will actually call v2 under the hood, implicitly pulling in the fix from v2.

> vendor the FieldMask into the repo

We still can't accept that approach. `golang/protobuf` cannot have a dependency on `genproto`.

Overall, I acknowledge that the current situation stinks and that v1 is broken (it's actually broken in a lot of ways which is precisely why we're building v2). Focusing all of our efforts on v2 is probably the best course of action.

PR #881, opened 2019-06-27
Make jsonpb serialise FieldMask messages according to the proto json spec

Tatskaari:
The protobuf json specification has a special case for FieldMask messages that jsonpb doesn't account for. See issue #745.

For reference jsonpb was producing json like 
```json
"mask": {
    "paths": [
      "field1",
      "field2",
      "field3"
    ]
  }
```
whereas the spec calls for:
```json
"mask": "field1,field2,field3"
```

The specification can be found here, note the entry for FieldMask at the bottom of the table: https://developers.google.com/protocol-buffers/docs/proto3#json 

This issue is fixed in APIv2 however there's no ETA on that release. This PR handles the FieldMask messages correctly for the time being.


Tatskaari:
I signed it!

dsnet:
Hi, thank you for your work here.

Unfortunately, we are no longer accepting feature changes to v1 as the v2 is the primary focus. It is stated for release within the next month or two, and this issue is fixed in v2.

martinxsliu:
Just ran into this issue myself. It turns out the Ruby proto library will cause a Ruby kernel panic when it encounters an ill formed field mask.

First of all, thank you @dsnet and team for the work you all have put into this library and into v2. It's been a few months and from looking at the v2 blocking issues, it appears we still have some time before release. Given this, would you consider merging this PR or a suitable fix for this bug until v2 is ready?

PR #916, opened 2019-07-31
jsonpb: add option to marshal int64 types as unquoted numbers

zaquestion:


dsnet:
The `jsonpb` package is intended to the be faithful implementation of the proto<->JSON specification, for which the C++ implementation is considered the canonical "reference" implementation.

The [C++ implementation only has small set of options](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.util.json_util#JsonPrintOptions) and we (the Go implementors of protobuf) have no intention of adding features that are beyond what C++ provides. The `EmitDefaults` option in Go is analogous to the `always_print_primitive_fields` option in C++. The other options that Go currently has are a historical mistake.

You're welcome to petition for this option in the reference implementation at https://github.com/protocolbuffers/protobuf/issues

PR #929, opened 2019-08-20
generator: use std to identify keywords and reserved identifiers

a8m:


dsnet:
Thanks for the PR. However, the v1 generator is frozen as all work is now focused on v2, which addresses this issue: https://github.com/golang/protobuf/blob/0080e68288c78240ab463b5285a2e9ef77d9b66c/compiler/protogen/names.go#L65-L71

PR #932, opened 2019-08-27
.travis.yml: include go1.12

thaJeztah:
noticed this was a bit behind, and go 1.13 is around the corner


thaJeztah:
hm.. that's confusing; so is this repository meant to be archived, or a mirror? Is that also the import-path to use for Go?

PR #933, opened 2019-08-28
generator: add `no_omitempty`option

ysugimoto:
This PR adds `no_omitempty` option for `protoc-gen-go`.

I would suggest that when I implement HTTP API using gRPC backend, and respond to a client as JSON, we'd keep Go's Zero Value field without omiting.

Particularly when a client uses TypeScript, might be omitted field have to define as nullable type. so I'd like to choose to keep that field is omitted or not by applying `no_omitempty` option in `protoc-gen-go`.

```
// with 'omitempty' as default, so this PR doesn't break current behavior
protoc .. --go_out=plugins=grpc:.

// without 'omitempty' tag
protoc ... --go_out=plugins,no_omitempty:.
```

But I think there's room for considering argument name, Thanks


ysugimoto:
@googlebot I signed it!

PR #936, opened 2019-08-30
proposal: protoc-gen-go: Add `OneOfValue()` next to oneof discriminator method.

ZjMNZHgG5jMXw:
Simplifies the access to fields that are common among all oneof subtypes
by means of casting `interface{}` returned by `OneOfValue()` to a more
specific interface containing a common getter method. Avoids the effort
of creating a switch case for each oneof subtype and its specific field.

Example
```
message A {
  oneof B {
    C c = 1;
    D d = 2;
  }
  message C {
    Common common = 1;
    SpecificForC x = 2;
  }
  message D {
    Common common = 1;
    SpecificForD y = 2;
  }
  message Common {
    string name = 1;
  }
  [...]
}
```

I would like to (manually) add a general `GetCommon()` method for `B` in a file next to the `pb.go` file, like this.
```
type getCommoner interface {
  GetCommon() (*A_Common)
}
func (m *A_B) GetCommon() (common *A_Common) {
  if commoner, ok := m.GetB().OneOfValue().(getCommoner); ok {
    common = commoner.GetCommon()
  }
  return
}
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1916,6 +1916,7 @@ func (f *oneofField) getter(g *Generator, mc *msgCtx) {
 	// The discriminator type
 	g.P("type ", f.goType, " interface {")
 	g.P(f.goType, "()")
+	g.P("OneOfValue() interface{}")
 	g.P("}")
 	g.P()
 	// The subField types, fulfilling the discriminator type contract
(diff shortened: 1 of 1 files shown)


ZjMNZHgG5jMXw:
@googlebot I signed it!

dsnet:
Thanks for the PR. However:
* Development of the v1 `protoc-gen-go` has come to end. This codebase is close to frozen except for bug fixes. All new work is in [`google.golang.org/protobuf`](https://google.golang.org/protobuf). In particular, the new generator is at [`google.golang.org/protobuf/cmd/protoc-gen-go`](https://godoc.org/google.golang.org/protobuf/cmd/protoc-gen-go).
* This feature is heavily dependent on a properties that the protobuf language itself does not have. For example, the intent is that message `A.C` and `A.D` "inherit" from message `A.Common`. Assuming that the members of a oneofs all "implement" `A.Common` then you're proposing that it should provide an accessor method to access the common parent. However, inheritance is not a concept in the protobuf language. The Go implementation of protobufs is intended to be a faithful representation of the protobuf language and has no interest in developing new features. If you want to propose new protobuf language features, please do so here: https://github.com/protocolbuffers/protobuf/issues.

PR #941, opened 2019-09-07
protoc-gen-go/grpc: Add null guard for gRPC connection

yagotome:
Generated code for gRPC Service Client does an RPC call by deferencing a connection pointer (`grpc.ClientConn`) without checking whether it is nil.

In this PR, an error is returned if that connection pointer is nil, so that no _nil pointer dereference_ is thrown (_panic_) due to nil connection.


yagotome:
Closing in favor of #942

PR #942, opened 2019-09-07
protoc-gen-go/grpc: Add null guard for gRPC connection

yagotome:
Generated code for gRPC Service Client does an RPC call by deferencing a connection pointer (`grpc.ClientConn`) without checking whether it is nil.

In this PR, an error is returned if that connection pointer is nil, so that no _nil pointer dereference_ is thrown (_panic_) due to nil connection.

### DIFF
--- a/protoc-gen-go/grpc/grpc.go
+++ b/protoc-gen-go/grpc/grpc.go
@@ -329,11 +329,14 @@ func (g *grpc) generateClientMethod(servName, fullServName, serviceDescVar strin
 	methName := generator.CamelCase(method.GetName())
 	inType := g.typeName(method.GetInputType())
 	outType := g.typeName(method.GetOutputType())
+	statusPkg := string(g.gen.AddImport(statusPkgPath))
+	codePkg := string(g.gen.AddImport(codePkgPath))
 
 	if method.GetOptions().GetDeprecated() {
 		g.P(deprecationComment)
 	}
 	g.P("func (c *", unexport(servName), "Client) ", g.generateClientSignature(servName, method), "{")
+	g.P("if c.cc == nil { return nil, ", statusPkg, `.Errorf(`, codePkg, `.FailedPrecondition, "grpc connection is nil") }`)
 	if !method.GetServerStreaming() && !method.GetClientStreaming() {
 		g.P("out := new(", outType, ")")
 		// TODO: Pass descExpr to Invoke.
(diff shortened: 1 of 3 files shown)


dsnet:
\cc @dfawley

puellanivis:
My concern about checking it at every function call is less concerned with performance. A smart compiler should be able to elide the code where the condition is never possible. And actually, I’m pretty sure, the compiler conditionally generates code to test for `nil` in order to implement the panic, anyways.

My concern is centrally that it should be a programming fault for such an error path to ever even be possible. Safety guards can be good, but runtime panics can sometimes be the correct choice. :woman_shrugging:

yagotome:
@dfawley alright! I agree that introducing a new `panic` is harmful, that's why I am not proposing that.

> @yagotome, this use case is exactly how gRPC clients are _supposed_ to work (but not how grpc-go works, unfortunately). The initial state of the client is IDLE until the first RPC is attempted, at which point the client starts connecting. Updating Go to this behavior is tracked here: [grpc/grpc-go#1786](https://github.com/grpc/grpc-go/issues/1786).

I wasn't aware of that open issue. That's exactly what I would like to have :slightly_smiling_face: 
However, would it be possible that feature is implemented in v1? Otherwise, i.e. if this is coming in v2, is v1 going to keep this current `panic` behavior? :confused: 

> However, I don't see how checking for nil ahead of time vs. for a special error after helps save any code or complexity.

The scenario is that if I tried to connect before each RPC, I would need to check for `err != nil` twice (one for connection error and other for RPC's).

> Also, the error code used, `FailedPrecondition` is explicitly called out in our spec as never produced by grpc:

My fault! If we go ahead with this PR, I will replace with a proper error.

> IMO we should leave things as they are, and work toward supporting _idleness_ in grpc-go instead.

Totally agree! Let's focus on supporting idleness in grpc-go!

That said, what about v1/v2's question? Should that `panic` be kept while _idleness_ is not released?

dsnet:
@dfawley may correct me here, but I'm not aware of a v2 effort for gRPC.

The discussion about v2 on this repository is specifically about the Go protobuf implementation, which does not cover the grpc generator. The fact that the grpc generator is part of this repository is more for historical reasons and we intend for it to eventually be part of the grpc-go repository in the future.

dfawley:
@dsnet, correct, we have no plans for a gRPC-Go v2.

@yagotome while we can't change the default behavior of `grpc.Dial`, we intend to add `grpc.NewClient` that will have this idleness behavior (in 1.x).  For now I do think we should leave the current panic in place, as explained in the comment above.

yagotome:
@dfawley @dsnet I agree with you! It's better keeping current `panic` since `grpc.NewClient` will have that idleness behavior soon.

Closing this PR therefore.

PR #952, opened 2019-09-20
protoc-gen-go: fix deprecated comment for grpc methods

menghanl:
The deprecated comment was only added to the method of an unexported type.

https://github.com/grpc/grpc-go/issues/3023


menghanl:
cc @dfawley

dfawley:
LGTM

PR #969, opened 2019-10-15
protoc-gen-go: allow specifying additional internal fields tags

aconchillo:
Add a new plugin option `internal_tags` to allow specifying additional
field tags. The `json:"-"` tag is maintained and the new tags will simply
be appended.

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -409,6 +409,7 @@ type Generator struct {
 	PackageImportPath string            // Go import path of the package we're generating code for
 	ImportPrefix      string            // String to prefix to imported package file names.
 	ImportMap         map[string]string // Mapping from .proto file name to import path
+	InternalTags      string            // String with additional internal field tags.
 
 	Pkg map[string]string // The names under which we import support packages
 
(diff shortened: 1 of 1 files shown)


aconchillo:
This change allows to pass, for example, `--go_opt='internal_tags=gorm:"-"'` in case one want to use protobuf types with GORM.

Note that for regular fields one can use https://github.com/favadi/protoc-go-inject-tag but with that it's not possible to also update internal fields.

As an alternative to this, a `sed` command can be used, but this seems a cleaner and safer way to do it.

puellanivis:
As I understand this, this is solely to add additional tags to `XXX_` fields in order to exclude them similar to the way that `json:"-"` ensures that the fields are excluded when `json.Marshal`ling them.

Beyond not being sure about the real need for this, I have concerns over the code, predominantly in the way of invalid input handling: what happens if you pass a backtick in the argument, what happens if I include invalid tag text, etc.

Perhaps it might be better to modify the `protoc-go-inject-tag` code, rather than the baseline tool?

aconchillo:
Yes, this is to add additional tags to `XXX_` fields.

A bit more about my use case. I use protobuf types both with gRPC and with GORM. This is so I only need one type definition instead of converting things from the protobuf type and another type back and forth. In order to do this, I use `protoc-go-inject-tag` to inject `gorm` tags to protobuf fields.

Why adding additional tags to `XXX_` fields? The reason is that at some point I performed a `reflect.DeepEqual` of two identical structures (at least that's what I expected), one was filled with GORM and the other was not. `reflect.DeepEqual` returned `false` because GORM initializes structure fields by default (for fields that are not ignored). Since the `XXX_` were not ignored they had non-nil values, however the other structure had nil values, therefore `reflect.DeepEqual` failed.

The solution is to simply add `gorm:"-"` to the `XXX_` fields. I guess this could be also done in `protoc-go-inject-tag` but since the  `XXX_` fields are generated by `protoc-gen-go` it felt safer to do it here (input validation could be improved to check for only valid tags).

aconchillo:
Updated PR with tag validation.

puellanivis:
> … because GORM initializes structure fields by default (for fields that are not ignored). Since the XXX_ were not ignored they had non-nil values…

The only value I see here of the `XXX_` that could even have a nil value is `"XXX_unrecognized []byte`. (One is a `struct{}` which cannot contain any value other than the zero value; one is an optionally embedded `XXX_InternalExtensions` value struct, which only has unexported values, so GORM shouldn’t be able to set any of the fields; and one is an `int32`)

Is GORM setting that `[]byte` to `[]byte{}` rather than `nil`? Because like… that would be bad behavior.

puellanivis:
That would be a “misbehavior” of GORM. The zero values of empty slices is `nil`, not a zero-length slice with an allocated backing array. And as well, there is nothing that _needs_ to be put into that field in order to make it usable, or to “initialize” it.

I think the better course of action here would be to submit a bugfix on GORM to stop this behavior.

dsnet:
The [protobuf compatibility agreement](https://github.com/golang/protobuf#compatibility) reserves the right to add/modify/remove `XXX_` fields and methods. Adding user-configurable knobs to alter internal implementation details is fundamentally going to be brittle.

For example, the v2 generator does not generate `XXX_` fields or method. In such a case, this option has no sensible behavior.

aconchillo:
ok, that makes sense. thank you for taking the time to look at this.

PR #977, opened 2019-10-28
protoc-gen-go/generator: removed superfluous case from switch statement

AdityaHarindar:
Removed an unused case, applied go fmt.


dsnet:
Thanks for the PR. The v1 generator is frozen as all development effort has moved to v2 (see the `api-v2` branch)

PR #985, opened 2019-11-11
Fix edge case, where JSONName is equal

AndreasBergmeier6176:
jsonpb handles json names specially. To keep this handling uniform, it is necessary to write JSONName in any case.

Example:
```
/* json=street_foo needs to be inserted, so that jsonpb.Marshaller.Marshal produced street_foo
   Omitting json means it will produce streetFoo
*/
string street_foo = 9 [json_name="street_foo"];
```

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -1552,7 +1552,7 @@ func (g *Generator) goTag(message *Descriptor, field *descriptor.FieldDescriptor
 			name = name[i+1:]
 		}
 	}
-	if json := field.GetJsonName(); field.Extendee == nil && json != "" && json != name {
+	if json := field.GetJsonName(); field.Extendee == nil && json != "" {
 		// TODO: escaping might be needed, in which case
 		// perhaps this should be in its own "json" tag.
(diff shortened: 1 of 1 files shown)


dsnet:
Thanks for the PR.

At the moment, the v1 generator is frozen. All new work is on v2.
That said, it seems that v2 already has similar logic: 
https://github.com/golang/protobuf/blob/fab1c8d9a3b0854d1041db56f7dd0ebe70cc9fcc/internal/encoding/tag/tag.go#L181

dsnet:
Actually, I just noticed this comment in v2:
https://github.com/golang/protobuf/blob/fab1c8d9a3b0854d1041db56f7dd0ebe70cc9fcc/internal/encoding/tag/tag.go#L182-L183

So there probably needs to be more investigation here.

I should note that the v2 codes doesn't even use the protobuf tags anymore, but relies directly on the FieldDescriptorProto information.

dsnet:
The `json` name tag is inconsistent throughout various versions of the `github.com/golang/protobuf` module. In the new protobuf API, the generated struct field tags are not even used anymore, so this has become a non-issue.

Issue #888, opened 2019-07-04
proto: Unmarshal takes more than 100 seconds at 22k QPS


Anteoy:
Is this normal performance?
mass goroutine are blocked at  table_unmarshal.go:117
unmarshalInfoLock.Lock().

```
207251 @ 0x43105f 0x441619 0x4415ef 0x44138d 0x474939 0xc9025a 0xc901de 0xc6ed8c 0xc6e8f1 0xce167c 0x45e8f1
#	0x44138c	sync.runtime_SemacquireMutex+0x3c							/home/zhoudazhuang/usr/local/go1.12.1/go/src/runtime/sema.go:71
#	0x474938	sync.(*Mutex).Lock+0x108								/home/zhoudazhuang/usr/local/go1.12.1/go/src/sync/mutex.go:134
#	0xc90259	acs/vendor/github.com/golang/protobuf/proto.getUnmarshalInfo+0x39			/home/zhoudazhuang/db11/jm/pro/src/acs/vendor/github.com/golang/protobuf/proto/table_unmarshal.go:117
#	0xc901dd	acs/vendor/github.com/golang/protobuf/proto.(*InternalMessageInfo).Unmarshal+0xdd	/home/zhoudazhuang/db11/jm/pro/src/acs/vendor/github.com/golang/protobuf/proto/table_unmarshal.go:59
#	0xc6ed8b	acs/vendor/github.com/golang/protobuf/proto.(*Buffer).Unmarshal+0x1eb			/home/zhoudazhuang/db11/jm/pro/src/acs/vendor/github.com/golang/protobuf/proto/decode.go:425
#	0xc6e8f0	acs/vendor/github.com/golang/protobuf/proto.Unmarshal+0x170				/home/zhoudazhuang/db11/jm/pro/src/acs/vendor/github.com/golang/protobuf/proto/decode.go:343
#	0xce167b	acs/comet/client.(*Client).HandleRequest+0xecb						/home/zhoudazhuang/db11/jm/pro/src/acs/comet/client/client.go:626
```

I use pb like that:
```
s := time.Now()
		err := pb.Unmarshal(req.Data, registerInfo)
		e := time.Now().Sub(s).Seconds()
		if e > 1 {
			fmt.Printf("pb unmarshal cost1: %+v\n",e)
}
```

I got : 
```
pb unmarshal cost1: 124.445243586
pb unmarshal cost1: 124.463055651
pb unmarshal cost1: 128.71240536
[...]

PR #1004, opened 2019-12-25
proto: optimize global (un)marshal lock using RWMutex

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

Thie PR Use `RWMutex` to optimize `getMarshalInfo` and `getUnmarshalInfo`, for these functions, only n (number of message type) will hit Write, and m(number of message) - n will hit Read, it's the best case to use RWMutex instead of Mutex.

This optimization introduce huge improvement in our scenario.

We have 1000 worker and 1 controller, and work and controller keep heartbeat with gRPC. They also exchange the job info with each other.

The message is like

```proto
message Job {
    uint64 job_id = 1;
    // Some other info
}

message Heartbeat {
    repeated Job jobs = 1;
}
```

About 10000 jobs in every Heartbeat, and the heartbeat QPS in controller is about 1000.

The controller handle the Heartbeat in about 10ms, and the network latency is about 10ms, but the client will use about 30s in maximum to finish a RPC call.

We use golang pprof block profile, and it seems that almost all block is caused by one global Mutex in protobuf package.

image

After optimization, in our use case, the rpc call from client will only use about 30ms, as our expected.

### DIFF
--- a/proto/table_marshal.go
+++ b/proto/table_marshal.go
@@ -106,11 +106,12 @@ func getMarshalInfo(t reflect.Type) *marshalInfo {
 		return u
 	}
 	marshalInfoLock.Lock()
-	// We don't need double check here.
-	// Only a little allocation will be introduced in parallel.
-	u = &marshalInfo{typ: t}
-	marshalInfoMap[t] = u
-	marshalInfoLock.Unlock()
+	defer marshalInfoLock.Unlock()
+	u, ok = marshalInfoMap[t]
+	if !ok {
+		u = &marshalInfo{typ: t}
+		marshalInfoMap[t] = u
+	}
 	return u
 }
 
--- a/proto/table_marshal.go
+++ b/proto/table_marshal.go
@@ -92,19 +92,24 @@ type marshalElemInfo struct {
 
 var (
 	marshalInfoMap  = map[reflect.Type]*marshalInfo{}
-	marshalInfoLock sync.Mutex
+	marshalInfoLock sync.RWMutex
 )
 
 // getMarshalInfo returns the information to marshal a given type of message.
 // The info it returns may not necessarily initialized.
 // t is the type of the message (NOT the pointer to it).
 func getMarshalInfo(t reflect.Type) *marshalInfo {
-	marshalInfoLock.Lock()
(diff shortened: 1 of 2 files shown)


TennyZhuang:
@dsnet PTAL

TennyZhuang:
OK, I've added a double check.

@puellanivis

TennyZhuang:
A simple reproduce code can be found at https://github.com/TennyZhuang/protobuf-lock-reproduce

The PR also resolve #888

TennyZhuang:
Any more reviewers?

dsnet:
`getMarshalInfo` is expected to be costly, but is only computed once per type in the program. After a program has been running for some time, it is no longer called. This PR is suggesting a solution in a situation where the root problem itself is not well understood.

TennyZhuang:
@dsnet in our scenario, I add counter log at the beginning of getMarshalInfo, and about 1000000 calls in 1 second. Is there some other bug to cause the function is called many times?

dsnet:
> Is there some other bug to cause the function is called too many times?

Possibly? That's the question that is more interesting to figure out. This code here:
https://github.com/golang/protobuf/blob/ed6926b37a637426117ccab59282c3839528a700/proto/table_marshal.go#L147-L159
atomically caches the computed `marshalInfo`, so it shouldn't happen again and again.

dsnet:
Have you used the most recent `protoc-gen-go` to generate your `.pb.go` files? The pprof graph you show shouldn't happen with newly generated `.pb.go` files.

TennyZhuang:
I reproduce it in https://github.com/TennyZhuang/protobuf-lock-reproduce (very high latency), protoc-gen-go 1.3.2.

I will try to inspect into it later.

TennyZhuang:
Sorry, the reproduce demo is not correct, I will try to create a more reproduce case.

TennyZhuang:
Sorry, this is a bug of https://github.com/gogo/protobuf/issues/656

dsnet:
Got it. I'm going to close this then. In v2, we use `sync.Map` to cache for the equivalent code, which scales much better as it reaches steady state.

PR #1018, opened 2020-01-16
generate json tags for oneof fields

galdorjt:
While protoc-gen-go generates JSON tags for most message structure members, it
does not do so when the member represent a oneof construct or when the
structure represent an element of a oneof member.

This means that programs using encoding/json instead of jsonpb will end up
with incoherent names which do not match the name of the Protobuf fields. They
will also be impredictable due to alternate names used to avoid collisions or
syntax errors.

Adding these tags makes the generated code coherent.


galdor:
@googlebot I signed it!

dsnet:
Hi, thanks for the PR, but we cannot accept this:
* The v1 `generator` is frozen and not accepting any changes except bug fixes or trivial documentation changes.
* The generation of the `json` tags on the struct as it actually does not work properly in certain cases. Even if you generate these on oneofs, it still does not enable the unmarsaler to handle them since it won't know what concrete type to use.

PR #1020, opened 2020-01-17
Int64 as int

sobolevsv:
add option to  render int64 values as integer values,  as opposed to string

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -652,14 +652,14 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 	if err != nil {
 		return err
 	}
-	needToQuote := string(b[0]) != `"` && (v.Kind() == reflect.Int64 || v.Kind() == reflect.Uint64)
-	if needToQuote {
-		out.write(`"`)
-	}
+	//needToQuote := string(b[0]) != `"` && (v.Kind() == reflect.Int64 || v.Kind() == reflect.Uint64)
+	// if needToQuote {
+	// 	out.write(`"`)
+	// }
 	out.write(string(b))
-	if needToQuote {
-		out.write(`"`)
-	}
+	// if needToQuote {
+	// 	out.write(`"`)
+	// }
 	return out.err
 }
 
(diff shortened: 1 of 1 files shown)


puellanivis:
This matter has been considered in other PRs: https://github.com/golang/protobuf/pull/916 https://github.com/golang/protobuf/pull/286 

The protobuf JSON mapping standard requires emission of `in64` values as quoted strings, in order to work around a well known complication of the JSON standard that there is no guarantee or requirement of accuracy for numbers beyond those afforded by `float64` type.

sobolevsv:
@googlebot I signed it!

gchaincl:
I tried to:

type MyObj struct {
	Id uint64 `protobuf:"varint,1,opt,name=id [...]"`
	[...]
}

```
func (obj *MyObj) MarshalJSONPB(m *jsonpb.Marshaler) ([]byte, error) {
	var buf bytes.Buffer

	alias := struct {
		*MyObj
		Id float64 `protobuf:"varint,1,opt,name=id [...]"`
	}{
		obj, float64(obj.Id),
	}
	if err := m.Marshal(&buf, alias); err != nil {
		return nil, err
	}

	return buf.Bytes(), nil
}
```

But jsonpb is calling MarshalJSONPB recursively, could you recomment other workaround?

puellanivis:
Your anonymous struct includes an embedded recursive `*MyObj` object pointer. When you try and marshal your anonymous alias, it has to marshal the embedded `*MyObj`, which will call your function, and wrap that `*MyObj` again, and recurse further.

I’m not particularly sure what you are hoping to accomplish with this design anyways. The `Id float64` will not supersede the `MyObj.Id` like this. Getting this code to “work” (i.e. not crash), it will just output: `{"MyObj":{"id":"42"},"id":42}` which I am pretty certain is not at all what you were hoping to achieve.

The problem here is that you are attempting to intentionally depart from the standards which this go package is defined by. In order to get your intended behavior, there is no support for this, and there is no work around with this library. You will need to either fork this library yourself in order to build in the feature you want (and then support it yourself), or find a fork which does that for you.

gchaincl:
thanks!

dsnet:
Thanks for the PR, but we can't accept this change. The behavior of `jsonpb` is dictated by the [protobuf<->JSON specification](https://developers.google.com/protocol-buffers/docs/proto3#json). The Go library team that maintains this project is not responsible for defining the specification, but only following it. Issues with the specification should be directed here: https://github.com/protocolbuffers/protobuf/issues

PR #1035, opened 2020-02-14
Add go modules installation instructions

johanbrandhorst:
This should help users make sure their
library and generator versions are kept in sync.


dsnet:
In the upcoming v1.4.0 release, we're going to be simplifying the README to just point to the protobuf developer guide. We're trying to reduce the number of redundant documentation and centralize on the developer guide. Unfortunately, the developer guide website is not open-source you can't easily contribute to it.

I'll keep the contents of this PR in mind and add something to the effect on the devsite.

PR #1038, opened 2020-02-17
Add typed error for non-linked in Any message, fixes #1037

### LINKED ISSUE #1037: types/known: typed error for non-linked in Any message
**Is your feature request related to a problem? Please describe.**
We use an `Envelope` `message` that has an `Any` field for holding another `message`. In our services whenever we add a new message we must go through each service and recompile our protos just so this new message is linked in, most of the time a new proto is ignored when switching on the type but an error is raised when we do `ptypes.UnmarshalAny` which happens before. At the moment, a non-typed error is returned: https://github.com/golang/protobuf/blob/d23c5127dc24889085f8ccea5c9d560a57a879d8/ptypes/any.go#L97

**Describe the solution you'd like**
A typed error that we can type switch against:
```go
type NonLinkedInMessageError struct {
	message string
}

func (e *NonLinkedInMessageError) Error() string {
	return fmt.Sprintf("any: message type %q isn't linked in", e.message)
}

...

if t == nil {
	return nil, &NonLinkedInMessageError{message: aname}
}
```

**Describe alternatives you've considered**
We can either painfully go through each service and update the protos or we can do a clunky string check on the error `strings.Contains(err, "isn't linked in")`.

**Additional context**
N/A

nylar:
Open to suggestions on the error name. Fixes #1037

### DIFF
--- a/ptypes/any.go
+++ b/ptypes/any.go
@@ -45,6 +45,15 @@ import (
 
 const googleApis = "type.googleapis.com/"
 
+// NonLinkedInMessageError respresents a message that has not been registered.
+type NonLinkedInMessageError struct {
+	message string
+}
+
+func (e *NonLinkedInMessageError) Error() string {
+	return fmt.Sprintf("any: message type %q isn't linked in", e.message)
+}
+
 // AnyMessageName returns the name of the message contained in a google.protobuf.Any message.
 //
 // Note that regular type assertions should be done using the Is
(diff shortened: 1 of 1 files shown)


dsnet:
Per https://github.com/golang/protobuf/issues/1037#issuecomment-587096838, we are no longer accept feature changes to v1.

PR #1055, opened 2020-03-12
go.mod: set go version to go1.9

neild:


puellanivis:
What does this even do?

neild:
Apparently causes test failures in modules that depend on this one when using older Go compiler versions.

To be honest, I don't understand the exact effects of the Go version statement in `go.mod` at all. However, since we have an explicit goal of supporting Go 1.9, setting the version to that seems reasonable.

PR #1056, opened 2020-03-13
proto: inline the implementation of protoimpl.X.ExtensionDescFromType

dsnet:


dsnet:
Technically, this differs from before in that it doesn't assign a nil `*ExtensionDesc` if the assertion fails. I don't know how much this matters.

PR #1060, opened 2020-03-20
Move Register${servName}Server to an interface

achew22:
This allows you to have a different implementation of `grpc.Server` which can be useful in testing.


dsnet:
Thanks for the PR, but the grpc plugin is no longer maintained by us. Please a file issue at https://github.com/grpc/grpc-go.

PR #1062, opened 2020-03-20
all: use google.golang.org/encoding/protowire

dsnet:
Use the exported protowire package instead of an internal fork.


dsnet:
Tests fail because of #1061

PR #1063, opened 2020-03-20
all: use google.golang.org/protobuf/testing/protopack for tests

dsnet:
Use protopack for tests instead of hard-coded binary data.


dsnet:
Tests fail because of #1061

PR #1066, opened 2020-03-28
jsonpb: `Marshaler.marshalValue()` incorrectly marshals nil slice elements

dragonsinth:
#1067 

This is against proto spec, and does not match the behavior of `proto.Marshal()`

### DIFF
--- a/jsonpb/jsonpb.go
+++ b/jsonpb/jsonpb.go
@@ -513,6 +513,9 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 		comma := ""
 		for i := 0; i < v.Len(); i++ {
 			sliceVal := v.Index(i)
+			if sliceVal.Kind() == reflect.Ptr && sliceVal.IsNil() {
+				return errors.New(fmt.Sprintf("repeated field %s has nil element", prop.Name))
+			}
 			out.write(comma)
 			if m.Indent != "" {
 				out.write("\n")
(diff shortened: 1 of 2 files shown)


dsnet:
Thanks for the PR, but the `jsonpb` implementation is no longer accepting changes. See my comment in #1067.

jsonpb/jsonpb.go
```diff
@@ -513,6 +513,9 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
 		comma := ""
 		for i := 0; i < v.Len(); i++ {
 			sliceVal := v.Index(i)
+			if sliceVal.Kind() == reflect.Ptr && sliceVal.IsNil() {
+				return errors.New(fmt.Sprintf("repeated field %s has nil element", prop.Name))
```
puellanivis: FYI: `fmt.Errorf("repeated field %s has nil element", prop.Name)`

dragonsinth: True, thanks!

dragonsinth:
Kind of a bummer this is already deprecated / no more bugfixes.  The new stuff doesn't seem completely ready / released yet?

- The new protojson library uses the new protoc-gen-go.
- But the new protoc-gen-go does not generate code for gRPC?
- protoc-gen-go-grpc has not landed yet?  grpc/grpc-go#3453

Are we missing something obvious?

cybrcodr:
It is possible to use the new `protojson` library with code generated from v1.4+ of protoc-gen-go.

neild:
And you can use `protojson` with code generated by older versions by using [`proto.MessageV2`](https://pkg.go.dev/github.com/golang/protobuf@v1.4.0-rc.4/proto?tab=doc#MessageV2) (available in `github.com/golang/protobuf` v1.4+) to convert an older message to the new interface.

PR #1071, opened 2020-04-02
[docs] remove references to import_path

drewgonzales360:
Since the `import_path` is marked as deprecated, I thought it'd be a good idea to take it out of the example command.


dsnet:
We're releasing v1.4.0 today, which is obsolete the need for this PR. Thanks for submitting it.

PR #1073, opened 2020-04-10
proto: add MarshalDeterministic; de-duplicate

karel-3d:
Currently, you have to use proto.Buffer for deterministic output;
let's make it a bit easier

Also de-duplicated the same logic copy-pasted twice

### DIFF
--- a/proto/table_marshal.go
+++ b/proto/table_marshal.go
@@ -2705,28 +2705,27 @@ func Size(pb Message) int {
 	return info.Size(pb)
 }
 
+func marshal(pb Message, deterministic bool) ([]byte, error) {
+	buf := NewBuffer(nil)
+	buf.SetDeterministic(deterministic)
+	err := buf.Marshal(pb)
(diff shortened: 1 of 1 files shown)


karel-3d:
@googlebot I fixed it.

neild:
Thanks for the contribution. We aren't making further API changes to this package, which has been superseded by the [`google.golang.org/protobuf/proto`](https://pkg.go.dev/google.golang.org/protobuf/proto) package. (See also [A new Go API for Protocol Buffers](https://blog.golang.org/protobuf-apiv2).)

The `google.golang.org/protobuf/proto` package does make this case simpler:

```go
err := proto.MarshalOptions{Deterministic: true}.Marshal(message)
```

karel-3d:
Oh.

Can you please link the new API in README, and/or add this to the default
Gihub message for PRs? :)

dsnet:
The README is already updated, but unfortunately has not been merged yet into master.

PR #1080, opened 2020-04-13
all: update to wrap google.golang.org/protobuf

neild:
All functionality is now implemented in terms of the next-generation protobuf API in 'google.golang.org/protobuf'.

For information on the new protobuf API, see:

  https://blog.golang.org/protobuf-apiv2

This commit merges the 'api-v2' branch with the 'master' branch.


neild:
> IIUC, this means we can also delete the `api-v1` branch and it won't break anyone depending on v1.4.0-rc.X already since those are tags to specific commits, right?

Correct.

PR #1114, opened 2020-05-04
all: upgrade to google.golang.org/protobuf@v1.22.0

dsnet:


neild:
Looks like `plugin.pb.go` needs to be regenerated.

dsnet:
I'll do that after the v1.22.0 release is actually tagged. There's a sequence actions.

PR #1125, opened 2020-05-11
jsonpb: fix a confusing error message

ijschwabacher:
Because `in` is a []byte, as written this prints a list of integers instead of the unrecognized enum value.


dsnet:
Bug fix approved since it fixes a regression relative to v1.3.5. See https://github.com/golang/protobuf/blob/84668698ea25b64748563aa20726db66a6b8d299/jsonpb/jsonpb.go#L905

PR #1129, opened 2020-05-14
proto: make InternalMessageInfo functional

dsnet:
The InternalMessageInfo type only exists to implement the
XXX methods on generated messages where those methods were
only ever intended to be called by this module itself.

Since v1.4.0, this module no longer relies on the XXX methods,
so the InternalMessageInfo and its implementation is supposed
to be dead code. Unfortunately, there are external usages that
violate our compatibility agreement and either directly call
the XXX methods or indirectly call it because some library
type-asserts to the existence of these methods.

This change adds minimal support for InternalMessageInfo by
just calling out directly to the v2 implementation.

### DIFF
--- a/proto/deprecated.go
+++ b/proto/deprecated.go
@@ -9,6 +9,8 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+
+	protoV2 "google.golang.org/protobuf/proto"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


proto/deprecated.go
```diff
@@ -85,8 +87,27 @@ func UnmarshalJSONEnum(m map[string]int32, data []byte, enumName string) (int32,
 // Deprecated: Do not use.
 type InternalMessageInfo struct{}
 
-func (*InternalMessageInfo) DiscardUnknown(Message)                        { panic("not implemented") }
-func (*InternalMessageInfo) Marshal([]byte, Message, bool) ([]byte, error) { panic("not implemented") }
-func (*InternalMessageInfo) Merge(Message, Message)                        { panic("not implemented") }
-func (*InternalMessageInfo) Size(Message) int                              { panic("not implemented") }
-func (*InternalMessageInfo) Unmarshal(Message, []byte) error               { panic("not implemented") }
+// Deprecated: Do not use.
```
RainbowMango, line 90: I suggest adding more detailed info about the deprecation. Include such as:
- why this would be deprecated?
- when this would be removed from codebase?
- how to migrate if the user happens using this?

RainbowMango: Same as below.

dsnet: I'd prefer not. I can mention that it's an internal type, but users really should not be looking at these.

PR #1199, opened 2020-09-03
proto.ExtensionDescs silently drops some extensions

jhump:
The `ExtensionDescs` is spec'ed to return "unknown extensions". The implementation will return unknown extensions from the `[]byte` unrecognized fields, but it just ignores unknown extensions that are actually in the extension map.

This fixes that so it correctly returns _all_ extensions, including all unknown ones.

### DIFF
--- a/proto/extensions.go
+++ b/proto/extensions.go
@@ -287,9 +287,8 @@ func ExtensionDescs(m Message) ([]*ExtensionDesc, error) {
 	mr.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
 		if fd.IsExtension() {
 			xt := fd.(protoreflect.ExtensionTypeDescriptor)
-			if xd, ok := xt.Type().(*ExtensionDesc); ok {
-				extDescs[fd.Number()] = xd
-			}
+			xd, _ := xt.Type().(*ExtensionDesc)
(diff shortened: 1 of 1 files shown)


proto/extensions.go
```diff
@@ -287,9 +287,8 @@ func ExtensionDescs(m Message) ([]*ExtensionDesc, error) {
 	mr.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
 		if fd.IsExtension() {
 			xt := fd.(protoreflect.ExtensionTypeDescriptor)
-			if xd, ok := xt.Type().(*ExtensionDesc); ok {
```
jhump, line 290: If the type is not `*ExtensionDesc`, that means it came from a different kind of extension type, such as a dynamic type that corresponds to an extension for which the generated code is not linked into the calling program.

jhump:
Hmm, never mind. This would require more extensive changes since `proto.GetExtension` would not work for these returned extensions...

I think this means I will have to go ahead and convert more of my code to API v2. I was trying to do that conversion incrementally, but things like this don't behave well when you have code in the same codebase using a mix of v1 and v2 APIs. In this case, I've moved over the code that _generated_ this message to use API v2, which is how it got such an extension in the extension map in the first place...

PR #1217, opened 2020-10-09
ptypes: deprecate the package

dsnet:
Deprecate the ptypes package since all the equivalent functionality
is now directly generated with the well-known types themselves.


puellanivis (approved):

neild (approved):

dsnet:
The crazy amount `go.sum` will be fixed in the future since `google.golang.org/protobuf@v1.26.0` will not have these entries.

PR #1300, opened 2021-03-22
jsonpb: bring back old behaviour of handling nulls and JSONPBUnmarshaler

krzysztofdrys:
Change how jsonpb.Unmarshal handles nested JSONPBMarshaler fields a null json values.

When json null is encountered for a field which implements JSONPBUnmarshaler, jsonpb will now calli unmarshal method from the field, instead of just assigning nil to this field.

### DIFF
--- a/jsonpb/decode.go
+++ b/jsonpb/decode.go
@@ -334,7 +334,7 @@ func (u *Unmarshaler) unmarshalMessage(m protoreflect.Message, in []byte) error 
 
 		field := m.NewField(fd)
 		// Unmarshal the field value.
-		if raw == nil || (string(raw) == "null" && !isSingularWellKnownValue(fd) && !isUnmarshalJSONPB(field, fd)) {
+		if raw == nil || (string(raw) == "null" && !isSingularWellKnownValue(fd) && !isSingularJSONPBUnmarshaler(field, fd)) {
 			continue
 		}
 		v, err := u.unmarshalValue(field, raw, fd)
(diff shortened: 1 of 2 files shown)


jsonpb/decode.go
```diff
@@ -390,6 +391,20 @@ func isSingularWellKnownValue(fd protoreflect.FieldDescriptor) bool {
 	return false
 }
 
+func isUnmarshalJSONPB(v protoreflect.Value, fd protoreflect.FieldDescriptor) bool {
```
dsnet, line 394: Please call this `isSingularJSONPBUnmarshaler` for consistency with `isSingularWellKnownValue` and please declare this function after `isSingularWellKnownValue`.

Also, please make the body of this similar to `isSingularWellKnownValue`:
```go
if fd.Message() != nil && fd.Cardinality() != protorflect.Repeated {
	_, ok := proto.MessageV1(v.Interface()).(JSONPBUnmarshaler)
	return ok
}
return false
```

jsonpb/decode.go
```diff
@@ -332,11 +332,12 @@ func (u *Unmarshaler) unmarshalMessage(m protoreflect.Message, in []byte) error
 			raw = v
 		}
 
+		field := m.NewField(fd)
 		// Unmarshal the field value.
-		if raw == nil || (string(raw) == "null" && !isSingularWellKnownValue(fd)) {
+		if raw == nil || (string(raw) == "null" && !isSingularWellKnownValue(fd) && !isUnmarshalJSONPB(field, fd)) {
```
dsnet, line 337: Make the equivalent change to L369 for extension fields.

krzysztofdrys:
Thanks for the review @dsnet! I think I have adjusted my PR according to your comments.

PR #1306, opened 2021-03-25
all: deprecate the module

dsnet:
Use the new deprecation feature to mark this module as deprecated.
See https://golang.org/issue/40357.

Considerations:
* google.golang.org/protobuf/cmd/protoc-gen-go@v1.25.0 and below
used to generate a hard dependency on github.com/golang/protobuf,
which would be frustrating since it would force an explicit dependency
on a deprecated module. However, that is no longer the case in v1.26.0.
* google.golang.org/protobuf and github.com/golang/protobuf have
a cyclic dependency on each other. However, this should not be a problem
since proposal 40357 only marks direct dependencies in the go.mod file,
rather than all transitive dependencies in the go.sum file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,4 @@
+// Deprecated: Use the "google.golang.org/protobuf" module instead.
 module github.com/golang/protobuf
 
 go 1.9


dsnet:
\cc @jayconrod

neild (approved):

jayconrod:
@dsnet I'm still implementing module deprecation, and this sounds like an interesting test case to add :)

Currently, the plan is to show deprecation messages in two scenarios:

* When `go list -m -u` is run to check for updates on a deprecated module; `go list -m -u all` would show this.
* When `go get` is run on a deprecated module or a package in a deprecated module needed to build one of the command line arguments.

Will users that depend on `google.golang.org/protobuf` end up transitively importing packages in `github.com/golang/protobuf`? If not, they'll probably only see a `(deprecated)` marker from `go list -m -u all`, which is probably fine.

If this ends up being too much of a nuisance, we could restrict deprecation warnings, e.g., to direct requirements only.

dsnet:
> Will users that depend on google.golang.org/protobuf end up transitively importing packages in github.com/golang/protobuf?

Nope. The dependency on on `github.com/golang/protobuf` from `google.golang.org/protobuf` is only a [weak module dependency](https://github.com/protocolbuffers/protobuf-go/blob/f821cae922ceaad0c2cb643f662901702e5cd79f/internal/weakdeps/weakdeps.go#L5-L14). There are no packages in `github.com/golang/protobuf`` that are imported under normal build tags.

PR #1341, opened 2021-07-19
comments for oneof subfields now pass to generated code

ShmilxZ:
Comments above oneof subfields will pass through to the generated code.


neild:
Thanks for the contribution!

This package is obsolete, deprecated, and we are no longer making changes to it. It is not used by the `protoc-gen-go` program. The current implementation of `protoc-gen-go` is located here: https://go.googlesource.com/protobuf/+/refs/heads/master/cmd/protoc-gen-go

Issue #1349, opened 2021-08-05
No plugins was loaded when not specify plugins from cli but from package import.


eval-exec:
**What version of protobuf and what language are you using?**
protobuf: 1.5.2
language: golang

**What did you do?**
I want to generate a filscanproto.proto from https://github.com/ipfs-force-community/filscan-backend/blob/master/filscanproto/filscanproto.proto with third party plugin: https://github.com/ipfs-force-community/gosf/tree/master/protoc-gen-force-jsonrpc

**What did you expect to see?**
 grpc and force-jsonrpc make effect

**What did you see instead?**
I didn't see grpc or force-jsonrpc related stuff was generated
```bash
cd ~/github.com/ipfs-force-community/filscan-backend/filscanproto
protoc -I . -I ../internal  --force-jsonrpc_out=.  filscanproto.proto
# then a filscanproto.pb.go was generated, but grpc and force-jsonrpc didn't make effect
```
**my fix**

I noticed the way I generate `pb.go` was not specify plugins from command line, but is import `grpc` and `force-jsonrpc`'s `	generator.RegisterPlugin(New())` will load custom plugins into the slice `plugins` : 

https://github.com/ipfs-force-community/gosf/blob/c7889dd90cf48e699a6d790a0f11933a78d5cc9c/plugin/jsonrpc/jsonrpc.go#L16

but in 
https://github.com/golang/protobuf/blob/ae97035608a719c7a1c1c41bed0ae0744bdb0c6f/protoc-gen-go/generator/generator.go#L499
 empty `nplugins` will cover non-empty `plugins` when `pluginList` is `"none"`.

neild:
It sounds like this is a problem with the `protoc-gen-force-jsonrpc` plugin. If there's a problem with that plugin's behavior, you should contact its owners for assistance.

The `github.com/golang/protobuf/protoc-gen-go/generator` package is deprecated. It was never designed for public use, and has been entirely superseded by the [`google.golang.org/protobuf/compiler/protogen`](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen) package. We are no longer making changes to the `generator` package.

eval-exec:
> It sounds like this is a problem with the `protoc-gen-force-jsonrpc` plugin. If there's a problem with that plugin's behavior, you should contact its owners for assistance.
> 
[...]

PR #1350, opened 2021-08-05
Fix  plugins cover by nil value

eval-exec:
if plugins was not specified from cli , but specify from third party package import, then empty `nplugins` will cover non-empty `plugins`, this is not correct.
#1349

### DIFF
--- a/protoc-gen-go/generator/generator.go
+++ b/protoc-gen-go/generator/generator.go
@@ -484,7 +484,7 @@ func (g *Generator) CommandLineParameters(parameter string) {
 			}
 		}
 	}
-	if pluginList != "" {
+	if pluginList != "none" && pluginList != "" {
 		// Amend the set of plugins.
 		enabled := make(map[string]bool)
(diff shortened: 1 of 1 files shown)


eval-exec:
> @googlebot I signed it!

protoc-gen-go/generator/generator.go
```diff
@@ -484,7 +484,7 @@ func (g *Generator) CommandLineParameters(parameter string) {
 			}
 		}
 	}
-	if pluginList != "" {
+	if pluginList != "none" && pluginList != "" {
```
neild, line 487: Thanks for the contribution.

The generator package is deprecated. While it exists in its current, frozen state to support existing programs that may depend on it, we are no longer making or accepting changes to it.

New code should use the [protogen](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen) package instead.

eval-exec: Thank you

PR #1413, opened 2022-02-15
delete useless code

ningmingxiao:
dm := m.(type) ,dm can not be nill

### DIFF
--- a/ptypes/any.go
+++ b/ptypes/any.go
@@ -47,9 +47,6 @@ func MarshalAny(m proto.Message) (*anypb.Any, error) {
 	case DynamicAny:
 		m = dm.Message
 	case *DynamicAny:
-		if dm == nil {
-			return nil, proto.ErrNil
-		}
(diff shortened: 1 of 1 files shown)


ptypes/any.go
```diff
@@ -47,9 +47,6 @@ func MarshalAny(m proto.Message) (*anypb.Any, error) {
 	case DynamicAny:
 		m = dm.Message
 	case *DynamicAny:
-		if dm == nil {
-			return nil, proto.ErrNil
-		}
```
puellanivis, line 52: Thanks for the contribution.

The ptypes package is deprecated. While it exists in its current, frozen state to support existing programs that may depend on it, we are no longer making or accepting changes to it.

I am also, unsure that this code is actually useless. As this code would panic with a nil-deref otherwise:
```
	var m *ptypes.DynamicAny
	ptytes.MarshalAny(m)
̀ ``

PR #1460, opened 2022-06-24
Show that oneof is broken

codygibb:
If a JSON object has two fields of the same oneof, one field
arbitrarily wins without any error.


puellanivis:
Properly, “If the parser encounters multiple members of the same oneof on the wire, only the last member seen is used in the parsed message.” https://developers.google.com/protocol-buffers/docs/proto3#json

If your purposed tests pass, the first one should be misbehavior? As the `"title"` comes second, and should overwrite the first one. However, the JSON mapping section does not really cover at all the behavior of oneof, so I’m just presuming the same behavior as we would expect from protobuf wire-format.

PR #1482, opened 2022-09-07
proto: remove redundant type conversion

cuishuang:


puellanivis:
The `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions.

cuishuang:
> The `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions.

Ok. Thanks

PR #1503, opened 2022-11-10
proto: make DebugPrint print unknown field

qilingzhao:
In `DebugPrint` method, byte slice `b` was set into Message's unknown fields, but the `MarshalOptions` 's field `EmitUnknown` use default value `false`. Actually, the method will print nothing about parameter `b`.


qilingzhao:
close, becase this module is frozen

PR #1585, opened 2024-01-02
Create FUNDING.ymls

Xela274444:


puellanivis:
Thank you for the attention, but the `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions.

PR #1637, opened 2024-08-05
Correctly advertise editions support

mkruskal-google:


puellanivis:
Thank you for the attention, but the `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions.

mkruskal-google:
I would actually label this a regression/bug fix.  AFAICT this plugin should support edition 2023 just fine because it uses the official golang protoc plugin infra, edition 2023 doesn't affect RPC codegen, and it already sets `FEATURE_SUPPORTS_EDITIONS` by copying `SupportedFeatures`.   It just fails to specify the edition window it supports, so anyone using this plugin will hit protoc errors when they try to upgrade to editions

puellanivis (approved):
A good point. 🤔 I do wonder though, how likely is it that people would still be using this `protoc-gen-go` rather than the newer one?

But still, also reasonable, and measured.

(Still needs a proper 👍 from someone with dev access.)

mkruskal-google:
FWIW I hit this trying to migrate someone to editions that was using this plugin :). I ended up just switching them to the newer one, but I figured there might be edge cases where that's not as much of an option.

jhump:
Should this PR be closed now?

stapelberg:
Sorry, I didn’t actually see this PR before. 

@jhump Why are you suggesting to close this PR? 

@mkruskal-google Are you still looking to get this merged?

jhump:
@stapelberg, it seemed like the conversation was suggesting that people that want to use Editions should migrate to the newer version of protoc-gen-go, over in [`google.golang.org/protobuf`](https://pkg.go.dev/google.golang.org/protobuf/cmd/protoc-gen-go).

PR #1717, opened 2026-06-24
ghoul: resource-leak - close gzip.Reader after decompressing descriptors

fereidani:
Hi,

I have built a static analyzer for Go called [Ghoul](https://fereidani.com/ghoul), and I ran it against the protobuf codebase. It flagged two resource leaks where a gzip.Reader used to decompress registered file descriptors is allocated but never closed.

Ghoul related reports:
```txt
proto/registry.go:36: [resource-leak] gzip.Reader from compress/gzip.NewReader never closed (confidence: 6.4/10, CWE-404)
descriptor/descriptor.go:128: [resource-leak] gzip.Reader from compress/gzip.NewReader never closed (confidence: 6.4/10, CWE-404)
```

This change adds `defer zr.Close()` right after the error check in `proto.RegisterFile` and `descriptor.deriveFileDescriptor`, so the reader is released once the descriptor bytes have been read. The decompression logic and its error handling are unchanged; the patch only adds the close.


puellanivis:
The `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions.

PR #1720, opened 2026-08-25
github: update github issue template with untriaged and issue type labels

zhangskz:
Applies labels untriaged, bug, feature request, question


aktau:
1.17.x and 1.18.x are failing due to the versions no longer being served from blob storage. But also, they are likely to fail due to these protobuf versions no longer being supported by the v2 repo on which this repo (v1) is based (being https://pkg.go.dev/google.golang.org/protobuf, which supports Go 1.23+).

Forwarding from chat:

> Given that v1 is built on v2 since v1-1.4, v1 requires at least a Go version supported by v2. v2 currently supports Go 1.23+ (can be changed to Go 1.24 now IIUC). 
>
> Thus, this matrix can be updated to [1.23.x, 1.24.x, 1.25.x, 1.26.x, 1.27.x]. Perhaps it would be better to do this in a separate PR that precedes this one.

I assume the matrix is stale because actual PRs to this repository (v1) are rare.

puellanivis:
Yeah, the `github.com/golang/protobuf` module is frozen except for essential bug fixes and regressions from one of the more recent versions. So, there hasn’t really been much attention paid on keeping the matrix up-to-date.

